PR #28, opened 2012-11-26
JENKINS-15925 - Changed delimiter of securityGroups from space to semi-colon.

garlandkr:
JENKINS-15925 - Changed delimiter of securityGroups from space to semi-colon.

https://issues.jenkins-ci.org/browse/JENKINS-15925

### DIFF
--- a/src/main/resources/hudson/plugins/ec2/EC2Slave/help-securityGroups.html
+++ b/src/main/resources/hudson/plugins/ec2/EC2Slave/help-securityGroups.html
@@ -1,5 +1,5 @@
 <div>
     Specify the security group names (do not specify the security group IDs) that will be applied to this instance.
-    If there are multiple security groups, separate them by a semi-colon (ex: Group One; Group2; Group 3;). If there is only one group, place a semi-colon at the end of the line (ex: Group 1;). If the instance uses a VPC (you have specified
+    If there are multiple security groups, separate them by a semi-colon (ex: Group One;Group2;Group 3;). If there is only one group, place a semi-colon at the end of the line (ex: Group 1;). If the instance uses a VPC (you have specified
     a subnet ID for the VPC), then all security groups need to be part of the VPC.
 </div>


johnnyshields:
@francisu is there an example of a plugin (this one or other) which converts the config when loading the new version? (I imagine it would also need a "rollback")

francisu:
This one is a bit tricky. Unfortunately there is no version mechanism in the metadata, so the migration is typically done by looking at the data and seeing if it's "old" or "new". However that's not possible in this case. One suggestion I got was to simply create a new name for the property with the data, but I find that unacceptable. The semantics of the property is exactly the same, so there is no cause to rename it. The only think I can think of is to add some kind of version property for the plugin as a whole and use that to help control this sort of migration. But then it gets even more messy as often the XML configuration files are updated directly (and not through the UI) so no checking is possible.

This is why this one has been sitting around for so long.

johnnyshields:
@francisu is there any way to display a warning message when you upgrade?

johnnyshields:
Hey wait a minute, the **existing code** looks like it's comma-delimited...

```
(Arrays.asList(securityGroups.split("\\s*,\\s*")));
```

This regex splits on commas and removes leading/trailing whitespace. 

For example, the string "foo bar  ,   bar qux" will be split into "foo bar" and "baz qux"

So I think the existing behavior should work as-is.

francisu:
It would be something good to write a JUnit for I think :)

On Mon, May 9, 2016 at 6:01 AM, Johnny Shields <notifications@github.com>
wrote:

> Hey wait a minute, the *existing code* looks like it's comma-delimited...
>
> (Arrays.asList(securityGroups.split("\\s*,\\s*")));
>
> This regex splits on commas and removes leading/trailing whitespace.
>
> For example, the string "foo bar , bar qux" will be split into "foo bar"
> and "baz qux"
>
> So I think the existing behavior should work as-is.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/jenkinsci/ec2-plugin/pull/28#issuecomment-217857571>
>



-- 
Cell +1 510 432 1589 (when in US)
Cell +33 7 83 72 48 66 (when in Europe)

johnnyshields:
Studying a bit more on the problem, according to the docs here:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html

The following are allowed:

> a-z, A-Z, 0-9, spaces, and ._-:/()#,@[]+=&;{}!$*

So, actually neither comma or semi-colon is a good idea.

So, if we want to permanently solve the problem, there are two options:

1. Choose a delimiter not on the list, such as pipe `|`. Do a one-time conversion from commas to pipes when upgrading from v1.32 or lower.
2. Make the field a list input (with an "add" button.) Again this would require a conversion from the old field.

PR #99, opened 2014-07-17
Workaround AWS bug that hides stopped slaves

acoulton:
Use instance.group-name instead of group-name as the describeInstances
filter when searching for stopped slaves to restart.

Some EC2 regions are incorrectly returning empty describeInstances
results when the group-name filter is used. This causes jenkins to start
new slaves instead of restarting stopped instances.

instance.group-name is functionally identical and seems to work
consistently.

Fixes https://issues.jenkins-ci.org/browse/JENKINS-23850

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -319,7 +319,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
                /* No subnet: we can use standard security groups by name */
                 riRequest.setSecurityGroups(securityGroupSet);
                 if (securityGroupSet.size() > 0)
-                    diFilters.add(new Filter("group-name").withValues(securityGroupSet));
+                    diFilters.add(new Filter("instance.group-name").withValues(securityGroupSet));
             }
 
             String userDataString = Base64.encodeBase64String(userData.getBytes());


iongion:
@acoulton you said:

>> Some EC2 regions are incorrectly returning empty describeInstances
>> results when the group-name filter is used.

How do you actually overcome this ?
My question is unrelated to Jenkins or this plugin, but you seem to be the only one who realized it, as of today I have the issue for `sa-east-1` region.

acoulton:
@iongion per the pull-request diff, use `instance.group-name` instead of just `group-name` as the filter. Assuming you have the same problem, that should work.

iongion:
@acoulton hehe, yes, I have changed the query to do filtering on security group client side by asking the instance's groups, I leave this info here for other googlers that might encounter the issue.
(PS: using boto, so `instance.groups` not `instance.group_name`)

acoulton:
@iongion that's not what I meant. If you pass AWS a request with a filter key `instance.group-name` instead of `group-name` you should still be able to do filtering server-side. If you look at the [describeInstances docs](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html) you'll see they list:

> *  `group-name` - The name of the security group for the instance. EC2-Classic only.

> ...

> *  `instance.group-name` - The name of the security group for the instance.

AFAIK boto like most other EC2 clients just pass whatever filter names you provide through to EC2 as-is, so it shouldn't make any difference.

iongion:
@acoulton oh, even better!

PR #111, opened 2014-10-19
JENKINS-25209: Add "reboot slave after build" for a slave template

arcivanov:
Adds slave template option to reboot a slave after a build
Once reboot commences the slave is monitored, reconnects are started and
the slave is reenabled once reconnect succeeds.
Fixes NPEs in tag handling
Make constructors more tolerant of null-strings.
Fix EC2ComputerLauncher switch state handling and breaks-to-label

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -33,6 +33,8 @@ import hudson.model.Node;
 import hudson.model.Slave;
 import hudson.slaves.NodeProperty;
 import hudson.slaves.ComputerLauncher;
+import hudson.slaves.OfflineCause;
+import hudson.slaves.OfflineCause.ByCLI;
 import hudson.slaves.RetentionStrategy;
 import hudson.util.ListBoxModel;
 
@@ -41,11 +43,19 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
+
+import javax.servlet.ServletException;
+
+import jenkins.util.Timer;
 import net.sf.json.JSONObject;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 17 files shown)


arcivanov:
retest this please

francisu:
Thanks Arcadiy, I'm having trouble testing this as there seems to be some problem with Eclipse when using the more recent Jenkins libraries. Please give me some more time on this. I'm not going to take anything else though until I have finished with your pull request.

arcivanov:
Thanks, Francis. There is something weird going on on Eclipse with older libraries indeed. I've had to modify pom to use 1.565.3 while testing locally and it worked fine. I would then change the pom.xml back to the older version of Jenkins libraries before compiling from the command line/CI and checking in.

francisu:
Hi Arcadiy, I spent a little time with this. I changed the version in the pom to use 1.565.3 and I had to remove the "@DataBoundConstructor" from EC2AbstractSlave to get it to compile. And then when I ran the tests, a couple of them failed: http://pastebin.com/Xqu3NNyN.  With the original pom version, 35 of the tests failed (just running "mvn clean test" I spent some time trying to work out the problems with the test that I think they are in the handling of passwords, probably using the Secret class, but I ran out of time I could spend to try and come up with a fix. If you are unable to reproduce the original test failures, let me know and I can give you the output. Can you look at this some more? 
Thanks,
Francis

arcivanov:
This is the reason why the tests fail:

WARNING: Caught exception evaluating: h.getPasswordValue(attrs.value ?: instance[attrs.field]) in /configure. Reason: java.lang.SecurityException: attempted to render plaintext ‘Rc9PPykm5L1LpETfW45sWXV9A/5+rNILB2tMxC9VPt8=’ in password field; use a getter of type Secret instead

The problem is that 	at hudson.Functions.getPasswordValue(Functions.java:1682) we now force unit test failures for security reasons.

johnnyshields:
Just a note, this option should not be enabled for spot instances as they cannot reboot.

johnnyshields:
@francisu I will pickup this feature and combine it with some others I have planned. We can leave this PR open for now.

arcivanov:
@johnnyshields was this merged in one form or the other?

johnnyshields:
Not yet, if you want to help with it you're welcome to.

arcivanov:
@johnnyshields help with what? :smile: 

I have rebased on the latest master and this is ready for merge on its own merits. I have removed the reboot option for spot instances per above and the option will be ignored for them.

arcivanov:
@johnnyshields @francisu if you need "help", could you at least specify help with what?

johnnyshields:
@arcivanov sorry for holding up the show here. I'm fine with this PR as is from looking at the code. I assume you've tested it locally and no issue? @francisu ?

arcivanov:
@johnnyshields Sorry for late response. Yes, I confirm the branch in the current state works in our production. Thanks!

PR #127, opened 2015-01-09
[FIXED JENKINS-26371] allow creating instances that terminate on shutdown

devennorton:
will confess to not creating a specific unit test for this feature however I don't really see a good way to unit test the change that doesn't boil down to does assignment operation work.


francisu:
I think there is another option that allows the same thing except that it's d pressed differently. It's to stop on shutdown. If that option is turned off the instance will terminate instead.

francisu:
I'm going to close his one. If I have it wrong please reopen and explain why. Thanks.

PR #129, opened 2015-01-23
Allow multiple clouds per region

renat-sabitov-sirca:
Allow multiple clouds per region
Also, new AWS SDK to get new instance types

Please don't merge, it is a quick hack and it should've been done better. We have multiple AWS accounts per region and therefore need it.

Anyone who has a good understanding of the plugin and could direct to the right way is welcome.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk</artifactId>
-      <version>1.8.3</version>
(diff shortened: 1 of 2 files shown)


devennorton:
Maybe am stupid and perhaps is besided the point but why  submit pull request you not want merged?

renat-sabitov-sirca:
@devennorton Valid point. I had two reasons on my mind:

- Instead of just submitting a bug report provide some way to work around
- increase visibility of the workarond for other people who might be interested in

Feel free to close the pull request if the reasons don't make sense.

PR #134, opened 2015-02-13
Code cleanup

renat-sabitov:
No change in functionality, just some
imports reorganisation, deprecated method calls removal etc.


kunickiaj:
Nice.. this code seems to have about ten different styles throughout. :+1:

PR #135, opened 2015-02-22
Small cleanup in logging statements

techdragon:
I cleaned up the on calls to log the launch of on demand and spot instances so that they are now the same format and also now both make the same calls, previously provisionSpot was missing the call to LOGGER.info


techdragon:
I cleaned up the commit a little more to merge nicer.

PR #136, opened 2015-02-22
Spot Instance Device Mapping

techdragon:
- Spot instances can now map block devices and have the instance ephemeral storage devices mapped automatically just like On Demand instances can.
- Overloaded the methods to provide support for the ephemeral storage and block device mapping on the LaunchSpecification object types.


techdragon:
This pull request passes the test suite, but having tested it on my own Jenkins server, it doesn't work, not sure why at the moment. 

It seems to create a spot instance request correctly, but never completely returns the new node's information to create the corresponding slave node object.

francisu:
I'm going to close this one since you have not been able to make it work. If you have something that works, you are welcome to resubmit. Thanks, Francis

PR #137, opened 2015-03-02
JENKINS-27193 Use proper error code to detect not-yet-started slave.

pfalcon:
"InvalidInstanceRequestID.NotFound" is not available in
http://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html .
"InvalidInstanceID.NotFound" matches both this doc and what actually appears
in Jenkins log.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -426,7 +426,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
                             updateRemoteTags(ec2, inst_tags, inst.getInstanceId());
                             break;
                         } catch (AmazonServiceException e) {
-                            if (e.getErrorCode().equals("InvalidInstanceRequestID.NotFound")) {
+                            if (e.getErrorCode().equals("InvalidInstanceID.NotFound")) {
                                 Thread.sleep(5000);
                                 continue;
                             }


pfalcon:
See https://issues.jenkins-ci.org/browse/JENKINS-27193

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -426,7 +426,7 @@ private EC2AbstractSlave provisionOndemand(TaskListener listener) throws AmazonC
                             updateRemoteTags(ec2, inst_tags, inst.getInstanceId());
                             break;
                         } catch (AmazonServiceException e) {
-                            if (e.getErrorCode().equals("InvalidInstanceRequestID.NotFound")) {
+                            if (e.getErrorCode().equals("InvalidInstanceID.NotFound")) {
```
KostyaSha: If you are unsure about previous ErrorCode or it was deprecated, then better do `if ( oneError || oneError) {` this will allow to have backward compatibility + put some comment that this error code doesn't exist anymore

KostyaSha: Nevermind, looks like a real typo, backward compatibility not need here

KostyaSha:
:+1: looks good
@francisu please review and merge for @pfalcon

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
pfalcon: > If you are unsure about previous ErrorCode

I am not 100% sure, that's why I submitted it for review linking to research I did in the Jenkins bugtracker ticket. If the maintainer asks to leave the old code, I can add it, but I'm reasonably sure it's just a typo, so leaving it in the code would be shame.

pfalcon:
Updated commit message to start with "[FIXED JENKINS-27193]", as suggested by @KostyaSha on IRC.

francisu:
Thanks for this, this was covered as well (with some additional refactoring) in #140 which I just merged. Going to close this one.

PR #138, opened 2015-03-06
[JENKINS-27260] SPNEGO for Windows in EC2 Plugin

jmittert:
As per [Jenkins-27260](https://issues.jenkins-ci.org/browse/JENKINS-27260):

JENKINS-25385 and JENKINS-4995 both have comments complaining about infinite loops when creating Windows slaves. Because SPNEGO is unregistered for, the httpclient throws an exception that is silently
caught and causes the infinite loop. 

I was able to connect to WinRM, but I had to re enable SPNEGO. I'm not sure what repercussions that will have otherwise, I'm hoping someone can comment on that. 

I also found that the tmpDir was getting set to "", causing a failure, so I added in a case for that.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
@@ -30,7 +30,7 @@ public class EC2WindowsLauncher extends EC2ComputerLauncher {
 
         try {
             String initScript = computer.getNode().initScript;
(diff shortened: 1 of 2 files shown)


rsennewald:
We're using this in production at my workplace and it's working perfectly for Windows AMIs to be spun up.  @francisu would you mind taking a look and merging if all seems well?

francisu:
Thanks for your contribution!

PR #140, opened 2015-03-18
JENKINS-26797 - Zombie instance created by the EC2 plugins

cbek:
* fixing error code for provisionOndemand - InvalidInstanceID.NotFound
* refactoring updateRemoteTags - no code duplication
* no exception throwing if remote tagging fails - only logging
** otherwise we got zombie instances which are not shown
* adding unit testing

https://issues.jenkins-ci.org/browse/JENKINS-26797

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,12 @@ THE SOFTWARE.
 
   <dependencies>
     <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <!-- we only use this to handle key fingerprint. should be able to replace this with trilead -->
       <groupId>bouncycastle</groupId>
       <artifactId>bcprov-jdk15</artifactId>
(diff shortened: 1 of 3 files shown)


cyrus-one:
thanks for this fix!

cbek:
@francisu 
Can we merge this?
People waiting for fix: https://issues.jenkins-ci.org/browse/JENKINS-26797
I tested the fix on our live jenkins, problem seams to be fixed.
Today, I will prepare the change log, so that we can trigger the release of 1.27 :-)

tjbaker:
I'd be great to get this in, thanks!

daniel-beck:
This increases the visibility of `updateRemoteTags` for no apparent reason. Is this deliberate?

cbek:
@daniel-beck 
What do you mean by increase the visibility of ```updateRemoteTags```?

The exception is now only logged, which will make sure that the new node/instance can be stored in jenkins config. Of course this means it can happen that instances miss tags in AWS, but this much better then having zombie instances.

daniel-beck:
@cbek It used to be `private`, and now is `protected`. There is no obvious reason for that.

cbek:
@daniel-beck 
Ok, agree.
Introduced powermock-module-junit4 for invokeMethod for testing private method.
I guess makes more sense now :-)

cbek:
@francisu - Any concerns? Can we merge and release?

PR #131, opened 2015-02-10
JENKINS-26854: Fix for not being able to start EC2 slaves after several hours

ximon18:
Minimal improvement to ensure EC2 instance profile credentials are refreshed so that after the credentials expire the new rotated credentials made available by the EC2 metadata service are automatically fetched. Upgrades the AWS Java SDK to v1.8.11, the first stable version after the background refresh of instance profile credentials was introduced in unstable v1.8.10.

PR #146, opened 2015-05-19
Fixing instance profile credentials with 'Check AMI' and 'Check Current Spot Price' buttons 

jjudd:
The 'Check AMI' and 'Check Current Spot Price' buttons  currently do not work when instance profile credentials are used.


jjudd:
It looks like #131 also fixes this. I have opened #147, which resolves merge conflicts from #131. Accordingly, this PR can be closed if either is merged.

PR #155, opened 2015-07-06
Magic file preventing from init script launch

mieciu:
Hello,

Today I've spent quite a much time on debugging, why my init script does not work for certain EC2 build slaves. After reading the code  I've found magic happening with ` ~/.hudson-run-init`, which I could not find in documentation. And true, I've built that AMIs from instance already containing `.hudson-run-init` file.

I guess that this might be handy when slave reconnecting happens, but on the other hand, I'd rather expect such a provisioning script to be idempotent.

It's not like I'm forcing my solution with removal here, but for sure this magic file presence should be either covered in documentation or handled from UI (`do not run within reconnect` flag under init script text field)...

Please let me know what do you think, I'd be happy to collaborate on handling this!

### DIFF
--- a/src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-initScript.html
+++ b/src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-initScript.html
@@ -1,0 +1,5 @@
+<div>
+  Init script specifies commands to be ran after very first slave connection to Jenkins.
+  Once finished, it created <tt>~/.hudson-run-init</tt> file to prevent Jenkins from running
+  it within each reconnection.
+</div>


francisu:
Thanks for your contribution. The intention of the init script is that it only be executed once. Not each time the instance is connected to. This is why the special file is there. Accepting your pull request would change that behavior and probably break some users, so I don't think it's a good idea. However the current documentation of the init script parameter is really not clear about this. I would welcome a pull request that clarified the documentation.

mieciu:
Yeah, I'm totally aware of breaking current workflow :)

What would you say for having some kind of switch/checkbox in UI, that would force running init script every single time instance connects?

mieciu:
This would be an option for preserving old workflow as default and adding this idea as an optional feature.

francisu:
The EC2 plugin is already horribly confusing with its many options. I think this could be a good approach, but we should handle it well. I would support two different types of init scripts. One as it is now that runs only once, and another one (maybe called "connection init script" or something more clear) that runs on a per-connection basis. What do you think?

mieciu:
So maybe I could do the following ?
1) Add question-mark in UI next to current init script explaining that this script will run only once...
2) Update documentation 
3) add a checkbox "force init script execution within each reconnection", but hidden behind 'ADVANCED' button, so that it won't complicate UI

francisu:
I like items 1 and 2, but for item 3 I would propose having an entirely separate script that's a "per-connection initialization" or something like that. You could imagine using both types on some occasions, and I think decoupling the concepts would be more helpful. You could put them side-by-side in the UI but so that their separate purpose is clear.

mieciu:
Okay, I'll begin my work on a separate branch then and eventually squash commits and merge it to this one, so you won't get notified on every single commit.

mieciu:
Hey @francisu I've just added the helper textfield and substituted the commit

francisu:
Thanks for this and I look forward to another contribution to add the additional options as previously discussed.

PR #157, opened 2015-07-08
modify the sed command in the ubuntu setup script to ignore comments

davidyorr:


davidyorr:
a new ubuntu instance has this line in the /etc/rc.local

    # Make sure that the script will "exit 0" on success or any other

as the script was, it changed that line to

    # Make sure that the script will "python /usr/bin/userdata
    exit 0" on success or any other

causing a syntax error

PR #158, opened 2015-07-29
Prefer the public IP to the private IP when ssh-ing into slave

alexcojocaru:


marvinpinto:
I feel like this PR will change the expectations around how the EC2 plugin operates, especially in a VPC scenario.

For example, our EC2 build workers in a VPC _are_ assigned public IP addresses (in order for them to get out directly to the Internet), but inbound SSH connections on those public IP addresses are firewalled off (via security groups). In that scenario, our jenkins instance would not be able to ssh into the build workers.

alexcojocaru:
Would it be acceptable to have an checkbox option in the UI of the plugin, to let the user decide whether to have the master connect to a slave using its public IP instead? Checking the checkbox would activate the new functionality which I put in, otherwise it will default to the previous functionality.

francisu:
I think this is a good idea, can you revise this PR?

On Thu, Jul 30, 2015 at 9:49 AM, alexcojocaru <notifications@github.com>
wrote:

> Would it be acceptable to have an checkbox option in the UI of the plugin,
> to let the user decide whether to have the master connect to a slave using
> its public IP instead? Checking the checkbox would activate the new
> functionality which I put in, otherwise it will default to the previous
> functionality.
>
> —
> <https://github.com/jenkinsci/ec2-plugin/pull/158#issuecomment-126399439>.
>



-- 
Cell 510 432 1589

alexcojocaru:
I added a config option to allow overriding the way the plugin connects to the EC2 instance; enabling the option would prefer the public IP (if available) to the private IP. Please review the changes.

francisu:
Thanks for your contribution.

PR #159, opened 2015-08-11
fix for instance caps which are calculated solely based on AMI ID

snallami:
Currently instance caps are calculated solely based on AMI, not based on template. So, if there are two types of slaves based on a common AMI 
but intended for different purpose, if the instance cap is low on one type, it may never get provisioned if the other type has many outstanding instances.

Fixed the issue by appending template description as well to tag 'jenkins_slave_type' so that number of instances per template can be identified.

If you want to track using a new tag instead of appending template description to tag 'jenkins_slave_type', please let me know.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -24,6 +24,7 @@
 package hudson.plugins.ec2;
 
 import com.amazonaws.ClientConfiguration;
+
 import hudson.ProxyConfiguration;
 import hudson.model.Computer;
 import hudson.model.Descriptor;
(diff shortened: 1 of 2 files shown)


lielran:
we had to duplicate the AMI in order to use it in more than one place - so this PR is useful for us.
the disadvantage is that tag can be delete/change(unlike ami id) and the tag number is limited

francisu:
Thanks for your contribution. May I ask for a couple of small changes:

1) can you make a static method in EC2Cloud that creates the key used to search for based on the instance type and description. This puts that calculation in one place rather than two. You could also move the strings "spot" and "demand" to be constants. 

2) add some help text to description to explain how it will be used in combination with the ami id to get the instance counts. 

Thanks again for this and your attention to these changes.

snallami:
Thanks for the review. I will make the changes you suggested.

snallami:
@francisu , I did changes as per your review. 

Did not add additional help text to description since it is coming from "/help/system-config/master-slave/description.html"

francisu:
Thanks so much for your contribution and working with me to improve it.

PR #161, opened 2015-09-06
Specify current AWS endpoints for EC2 and S3

vassilevsky:
The plugin wasn't working for me so I dug into it and into AWS docs and here's what I came up with 👻


francisu:
Thanks for the pull request. It looks like the S3 regions will work with the simple prefix s3- as well, like the ec2 regions (look at the link you provided and you can see even the "s3.eu-central-1.amazonaws.com" region allows the "-" syntax.
Can you revise the pull request to remove the table and use the prefix?
Also, can you confirm that you deleted the methods because they were unused (or was there some other reason)?

francisu:
I think this is superseded by another pull request which just handles the EC2 regions, which is all that's necessary for this plugin.

PR #163, opened 2015-09-08
Fix NullPointerException when deleting already terminated slave

ahirreddy:
Fixes an issue where checking for uptime of a deleted node would cause an NPE.

Also fixes an issue when terminating spot instances. The new behavior will remove the node from Jenkins even if EC2 termination failed. One way for this issue to occur is that an instance was terminated manually or a spot instance was killed due to pricing. In these cases termination will fail, but the node will hang around in Jenkins forever. If we don't remove the node, we screw up auto-scaling, since it will continue to count against the quota.


ahirreddy:
Bump

PR #166, opened 2015-09-25
Create block device mappings for both ondemand and spot slaves

mtolan:
Sorry if this isn't the right process, and happy to take direction about what to do better. I just noticed that choosing the "useEphemeralDrives" checkbox for my spot instances wasn't having the effect I expected. I was seeing only a single ephemeral volume being automounted.


johnnyshields:
@francisu this looks good to merge.

PR #167, opened 2015-10-05
Use new EC2 API endpoint hostnames (JIRA-26493)

weisslj:
Every region seems to support the format ec2.<region>.amazonaws.com:
http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region

Disclaimer: I did not compile or test this change.


weisslj:
I tested this now and it works for me.

PR #170, opened 2015-10-24
Revert "JENKINS-23787 EC2-plugin not spooling up stopped nodes - "sti…

fengxx:
This reverts commit c9d69a5da8c9be094701d4c191ba7b1d06c200c9.
User can not launch multiple instances against one AMI since it will always return running instances.


francisu:
I believe I have fixed the problems caused by this original commit; it now properly starts multiple instances with the same AMI in addition to starting previously stopped instances, so this reversion is not necessary.

PR #173, opened 2015-12-02
Upgrade jenkins-core dependency to 1.580

cyrille-leclerc:
Can we bump jenkins-core dependency to 1.580 so that I can propose a Pull Request to use the aws-credentials plugin to manage the access keys in Jenkins credentials store?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-      <version>1.447</version>
+      <version>1.580</version>
   </parent>
 
   <artifactId>ec2</artifactId>
@@ -52,7 +52,7 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-      <powermock.version>1.6.2</powermock.version>
+      <powermock.version>1.6.3</powermock.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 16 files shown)


abayer:
Works for me, but adding @reviewbybees to get other thoughts. =)

stephenc:
:bee: but maybe better to go to 1.607+ so that the queue locking and node refactoring can be attempted to be picked up

oleg-nenashev:
It would be better to create a pull request with the aws-credentials plugin integration. I don't see a value in a PR, which bumps the dependency w/o proposing a feature.

Regarding the version bump, it requires a decision from plugin owners (CC @francisu). I now several big installations heavily utilizing this plugin and using 1.509 and 1.565

francisu:
Right now the current non-released version (master) has some problems which I need to work out (sometimes it starts an extra instance when it should not). These problems are regressions from the current released version. I'm going to try and fix them at best possible speed and after that do a release, which should address some important bugs people are waiting for. I would like to keep that on the current  version. Then after that I will consider functional changes. I hope this strategy makes sense.

b-long:
@francisu Thanks for the update.  So that others might help resolve the regression you're describing, can you point to the issue or pull request that contains the details?

francisu:
The problem is slaves are started beyond the maximum instance count globally  or for an AMI. I have reproduced the problem and I think I'm close to solving it, so I'm not sure I need much help at this point. But if anyone does have a solution, a pull request is always welcome.

cyrille-leclerc:
Hello @francisu, can I close this Pull Request and just keep #174 as @oleg-nenashev had suggested?

francisu:
@cyrille-leclerc yes please.

francisu:
@b-long I believe with a commit I made just now, I have fixed the regressions. If you can check the snapshot and make sure it respects the instance cap, and starts and stops slaves as expected, that would be good.

b-long:
Roger that, thanks @francisu .  We'll take a look at it! :+1: 

cc/ @dtenenba

PR #176, opened 2016-01-11
Jenkins 19059/ssh launcher for spot instance

francisu:

PR #179, opened 2016-01-25
Poll for spot instances instead of JNLP launcher JENKINS-27529 JENKINS-19059

mbarrien:
This pull request replaces the current JNLP registration and launching of spot instances with launching spot instances via SSH. After the initial round of polling for spot instance start, the resulting instance gets treated like any other (on-demand) instance. With this pull request, you no longer need to have special AMIs to handle spot instance requests. This pull request also strips out the code that mangled the EC2 user data for spot requests.

It is inspired by nexus49's contribution from JENKINS-27529, but is heavily refactored to remove redundancy with existing code and brought up to date with the head of ec2-plugin. It replaces #176 in that it uses SSH to connect to spot slaves, but also adds more formal polling and logging of spot instance state, as well as more gracefully handling cancellation of spot instance requests. It is currently being used in production Jenkins without problem, and we can freely switch between on-demand and spot request slaves without changing our script or AMI.

In detail this pull request:
* Removes @DataBoundConstructor decorator from an abstract class where it isn't compatible.
* Better encapulates EC2SpotSlave.getSpotRequest() by making it an instance method instead of a pseudo-static method.
* Adds a wait loop to the standard EC2ComputerLauncher if the instance ID is missing (which should only happen in EC2 slaves), polling every 10 seconds to see if the instance ID is set (which is automatically set when EC2SpotSlave.getInstanceId() finally gets a fulfilled spot request.
* Removes the EC2SpotComputerLauncher class; this now uses EC2UnixLauncher and EC2WindowsLauncher directly just like EC2OndemandSlave does.
* Copies the naming scheme from EC2OndemandSlave to put the description in front of the name, and uses the spot request ID as part of the name instead of a random string. It results in names like "jenkins-slave tag1 tag2 (sir-12abcd34)".
* Removes the warning from the UI about spot slaves needing special AMIs, since that is no longer true.
* Moves the logging of messages in EC2ComputerLauncher to be before the sleep instead of after (so that it actually will wait for 5 seconds after it prints, instead of saying how long it just waited.)

In the event that spot pricing is too expensive, the slave will wait loop inside EC2ComputerLauncher.launch() waiting for an instance, printing to the debug log.


francisu:
Thanks very much for this. I have one question though. As I'm not that familiar with Jenkins core, I'm concerned that the sleep loop in waiting (perhaps a long time) might tie up all of Jenkins so that it cannot service other work until the spot request is satisfied. Is this the case?

mbarrien:
Not sure myself, but I do note this comment at https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/ComputerLauncher.java#L76 in the base class ComputerLauncher.launch() method that is implemented using this wait loop:
```
This method must operate synchronously. Asynchrony is provided by {@link Computer#connect(boolean)} and its correct operation depends on this.
```

mbarrien:
ComputerLauncher.launch() appears to be called at https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/SlaveComputer.java#L253 which submits it to a thread pool, with a comment saying that it will "do this on another thread so that the lengthy launch operation (which is typical) won't block UI thread."

mbarrien:
For the record, while testing on this, I definitely had extended time waiting for a spot slave to launch (like over a minute), and I was still able to act in another window to add and remove tasks from the queue.

Also keep in mind, there *is already* a sleep loop in the code even when in on demand immediately below the sleep I added, while it waits for a computer to go from pending to started. Even when using on demand slave, it can wait there a while and if you look in the log it often does wait. I have never had a UI issue during that time (the fact that the log keeps updating in the UI is already proof the UI isn't blocked), and given how widespread use of on demand slaves is, there should already be complaints if it were blocking the UI there.

francisu:
Thanks again for your response and this wonderful work. As soon as I give it a spin I will merge it.

mbarrien:
Any word on status?

zajacmp3:
Hi I am also interested in this. I guess it will solve my issues that I reported in JENKINS-32915. Any update?

zajacmp3:
I reviewed the code and sadly I must say that it does not fix what I have reported in JENKINS-32915. It brings me even more trouble with ssh connection as my VPC in AWS is very restrictive about open ports and ssh configuration. I dare to disagree that removing JNLP launcher is a good idea. Having this removed we deny users functionality of self configuring instances. **It is a common practice to unload ssh keys on boot time and do automatic configuration of the environment with puppet scripts.**

mbarrien:
Since this makes a spot instance look like an on-demand instance, wouldn't you also have the same issues with VPC and open ports and SSH configuration with on-demand instances? (e.g. aren't the issues you have with VPC and open ports orthogonal to this PR?)

zajacmp3:
Thank you for quick answer. I will test that debug it extensively as soon as it will be possible for me and I will get back to you. I expect that onDemand request also will throw problems with timeouts for ssh connection. And the SUV will not get up and register itself as we removed the additional dirty hack of attaching userdata to request url. Just saying that if this goes in I and maybe others would have to use either old code or own forks of this plugin.

aheuermann:
Any update on this? This would be very useful.

schottsfired:
I assume https://github.com/jenkinsci/ec2-plugin/blob/master/README.md#configure-ami-for-spot-support is no longer relevant after this PR, correct?

PR #186, opened 2016-03-22
Jenkins 32915

zajacmp3:
This pull addresses detected bug in self registering EC2 Slave spot instance.
https://issues.jenkins-ci.org/browse/JENKINS-32915
It was caused by the fact that we return a future that calls node instantly even before instance is ready and could not self deploy itself. I added a timeout that is respected and during that time it will retry connecting to that instance. While doing that promise wrapped in 'PlannedNode' looks like an instance that will be ready soon to take tasks from the queue and not an instant failure. By that Jenkins properly calculates and scales spot instances. And it had significant problem with it. I also noticed that if you do not provide a label to which ec2 cloud job is restricted it will have problems calculating default number of executors for the planned node which also affects excess workload calculation. I added a log message that addresses that. It can be 0 if you set you master server to use nodes only for doing jobs with number of executors to 0.

It is scaling properly in my company for at least a month now. We are scaling from 0 instances to ~20 during peaks and I did not notice any problems even with many cloud configurations.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -89,6 +89,7 @@ public abstract class EC2AbstractSlave extends Slave {
     public final String idleTerminationMinutes;
     public final boolean usePrivateDnsName;
     public final boolean useDedicatedTenancy;
+    public boolean isConnected = false;
     public List<EC2Tag> tags;
     public final String cloudName;
     public AMITypeData amiType;
@@ -386,7 +387,7 @@ public abstract class EC2AbstractSlave extends Slave {
      * Called when the slave is connected to Jenkins
      */
     public void onConnected() {
-        // Do nothing by default.
+        isConnected = true;
     }
 
     protected boolean isAlive(boolean force) {
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -426,31 +426,32 @@ public abstract class EC2Cloud extends Cloud {
         try {
             List<PlannedNode> r = new ArrayList<PlannedNode>();
             final SlaveTemplate t = getTemplate(label);
-            LOGGER.log(Level.FINE, "Attempting provision, excess workload: " + excessWorkload);
+            LOGGER.log(Level.INFO, "Attempting provision, excess workload: " + excessWorkload);
             if (label == null) {
-                LOGGER.log(Level.WARNING, "Label is null");
+                LOGGER.log(Level.WARNING, String.format("Label is null - can't caculate how many executors slave will have. Using %s number of executors", t.getNumExecutors()));
             }
-            boolean canTakeTask = itCanTakeTask();
-            while (excessWorkload > 0 && label != null && !canTakeTask) {
+            while (excessWorkload > 0) {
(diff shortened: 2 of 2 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -426,29 +426,32 @@ private synchronized EC2AbstractSlave provisionSlaveIfPossible(SlaveTemplate tem
...
-                            if (t.spotConfig != null) {
-                                LOGGER.log(Level.INFO, "Expected - Spot instance " + slave.getInstanceId()
-                                        + " failed to connect on initial provision");
-                                return slave;
+                        long startTime = System.currentTimeMillis(); // fetch starting time
+                        while (false || (System.currentTimeMillis() - startTime) < slave.launchTimeout * 1000) {
+                            try {
+                                return tryToCallSlave(slave, t);
+
+                            } catch (Exception exception) {
```
francisu: You should not swallow the exception here. Simply don't rethrow it from the trytocallslave method.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -426,29 +426,32 @@ private synchronized EC2AbstractSlave provisionSlaveIfPossible(SlaveTemplate tem
...
                     public Node call() throws Exception {
-                        try {
-                            slave.toComputer().connect(false).get();
-                        } catch (Exception e) {
-                            if (t.spotConfig != null) {
-                                LOGGER.log(Level.INFO, "Expected - Spot instance " + slave.getInstanceId()
-                                        + " failed to connect on initial provision");
-                                return slave;
+                        long startTime = System.currentTimeMillis(); // fetch starting time
+                        while (false || (System.currentTimeMillis() - startTime) < slave.launchTimeout * 1000) {
```
francisu: Please remove "false ||"

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -386,7 +387,7 @@ public boolean getStopOnTerminate() {
      * Called when the slave is connected to Jenkins
      */
     public void onConnected() {
-        // Do nothing by default.
+        isConnected = true;
```
francisu: Who is using this new Boolean?

francisu:
Please address the issues in this pull request and I will merge it.

jjudd:
Curious what the status of this is. Would love to see this issue fixed. I can jump in and address the concerns, but if something is being done, then I don't want to step on any toes.

francisu:
@jjudd thanks, if you would clean up the conflicts, I will merge it.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
jjudd: Looks like it is used here https://github.com/jenkinsci/ec2-plugin/pull/186/files#diff-b1df2806507eaf67657940f500845f0bR476

francisu: Thanks, did not see that.

jjudd:
I cleaned up the conflicts and made the changes you requested in #193.

johnnyshields:
@francisu please close this PR since #193 has been merged.

PR #187, opened 2016-03-31
JENKINS-33945 allow node provisioning when matching node is offline

tjbaker:
You can not provision a new node when an existing node from the template
is marked offline.  This change consults isOffline() to allow new node
provisioning.  Nodes can either be offline due to channel errors or user
interaction.  Either way, don't block new nodes.


francisu:
I'm going to back this one out for now. I want to solve the problem of https://issues.jenkins-ci.org/browse/JENKINS-32690 which I think will also solve this problem. I think putting your fix in will break the feature of using a stopped node that already exists which is also important. I will try to get to this in the next couple of days.

PR #189, opened 2016-04-09
add skipCheckInstance

jimjh:
Setting `skipCheckInstance` to true allows the plugin to start instances even when there is idle capacity. This works because the [standard strategy](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/NodeProvisioner.java#L603) implemented by the NodeProvisioner already takes into account idle capacity and planned capacity.

Without this flag, it's impossible to start EC2 instances in parallel.


johnnyshields:
@francisu I believe this PR is trying to solve the same issue you have in https://github.com/jenkinsci/ec2-plugin/pull/195 ? If so should close it.

jimjh:
Yep. It's the same.

PR #191, opened 2016-04-14
Cleanup

Jimilian:
Small code cleanup:
* Use `final` if it's possible
* Use `isEmtpy()` instead of `length comparison`
* Use CONSTANTS instead of 'magic numbers'

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -18,45 +18,6 @@
  */
 package hudson.plugins.ec2;
 
-import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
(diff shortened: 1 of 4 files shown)


johnnyshields:
@Jimilian this is conflicted

johnnyshields:
@Jimilian thanks for this PR. Please help resolve conflicts on this. I will review and merge promptly if you can do so.

Jimilian:
@johnnyshields I had solved conflicts

johnnyshields:
@francisu this one looks good to me, please review and merge

francisu:
Hi, looks like there are some more conflicts, sorry about that. Also, can you change back the import *'s. In general, I think it's better to have a separate import for each class. If you do these things, then I will merge it quickly. Thanks for the cleanup, it's nice.

PR #195, opened 2016-05-07
JENKINS-32690 Make manually provision really work

francisu:


francisu:
@johnnyshields - can you give this a test?

johnnyshields:
Sure, I can't seem to reproduce the error anymore tho...

johnnyshields:
@francisu I'm reading the code and it seems all this bitmask stuff is unnecessarily complex.

Would it be possible instead to remove both the allowCreateNew and forceCreateNew flags and just always attempt to provision in the instance?

francisu:
@johnnyshields. The point of the original change is to allow a slave that it stopped to be started. If I removed the flags, that would not happen. Requested provisioning is different than when it's trying to find a node to run a job.

johnnyshields:
I think the bitmask logic along with a variable the generic name "options" makes this code hard to read. As long as there are only two possible flags, we should make everything to be booleans.

johnnyshields:
Also should have some form of test here if possible, given that this logic is the source of a lot of noise on JIRA.

francisu:
I generally don't like more than one boolean in a method, but I think
changing it to an EnumSet would make if more clear (and modern). Can you
live with that?

A test with this is difficult though since there is no means of testing
with the live EC2. I agree that we need tests, but its a much bigger
project to set something up to allow testing with EC2 (even a mock EC2).

On Sun, May 8, 2016 at 2:55 AM, Johnny Shields <notifications@github.com>
wrote:

> I think the bitmask logic along with a variable the generic name "options"
> makes this code hard to read. As long as there are only two possible flags,
> we should make everything to be booleans.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/jenkinsci/ec2-plugin/pull/195#issuecomment-217706911>
>



-- 
Cell +1 510 432 1589 (when in US)
Cell +33 7 83 72 48 66 (when in Europe)

francisu:
Please see the changes to EnumSet, hope it's more clear.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -571,7 +569,8 @@ private EC2AbstractSlave provisionOndemand(TaskListener listener, int options) t
             }
 
             if (existingInstance == null) {
-                if ((options & (PROVISION_ALLOW_CREATE_NEW | PROVISION_FORCE_CREATE_NEW)) == 0) {
+                if (!provisionOptions.contains(ProvisionOptions.FORCE_CREATE) &&
+                    !provisionOptions.contains(ProvisionOptions.ALLOW_CREATE)) {
```
johnnyshields: This can be simplified to:

```
if (provisionOptions.isEmpty())
```

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -366,22 +366,20 @@ public String getIamInstanceProfile() {
...
     /**
      * Provisions a new EC2 slave or starts a previously stopped on-demand instance.
      *
      * @return always non-null. This needs to be then added to {@link Hudson#addNode(Node)}.
      */
-    public EC2AbstractSlave provision(TaskListener listener, int options) throws AmazonClientException, IOException {
+    public EC2AbstractSlave provision(TaskListener listener, EnumSet<ProvisionOptions> provisionOptions) throws AmazonClientException, IOException {
         if (this.spotConfig != null) {
-            if (options != 0)
+            if (provisionOptions.contains(ProvisionOptions.ALLOW_CREATE) || provisionOptions.contains(ProvisionOptions.FORCE_CREATE))
```
johnnyshields: This can be be simplified to:

```
if (!provisionOptions.isEmpty())
```

johnnyshields:
@francisu OK this looks good.

By the way, you could use `if (!provisionOptions.isEmpty())`, though the way you have it may be more explicit and future proof in case other Enum values are added so it's fine as-is I guess.

Green-light to merge!!

francisu:
Yes. That's exactly why I made that choice. In case other options were added. The point of the options is to be independent. I would like to have others test it before I merge it to make sure it actually fixes the problem. 

Thanks for the review.

PR #196, opened 2016-05-08
JENKINS-34671 - Remove spot request bid type setting

johnnyshields:
With slaves now being fully controlled by Jenkins master, there is no longer a need to be able to set "persistent" spot bids--in fact it's dangerous because it will cause "zombie" spot instances to be auto-created by AWS which are not visible to Jenkins!


johnnyshields:
I've tested in Jenkins itself, working fine.

johnnyshields:
@francisu ready to merge.

PR #197, opened 2016-05-09
JENKINS-34667 Provision attempt is made when possible slaves count is 0.

francisu:


johnnyshields:
:+1: I saw this one too. Looks good.

PR #199, opened 2016-05-12
Clarifies documentation of negative idle termination times.

Antillean:
I was so confused by the documentation in `src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-idleTerminationMinutes.html` (which shows up on the Manage Jenkins page) on what negative idle times do that I consulted the source code to understand what it meant. This pull request is my attempt to clarify that documention.

I've also documented that the idle termination time is also used to determine how frequently a node is checked for idleness. As part of that I made a guess (based on my observation of behaviour) as to from when and how frequently the plugin checks when a negative idle time is specified -- that's the sentence "_If it has an idle timeout of -5 it will first be checked at 11:55, and then continuously until 12:00_". I'm particularly unsure about the "continuously".


johnnyshields:
Looks good to me.

PR #201, opened 2016-06-02
Now provisions slaves even if other slaves having different labels ha…

mathjeff:
…ve available executors

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -337,7 +337,7 @@ public abstract class EC2Cloud extends Cloud {
         }
 
         try {
-            EC2AbstractSlave node = provisionSlaveIfPossible(t, true);
+            EC2AbstractSlave node = getNewOrExistingAvailableSlave(t, null, true);
             if (node == null)
                 throw HttpResponses.error(SC_BAD_REQUEST, "Cloud or AMI instance cap would be exceeded for: " + template);
             Jenkins.getInstance().addNode(node);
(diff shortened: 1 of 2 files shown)


mathjeff:
What's been going on in our Jenkins is that we have several pools of servers that are scaled up and down by the EC2 plugin. Different jobs generate different state (downloading Maven dependencies or deploying to Tomcat) that could be recreated on other servers but that it takes more time to recreate from scratch. 

So, we have a separate pool of build slaves for each job. Each pool generally generates slaves having one label and each job only runs on slaves of a specific label.

The AMI required by each job is the same, however. This seems to cause the EC2 plugin to deem the servers interchangeable, hence this patch.

Does this patch seem like something that merits a unit test or a Jira issue or anything like that?

tjbaker:
I think this is a dupe of JENKINS-2760, which has been fixed.  This works as long as your slave configurations have unique "descriptions".
https://issues.jenkins-ci.org/browse/JENKINS-27601

mathjeff:
Hey thanks for the response.
I think that JENKINS-27601 is another situation exhibiting the same symptom of slaves being not provisioned.

We do have different values set for the "Description" field of each SlaveTemplate (just above "AMI ID"), and we were observing problems with version ec2-1.31 (it looks like the patch for the JENKINS-27601 was merged before version ec2-1.30).

When we were using version ec2-1.31, I observed a bunch of lines in the log saying " true - Node has capacity - can use it". During those times, I observed that some pools had extra executors (presumably what was triggering this message) and some pools were backlogged but no nodes from the backlogged pools were being provisioned. I also found that if I deleted one of then idle slaves then Jenkins would be willing to provision more slaves than it had had when I had deleted one.

I also found that when I built and deployed this patch (so 1.34-SNAPSHOT), then I saw a bunch of lines in the log saying "false - we need a Node having label...", as if without the newly added check, the existing slave of a different label would have been deemed satisfactory for the provisioning requirement and prevented the provisioning of a new node (yet the new node is actually required because the existing node didn't have the label needed by the asking job).


Your comment does bring up the question about whether the check that I'm adding should actually be validating that the description of the slave in question matches the description of the SlaveTemplate that is needed. I guess that that would be easier to reason about. It might cause a problem though if someone manually removes a label from a bunch of EC2-provisioned slaves then we could run into this same problem. I don't expect people to be doing that, although having that also be preemptively fixed seems good to me. Thoughts?

JEBoothjr:
We are running into this exact scenario as well. Same AMI, but each job has a different label. Job "A" is started and provisions "slave-a" label fine. Job "B" is started but instead of starting an instance of "slave-b" it logs "Found existing pending or running: running instance:..." whereas I would expect to see "No existing instance found - created new instance". Because we have capped the executors, it just loops, attempting to provision using "slave-a". If I go into manage nodes and manually provision "Slave B", Job "B" picks it up and continues on as it should. The descriptions are all unique.

francisu:
I think this change is right. @mathjeff can you file a JIRA about this? Sadly, we don't have the ec2 service mocked so it's hard to do junit tests for it (even though we should). It would be great if someone could spend some time to set us up with a junit test environment for a mock ec2 setup. That would make the quality of this beast a whole lot better.

francisu:
nvm, I found the JIRA for this.

PR #203, opened 2016-06-07
[JENKINS-35457] Migrate from using bouncy castle to bouncycastle-api plugin

alvarolobato:
[JENKINS-35457](https://issues.jenkins-ci.org/browse/JENKINS-35457)
Dependency to multiple Bouncy Castle versions from jenkins core and plugins is causing problems due to the binary incompatibility between versions, the different supported algorithms, etc.

Plugins should be migrated to use a common API and BC version in order to avoid these problems, see [bouncycastle-api plugin](https://github.com/jenkinsci/bouncycastle-api/blob/master/README.md)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,10 +105,9 @@ THE SOFTWARE.
 			</exclusions>
     </dependency>
     <dependency>
-      <!-- we only use this to handle key fingerprint. should be able to replace this with trilead -->
-      <groupId>bouncycastle</groupId>
-      <artifactId>bcprov-jdk15</artifactId>
-      <version>140</version>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>bouncycastle-api</artifactId>
+      <version>1.0</version>
     </dependency>
   	<dependency>
 	  <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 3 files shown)


johnnyshields:
:+1: fine by me. @francisu

rsandell:
:bee:

recena:
🐝

recena:
@reviewbybees done

alvarolobato:
@johnnyshields I consider this PR is ready to be merged. If you agree it would be great if you could release a new version. Thanks.

recena:
@francisu Thanks so much! Do you plan to cut a new release?

francisu:
Sure, I can do that in the next couple of days.

recena:
@francisu This would be great!

francisu:
We did this, but then we got this bug filed: https://issues.jenkins-ci.org/browse/JENKINS-36197  Can someone look into this? I don't have the time right now. If it can't be addressed in a couple of days, I will need to revert this one, as I hate having a broken current release out there.

johnnyshields:
@alvarolobato can you help with JENKINS-36197 ?

alvarolobato:
Just responded on the Jira Issue.

PR #208, opened 2016-06-30
[JENKINS-36197] Upgrade to plugin-pom 2.11

alvarolobato:
Upgraded to plugin-pom 2.11 and raised core version to 1.648 to avoid having to update from BC API 1.0 to 1.648 as requested on [JENKINS-36197](https://issues.jenkins-ci.org/browse/JENKINS-36197)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 -->
+
 <!--suppress ProblematicWhitespace -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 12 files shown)


alvarolobato:
@reviewbybees

pom.xml
```diff
@@ -55,6 +56,7 @@ THE SOFTWARE.
 
     <properties>
         <powermock.version>1.6.3</powermock.version>
+        <jenkins.version>1.648</jenkins.version>
```
oleg-nenashev: 🐛 for bumping to a non-LTS version. Better to use 1.651.3 IMHO

src/main/resources/hudson/plugins/ec2/EC2SpotComputerLauncher/main.properties
```diff
@@ -1,25 +1,24 @@
-<!--
-The MIT License
-
```
oleg-nenashev: Oh... wow. Thanks for catching it

oleg-nenashev:
The version dep needs fix, other stuff LGTM.
Maybe we could avoid updating the core dependency, but it requires a decision by mentor.

The CI needs a more powerful machine from what I see. I'll fix it

oleg-nenashev:
retriggering the build

alvarolobato:
@oleg-nenashev changed the baseline

oleg-nenashev:
🐝

PR #210, opened 2016-07-07
Ensure EBS optimized flag is passed to spot instances

jamiesnape:
Currently, the EBS optimized flag is ignored for spot instances.


johnnyshields:
:+1:

PR #211, opened 2016-08-04
Avoid double counting spot instances

lsmith62:
It looks like spot instances were being double counted (at the instance level and then again for the spot instance request itself). This appears to fix it for me.


francisu:
Thanks for your contribution!

PR #213, opened 2016-08-16
JENKINS-37449. Now recovering if a Jenkins node has no corresponding …

mathjeff:
…EC2 instance at the time that Jenkins starts


francisu:
Very nice, thanks for your contribution!

PR #214, opened 2016-08-17
Tweaks to avoid threads deadlocking

lsmith62:


francisu:
There was a problem with this?

PR #215, opened 2016-08-18
Fix issue when counting non-fulfilled spot requests

lsmith62:
Spot instances that did not have associated instance IDs (e.g. just started or price too low) were not being counted properly. Added stricter checks to fix this behavior and still avoid double counting fulfilled requests.

Unrelated, we're still looking at the deadlock issue but it needs a bit more testing...


lsmith62:
Todd's PR includes this same fix, so decline this if Todd's is accepted.

PR #216, opened 2016-08-18
workaround for JENKINS-37483; refactor slave counting to make it more efficient

trose67:
There's a few different ways to fix JENKINS-37483, I'm trying to do so without changing the overall locking behavior too much.

Included in the EC2Cloud changes are more fixes for correctly counting SIRs when they are in open/unfulfilled state.  Also a refactor for the slave count methods so that they only call ec2-describe-instances once.  We have hundreds of instances and calling this twice seems inefficient.

We are still testing this at scale.  

One open issue with the spot feature is that there's no support for canceling the SIR when the market moves ahead of your bid for hours - we end up with starvation in the queue and are forced to re-label pending jobs manually.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -349,26 +349,33 @@ public abstract class EC2Cloud extends Cloud {
     }
 
     /**
-     * Counts the number of instances in EC2 that can be used with the specified image and a template. Also removes any
+     * Counts the number of instances in EC2 that can be used with the specified image and a template. Also returns any
      * nodes associated with canceled requests.
      *
-     * @param template If left null, then all instances are counted.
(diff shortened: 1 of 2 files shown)


francisu:
Hi @trose67, what's the status of your testing with this? I don't have much of a test environment these days, so I rely on folks to test this to make sure it's good before I take it.

trose67:
The issues addressed with this PR seem resolved, but we have lots of other challenges with the spot feature so it's hard to say if these alone are really worth taking.

We have a highly customized version of this plugin that tries to play nice with the heavy-jobs plugin, and keeps the spot bid / instance-type / availability zone up date in near realtime.   Our divergence makes it hard to contribute back to the main code line.

The ability to plug-in in custom bidding algorithms is vital in order to utilize spot effectively...we're still working through some challenges in this area.

PR #217, opened 2016-08-30
Reduce memory usage in getAmiBlockDeviceMappings method

Jimilian:
Right now plugin tries to describe all images that can be be accessed by user (including public ones). It creates a huge memory consumption in some cases. This PR solves this problem by requesting only image id that should be raised.
screen shot 2016-08-30 at 09 47 27

Also this PR has nice side-effect: new node could be raised much faster (1-3 seconds instead of 20-30 before according our internal metrics).


Jimilian:
@francisu @johnnyshields , can you take a look into my PR?

francisu:
Nice, thanks!

PR #220, opened 2016-09-28
[JENKINS-38481] Fixing compatibility with AWS SDK 1.11.37

AlexejK:
Following up on issue JENKINS-38481, which seemed to cause compatibility issues with AWS-SDK plugin v1.11.37.

Bumping up version and updating signatures of methods in tests.


daniel-beck:
I recommend closing this in favor of #221 which doesn't have unrelated changes, including one file without any relevant change.

francisu:
Thanks for your fix, I took the morally equivalent #221 instead as it did not have unrelated formatting changes.

PR #222, opened 2016-10-06
Support native SSH client for interaction with slave.

neigor:
This will allow chose to use of native ssh/scp to configure slave node. Helps to overcome limitations of java-based ssh client, e.g. launch slaves in different VPC and access them through jumphost (can be configured using system ssh config file)

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2Logger.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2Logger.java
@@ -1,0 +1,28 @@
+package hudson.plugins.ec2.ssh;
+
+import hudson.model.TaskListener;
+import hudson.plugins.ec2.EC2Cloud;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class EC2Logger {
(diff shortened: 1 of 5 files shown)


pellcorp:
I note that the native client does not set the mode of the uploaded file, so it can't be executed upon launch of a job, but otherwise this MR is a life saver for me.

josqu4red:
I'd really like to use that plugin but the lack of jump host support just prevents me to do so. So 👍

msheiny:
@neigor this is an awesome PR :+1: :+1: 

I didnt realize at first that the ssh subprocess setting didn't also pertain to the initial bootstrapping of the node. Anyways, yeah this worked great for me. Crossing my fingers this gets merged in soon

river-fall:
It would be really helpful if this PR was merged. Native SSH plugin does not allow to use the .ssh/config file to use SSH gateways, for example

lillarspillars:
Any plans to merge this change? I know this is an old PR, but it's required for bastion usage since the native SSH option doesn't fully replace the Java SSH/SCP requirement.

PR #224, opened 2016-12-02
Add slave prefix command

imduffy15:
Allow a user to input a slave prefix command that will be executed
before launching the slave jar.


ctran:
Any chance we can get this merged in?

PR #225, opened 2016-12-05
Do not require the ssh key to exist on AWS

imduffy15:
In some cases you may not want a key to exist within AWS.

Examples:

  - The public key is prebaked on your image
  - Your slave is actually a docker container running on an instance and started by user-data

This PR forces the plugin to pull the ssh private key from the private key field rather than enforcing the requirement to have a key on EC2.
 
Signed-off-by: Ian Duffy <ian.duffy@zalando.ie>

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -762,7 +762,7 @@ public abstract class EC2Cloud extends Cloud {
                     EC2PrivateKey pk = new EC2PrivateKey(privateKey);
                     if (pk.find(ec2) == null)
                         return FormValidation
-                                .error("The EC2 key pair private key isn't registered to this EC2 region (fingerprint is "
+                                .warning("The EC2 key pair private key isn't registered to this EC2 region (fingerprint is "
                                         + pk.getFingerprint() + ")");
                 }
 
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -763,7 +763,7 @@ public abstract class EC2Cloud extends Cloud {
                     if (pk.find(ec2) == null)
                         return FormValidation
                                 .warning("The EC2 key pair private key isn't registered to this EC2 region (fingerprint is "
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -762,7 +762,7 @@ protected FormValidation doTestConnection(URL ec2endpoint, boolean useInstancePr
                     EC2PrivateKey pk = new EC2PrivateKey(privateKey);
                     if (pk.find(ec2) == null)
                         return FormValidation
-                                .error("The EC2 key pair private key isn't registered to this EC2 region (fingerprint is "
+                                .warning("The EC2 key pair private key isn't registered to this EC2 region (fingerprint is "
                                         + pk.getFingerprint() + ")");
```
francisu: I would give a hint here about the reason you would not specify the private key since it's no longer required.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -859,10 +863,10 @@ protected EC2SpotSlave newSpotSlave(SpotInstanceRequest sir, String name) throws
     /**
      * Get a KeyPair from the configured information for the slave template
      */
-    private KeyPair getKeyPair(AmazonEC2 ec2) throws IOException, AmazonClientException {
+    private KeyPair getKeyPair(AmazonEC2 ec2) throws IOException {
         KeyPair keyPair = parent.getPrivateKey().find(ec2);
         if (keyPair == null) {
-            throw new AmazonClientException("No matching keypair found on EC2. Is the EC2 private key a valid one?");
+            LOGGER.log(Level.WARNING, "No matching keypair found on EC2. Is the EC2 private key a valid one?");
         }
```
francisu: I would not have this be a WARNING, but INFO and have the text indicate the circumstances where the key is not required, as this could be part of normal operation.

francisu:
Could you also change the help text for the private key field to indicate when it would not be required?

imduffy15:
Sorry for the delay on this @francisu completely missed your review. Changed as a requested.

imduffy15:
Closing this due to the initial delay in a review taking place and failure to follow up.

PR #227, opened 2017-03-07
[JENKINS-42530] Add support of switching role

gboucherie:
- add roleArn and roleSessionName in main config for AmazonEC2Cloud
  - pass roleArn and roleSessionName in all needed methods, except for doFillRegionItems (we don't need to switch role to be able to retrieve regions)
  - externalize ClientConfiguration creation to ba able to reuse it in StsClient creation (or Sts will failed in proxy environment)
  - remove deprecated StaticCredentialsProvider with AWSStaticCredentialsProvider
  - make createCredentialsProvider() protected and Override it in AmazonEC2Cloud

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
@@ -23,6 +23,8 @@
  */
 package hudson.plugins.ec2;
 
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClientBuilder;
 import hudson.Extension;
 import hudson.Util;
 import hudson.model.Failure;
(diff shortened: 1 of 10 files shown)


ebutler-allies:
With the exception of apparently missing help entries on the new fields, this PR is extremely vital to a deployment I'm working towards. Relying on API keys only is not a long term solution due to security implications.

levisbakalinsky:
@ gboucherie thank you for working on this. @francisu, looping you in since your an active committer. Anyway we can get this functionality reviewed and merged? Thank you.

Vlatombe:
This use case is already supported by using https://github.com/jenkinsci/aws-credentials-plugin latest version. When defining an AWS Credential, you can provide a role to assume.

akoi:
@Vlatombe I'm using this together with the [AWS Credentials Plugin](https://github.com/jenkinsci/aws-credentials-plugin), using an assume-role to manage the EC2 slave instances. As you mentioned, this works initially to communicate with EC2 and manage slaves using that role, but as the credentials provided by the STS for assuming the role are only temporary - between 15mins and 1hour (default) - after the credentials expire, the plugin is no longer able to authenticate with AWS, and therefore fails to manage slaves. It looks like the plugin only assigns the credentials on initialisation, and doesn't request new credentials on expiry. 

Manually going to the Jenkins → Manage Jenkins → Configure System UI page, and just saving that page seems to re-initialise the credentials, causing the plugin to work until the new credentials expire. 

Interested to hear if anyone else is experiencing this issue or has a workaround.

tylersmith34:
I'm highly interested in this PR, I have this same problem.  I also am trying to create slaves in a different account, it's not working with the AWS Credentials Plugin.  This PR allows me manage slaves in other accounts and Jenkins can get valid credentials each time makes AWS API calls.  I cannot use IAM users, my company's security policy disallows this.

tylersmith34:
@gboucherie Can you merge the latest changes into your PR?  You have a few conflicts.

EAnushan:
I've created a new fork with the conflicts resolved https://github.com/jenkinsci/ec2-plugin/pull/277

nightswimmings:
@gboucherie if I fill the arn role field with the arn of my role (not the Instance profile ARN), then your feature should make it visible it in EC2 image tab where it says "IAM Role" right? If so, it is not working for me with latest jenkins and plugin

PR #229, opened 2017-04-17
Test

Eriten:


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -510,56 +526,78 @@ private int getPossibleNewSlavesCount(SlaveTemplate template) throws AmazonClien
      * Obtains a slave whose AMI matches the AMI of the given template, and that also has requiredLabel (if requiredLabel is non-null)
      * forceCreateNew specifies that the creation of a new slave is required. Otherwise, an existing matching slave may be re-used
      */
-    private synchronized EC2AbstractSlave getNewOrExistingAvailableSlave(SlaveTemplate template, Label requiredLabel, boolean forceCreateNew) {
+    private EC2AbstractSlave getNewOrExistingAvailableSlave(SlaveTemplate template, Label requiredLabel, boolean forceCreateNew) {
```
Eriten: We need to remove `synchronized` in order to provision nodes in parallel. The consequence is that we might provision more nodes than we want.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -510,56 +526,78 @@ private int getPossibleNewSlavesCount(SlaveTemplate template) throws AmazonClien
...
                 LOGGER.log(Level.WARNING, String.format("Label is null - can't calculate how many executors slave will have. Using %s number of executors", t.getNumExecutors()));
             }
             while (excessWorkload > 0) {
-                final EC2AbstractSlave slave = getNewOrExistingAvailableSlave(t, label, false);
+
                 // Returned null if a new node could not be created
-                if (slave == null)
-                    break;
+//                if (slave == null)
+//                    break;
```
Eriten: Moving slave provision via calling `getNewOrExistingAvailableSlave` within the anonymous call method when adding `PlannedNode`.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -510,56 +526,78 @@ private int getPossibleNewSlavesCount(SlaveTemplate template) throws AmazonClien
...
+//                        long startTime = System.currentTimeMillis(); // fetch starting time
+//                        while ((System.currentTimeMillis() - startTime) < slave.launchTimeout * 1000) {
+//                            return tryToCallSlave(slave, t);
+//                        }
+
+//                        LOGGER.log(Level.WARNING, "Expected - Instance - failed to connect within launch timeout");
+//                        return tryToCallSlave(slave, t);
+                        Jenkins.getInstance().addNode(slave);
+                        slave.toComputer().connect(false).get();
+                        return slave;
```
Eriten: All the `call` method should do is to provision the slave, add the node to Jenkins and return the slave.

PR #230, opened 2017-04-18
Messages class missing?

glnds:
I couldn't get the code to compile locally after a checkout due the fact that the Messages class was missing?
I found the 'missing' class in the latest jar, decompiled it and add it to the source code.


glnds:
My bad, IntelliJ couldn't find the class S-) should have tried a mvn clean compile right away :)

PR #232, opened 2017-05-31
[JENKINS-44560] Provision AWS instance from within a pipeline

adoblas:
Jenkins vsphere plugin allows to invoke some methods to the backend API, as noted on https://github.com/jenkinsci/vsphere-cloud-plugin/blob/master/src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java

To enable aws provisioning (and avoid mantaining a Jenkins agent via the node step), a modification over https://wiki.jenkins-ci.org/display/JENKINS/Amazon+EC2+Plugin is needed. This could be implemented via a new pipeline step such as 

`ec2 cloud: 'AWS', template: 'aws-CentOS-7'
`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,40 @@ THE SOFTWARE.
             <artifactId>httpclient</artifactId>
             <version>4.5.2</version>
         </dependency>
+        <dependency>
+                <groupId>org.jenkins-ci.plugins.workflow</groupId>
+                <artifactId>workflow-cps</artifactId>
+                <version>2.25</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-durable-task-step</artifactId>
+            <version>2.8</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-step-api</artifactId>
+            <version>2.9</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-support</artifactId>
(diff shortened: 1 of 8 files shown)


adoblas:
I'm unable to reproduce the OOM locally.

It has been tested on different JVMs and machines.

adoblas:
I'm unable to reproduce the OOM locally.

It has been tested on different JVMs and machines.

francisu:
Sorry I don't know anything about the Jenkins pipeline support, I have been away from it for a while. I'm concerned about the OOM though. Can you provide some tracing so that when this happens in the batch testing job we might be able to discover what this is. I'm reluctance to merge this until we get a clean test run in Jenkins.

adoblas:
Hi, thanks for answering, I was about to lose the faith, haha.

Since we are not sure about the instance in Jenkins, how about this?
https://ci.jenkins.io/

This change will affect the type of job running in Jenkins, it will be runned in containers so the OOM may no longer be an issue and we wouldn't need to change the pom.xml for that purpose.

i.e: 
https://github.com/jenkinsci/ant-plugin/blob/master/Jenkinsfile

francisu:
I'm not sure what you mean by your previous comment, perhaps you could elaborate?

The build logs that show the OOM exception are here: https://jenkins.ci.cloudbees.com/job/plugins/job/ec2-plugin/511/. We need to get a clean build on the plugin CI build server before I can accept the PR, as if it does not build correctly there, it will not be usable.  

If you think we need more memory allocated to the Junit tests, then you could modify the POM: https://stackoverflow.com/questions/4066424/java-lang-outofmemoryerror-java-heap-space-in-maven. But I think you need to have a good explanation as to why it works locally for you but not in the build server.

adoblas:
:)
Thanks

PR #193, opened 2016-05-03
JENKINS-32915

jjudd:
Cleaning up the PR in #186 so it can be merged and the issue can be closed.

PR #233, opened 2017-06-07
Force EC2 Instance creation (#1)

zcicala:
* The idea behind this PR was to allow the EC2 plugin to aggressively provision slaves when high load is present

* Add configurable flag for EC2 that when enabled forces the creation of a new instance even if one is already being provisioned

* Remove custom version from pom

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@ THE SOFTWARE.
     </parent>
 
     <artifactId>ec2</artifactId>
-    <version>1.37-SNAPSHOT</version>
+    <version>1.39-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Amazon EC2 plugin</name>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/Amazon+EC2+Plugin</url>
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@ THE SOFTWARE.
     </parent>
 
     <artifactId>ec2</artifactId>
-    <version>1.39-SNAPSHOT</version>
+    <version>1.37.1-IPD53026</version>
     <packaging>hpi</packaging>
     <name>Amazon EC2 plugin</name>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/Amazon+EC2+Plugin</url>
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@ THE SOFTWARE.
     </parent>
 
     <artifactId>ec2</artifactId>
-    <version>1.37.1-IPD53026</version>
+    <version>1.38-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Amazon EC2 plugin</name>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/Amazon+EC2+Plugin</url>
@@ -114,7 +114,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>aws-java-sdk</artifactId>
-            <version>1.11.37</version>
+            <version>1.11.119</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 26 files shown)


pom.xml
```diff
@@ -34,7 +34,7 @@ THE SOFTWARE.
     </parent>
 
     <artifactId>ec2</artifactId>
-    <version>1.37-SNAPSHOT</version>
+    <version>1.37.1-IPD53026</version>
```
Jimilian: do not touch this line - it will be handled my release plugin.

rtkkroland:
@Jimilian You can raise hundreds in seconds? My AMI's spin up in about 5 minutes. Since the plug-in does them sequentially, if I get 100 jobs I'm waiting 5 hours. 
I have about 75 jobs. When a common file changes, I'd love for it to spin up 75 nodes immediately, build the jobs and be done.

zcicala:
@Jimilian  My experience was also that it would provision the EC2 instances sequentially. 
https://github.com/jenkinsci/ec2-plugin/blob/5ceed0b35501c8b716028278cda72ea12c934b5a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L581

This loop checks for currently provisioning instances and if one exists it just returns that instance rather than creating a new one. I don't understand this logic, so I instead passed the FORCE_CREATE parameter in

Jimilian:
@rtkkroland, yes. But as I mentioned in https://issues.jenkins-ci.org/browse/JENKINS-25832?focusedCommentId=296698, it was introduced more-or-less recently (after 1.29). I measured maximum throughput in vacuum (up to 200 in 50 seconds), and checked that happens in real case based on statistics we collect (40 nodes in 5 minutes). 

@zcicala, I do not remember code you mentioned, so, I can't comment anything on top of that, but according my experience velocity of raising new instances dependence on AMI as well. It should be lightweight with small "disk", etc.

zcicala:
@Jimilian I can imagine lightweight AMIs leading to fast overall provisioning, but for those of us with heavier images its necessary to provision instances in parallel. This PR allows that to happen.

Jimilian:
@zcicala I'm not a maintainer of this plugin - so, my opinion does not mean much :)
I just found suspicious suggested approach.

If you have a little bit more time, you can take a look in difference between 1.29 and current master. Maybe some new checks prevent raising new nodes faster.

LuisPiedra:
@Jimilian @zcicala Plugin is launching the instances sequentially because of a bug.

Jenkins cloud API sends the excess workload amount to the plugin, and the plugin iterates over this number.

Problem is that after provisioning one instance, the rest of the excess workload is not provisioned because there is already an instance in launching state (the one the plugin has just provisioned!)

I have uploaded a pull request with a fix keeping a list with the instances already provisioned in the same request.

https://github.com/jenkinsci/ec2-plugin/pull/241

Jimilian:
@LuisPiedra, @zcicala, @rtkkroland, finally we updated plugin and now we have the same problem as you do :)

The regression was introduced in https://github.com/jenkinsci/ec2-plugin/pull/193. Looks like the previous state (fire as many instances as you can to get rid of queue) did not work for part of community really well...

LuisPiedra:
@Jimilian If I understand correctly #193, previous state was not "fire as many instances as you can to get rid of queue", but "fire as many instances as you can until the queue is empty" which means launching a huge number of instances even for a single element in queue.

Anyway, the current state is also quite nasty, launching the instances sequentially. Jenkins cloud API provides the plugins a excessWorkload value based on the current demand in the queue, so the plugins should try to provision up to this number, but also take into account how many instances are in launching process. 

I think my #241 respects this excessWorkload, I have been using it for some weeks without problems. In fact, together with AWS now billing by minutes, it has provided us with amazing scaling capabilities based on actual workload.

zcicala:
@LuisPiedra I haven't had a chance to check out your branch, but it sounds like its addressing the same problem. I'm pretty disappointed that this problem remains unaddressed.

Jimilian:
@LuisPiedra If you want to save a little bit more money, you can give a try to https://github.com/heremaps/saturate-nodes-plugin. If you have a lot of nodes, this guy tries to compact all builds.

About the bug...
> If I understand correctly #193
Maybe, never saw this behaviour, but in case of huge queue with short-live builds  additional nodes were raised, yes.
Unfortunately, your fix does not work fast enough for us :(

I agree with @zcicala, that it would be nice to have implementation that fits everyone's needs or, at least, follows the Jenkins Cloud API specs :)

Jimilian:
@rtkkroland, @zcicala, @LuisPiedra Here is my candidate: #252. 

I tested it in our environment (50k jobs per day) and did some stress tests as well (raises up to 200 nodes in 2 minutes). Feel free to give a try - feedback is welcome.

vitalykarasik:
>usually EC2 Plugin was able to raise hundreds of nodes in seconds
Sorry for hijacking this thread, but in my environment, with ver.1.36, it takes about 90 seconds to launch a new instance.

julienduchesne:
@thoulen, This should be fixed with https://github.com/jenkinsci/ec2-plugin/pull/252, right?

PR #237, opened 2017-08-24
update minutes remaining in billing period log entry

mkozell:
The "minutes remaining in billing period" log entry was converting milliseconds to minutes instead of seconds to minutes.  The input was being passed in as seconds.


francisu:
Thanks for your contribution!

PR #239, opened 2017-09-22
master to agent startup optimizations

mkozell:
We use the EC2 plugin and start a large number of Jenkins agents at once.  Frequently we would receive errors starting the agents.

_java.io.IOException: Publickey authentication failed.
Caused by: java.io.IOException: Sorry, this connection is closed._

_ERROR: Unexpected error in launching an agent. This is probably a bug in Jenkins
Caused: hudson.remoting.Channel$OrderlyShutdown
Caused: hudson.remoting.RequestAbortedException_

After further research and debugging, we found the call to get the SSH keypair (_computer.getCloud().getKeyPair()_) was taking several minutes when our Jenkins master was loaded.  Since retrieving the SSH key occurred after the SSH connection was made but before _authenticateWithPublicKey_, the SSH connection timed out.  Perhaps the agent's SSHD LoginGraceTime limit was being reached.

The change in this PR retrieves the SSH key once before any SSH connections are made and uses it for all SSH authentications.

We also saw the _computer.updateInstanceDescription()_ call take several minutes to perform.  We switched to _computer.describeInstance()_ which eliminated this delay.  The docs say it uses a cached state and is safe to use outside of the _getStateCode()_ call which I didn't see being performed here. 

We have implemented these changes into our production Jenkins installations and it greatly reduced our agent startup issues.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
@@ -119,24 +119,23 @@ public class EC2UnixLauncher extends EC2ComputerLauncher {
     @Override
     protected void launch(EC2Computer computer, TaskListener listener, Instance inst) throws IOException,
             AmazonClientException, InterruptedException {
-        final Connection bootstrapConn;
         final Connection conn;
         Connection cleanupConn = null; // java's code path analysis for final
                                        // doesn't work that well.
         boolean successful = false;
+        KeyPair key = computer.getCloud().getKeyPair();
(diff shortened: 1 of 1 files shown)


thoulen:
Can you rebase ?

thoulen:
I am not sure about to switch to computer.describeInstance(), because the it is not a static node that has the ip predefine

mkozell:
> Almost forgot about this. Code looks good, though I haven't done any manual testing.

This is an old PR and can be closed. Maybe someone has the time to take the recommendations and see if they offer performance improvements in the current version of the plugin.

PR #240, opened 2017-10-27
Do not add stopped instances to count

InTheCloudDan:
Currently if you have over provisioned your EC2 instances you will not be able to boot ones that are off. It does this check before calling provision which then looks for stopped instances that can be started.

I was looking to change the logic where it does the getPossibleNewSlavesCount but couldn't see a clean way to do it. By changing the count conditional to only look at the running/pending instances it will avoid the issue.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -354,7 +354,7 @@ public abstract class EC2Cloud extends Cloud {
      *
      * @param template If left null, then all instances are counted.
      */
-    private int countCurrentEC2Slaves(SlaveTemplate template) throws AmazonClientException {
+    private int countCurrentRunningEC2Slaves(SlaveTemplate template) throws AmazonClientException {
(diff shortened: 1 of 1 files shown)


InTheCloudDan:
I'm not sure the error it's throwing is related to my code? If so is it possible to get a clearer error...

```

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project ec2: ExecutionException Error occurred in starting fork, check output in log -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project ec2: ExecutionException Error occurred in starting fork, check output in log
```

InTheCloudDan:
@jvz there wasn't any when I originally wrote it. I'll take a look again this weekend and see if I can find the right spot to add some in.

jvz:
Yeah, this plugin needs more tests, that's for sure. If we can flesh out the tests, then we can release this more often and merge PRs more quickly!

batmat:
@InTheCloudDan this is conflicted FYI, so needs a resolution to make it mergeable.

thoulen:
The fix has been fixed by another patch

PR #248, opened 2017-11-28
JENKINS-32779:

abrys:
Fix instance cap by using the same checks as with regular instances.
Also small fix that currently allows instance_cap + 1 instances to be started


abrys:
Fixes JENKINS-32779

PR #242, opened 2017-11-09
Update from .119 to .226

goodwinb99:
.... to take advantage of the new C5 instance types.
Now covers M5 types and sets default executors based on those types.
The build system is having an intermittent problem starting Surefire :-(  Not sure how to get this to try again.  It builds & passes on my system and I see other builds here are suffering from the same Surefire problem.

PR #250, opened 2017-11-29
Bump aws-java-sdk to pick up c5, m5 instance types

psung:
Hello,

I've applied the following patch (similar to that in #242) to the ec2-plugin in our Jenkins installation so we can use c5 instances. I have a couple of questions though:

1. I saw that #242 failed CI, do you have some idea of whether that's a problem introduced by the version bump / something wedged on the CI environment right now / a spurious failure?
2. What are the values in `toNumExecutors` supposed to be? For the c5 instances I introduced, I've copied numbers from similar instance types (where possible) but I'm not sure I did the right thing.

Thanks in advance for your help!

Phil

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -199,6 +199,8 @@ public abstract class EC2AbstractSlave extends Slave {
             return 7;
         case C4Large:
             return 7;
+        case C5Large:
+            return 7;
         case M1Xlarge:
             return 8;
         case M22xlarge:
@@ -210,6 +212,8 @@ public abstract class EC2AbstractSlave extends Slave {
         case C3Xlarge:
             return 14;
         case C4Xlarge:
+            return 14;
+        case C5Xlarge:
             return 14;
         case C1Xlarge:
             return 20;
@@ -225,6 +229,8 @@ public abstract class EC2AbstractSlave extends Slave {
             return 28;
         case C42xlarge:
             return 28;
+        case C52xlarge:
+            return 28;
         case Cc14xlarge:
             return 33;
         case Cg14xlarge:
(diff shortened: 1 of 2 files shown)


jbapple-cloudera:
There is now also an m5 family.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -199,6 +199,8 @@ public EC2Cloud getCloud() {
             return 7;
         case C4Large:
             return 7;
+        case C5Large:
+            return 7;
```
jbapple-cloudera: What data source is used to infer these numbers?

jaceksokol:
Hey, could you bump it to `1.11.264` to have M5 instances as well?

psung:
@jaceksokol Sure, done.

Meanwhile, can anyone provide guidance on how the `toNumExecutors` numbers are supposed to be chosen? I haven't added any of the m5 types and I don't know whether the numbers I added for c5 are appropriate.

jaceksokol:
Frankly speaking it depends on tasks being executed and imho should be provided explicitly.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
colinbjohnson: I feel like using the AWS Pricing files might be a good method for determining this. One question in my mind is if these values are hard-coded (as they are now) or if a "pricing" API call (https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/pricing/model/package-summary.html) is used to determine the "vcpu" or "ecu" value (and hence, the NumExecutors).

jaceksokol: Make it simple - no API call.

mprince:
@jaceksokol any insight on when is the next release is due?

jaceksokol:
@francisu

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
francisu: @colinbjohnson I would welcome a contribution that uses the API call rather than this hardcoded nonsense that has to be updated all the time.

PR #251, opened 2017-12-08
Add I3.large instance type support

gbastkowski:
Update from .119 to .226


gbastkowski:
My hack of just updating just the dependency of Amazon's SDK was a bit naive. Reverted.

clifford-github:
As commented in JavaDoc instance types are not configured in this location:
```
/*
 * Note this is used only to handle the metadata for older versions of the ec2-plugin. The current
 * versions use the Amazon APIs for this. See PluginImpl for where this is used.
 */
```

PR #254, opened 2017-12-18
Add optional bootDelay parameter for SSH slaves

varju:
Our EC2 Windows slaves were suffering from frequent disconnects, causing any in-progress builds to fail.  Knowing that our Linux slaves were far more stable, we decided to experiment with SSH connections for Windows.

Using a customized daemon (https://github.com/varju/ec2-sshd), we have found that our Windows slaves are now consistently staying connected.  Unfortunately sshd on Windows suffers from the same problem as WinRM, requiring a bit of a delay after startup due to services that restart shortly after boot.

This PR adds a boot delay option to Unix slaves.  Normally the delay will be 0, but this can be manually increased if using a less conventional setup like Windows over SSH.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AMITypeData.java
+++ b/src/main/java/hudson/plugins/ec2/AMITypeData.java
@@ -6,4 +6,6 @@ public abstract class AMITypeData extends AbstractDescribableImpl<AMITypeData> {
     public abstract boolean isWindows();
 
     public abstract boolean isUnix();
+
+    public abstract int getBootDelayInMillis();
 }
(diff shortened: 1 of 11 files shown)


djesionek:
What is, apart from the conflicts, missing here for a merge?

@varju I suppose you probably also have this issue: https://stackoverflow.com/questions/12369090/ec2-windows-instances-initializing-twice 

I noticed that WinRM is more or less stable, however certain builds take more than 2-4x times as long compared to using the SSH connection so this would definitely be a very helpful addition here!

PR #256, opened 2018-01-16
JENKINS-39702: Tag volumes attached to an instance

seelmann:
This change always tags the attached volumes. Allowing a config option via UI obviously would require more changes.


zacblazic:
Keen to see this get in, it's something we've been wanting!

julienduchesne:
LGTM!

PR #257, opened 2018-01-16
JENKINS-48979: Fix race condition when push/fetch tags

seelmann:


mikegron:
I'm having the same issue, it seems related to spot instances. Hopefully it will get merged and released quickly :)

BastienM:
Any updates on this ?
So far, none of the instances (spot, windows) created by the plugin has been tagged ...

zoeimogen:
Seeing the same issue here with intermittent tagging of nodes, only on spot instances.

surry:
Seeing a similar issue where spot instances don't always get tagged. We use tags for billing so it's important that EC2 instances are tagged properly.

@francisu - any thoughts on when this could get merged and made available in a new release?

@reviewbybees

EricKnecht:
I'm also seeing this issue, and also using tags for billing reports. It would be great to have this merged and released.

obazoud:
We have troubles identifying AWS costs with this plugin, this fix would help us a lot.
And PR seems pretty simple to review.

thoulen:
I will release a new plugin version in a couple of weeks

obazoud:
Thanks @thoulen

seelmann:
@thoulen That's great!
Please also consider to include https://github.com/jenkinsci/ec2-plugin/pull/256 which also tags the attaches volumes. This way cost tracking does not only apply to EC2 instances but also to the volumes. We run a custom version with both patches since beginning of the year which works without any problem.

PR #259, opened 2018-01-31
Make fetch time configurable via system property

Jimilian:
`20 seconds` is too often in our case (we have a lot of nodes). So, let's users to decide the best fetch interval.


Jimilian:
@francisu, can you take a look, please?

PR #260, opened 2018-01-31
Add public getter for instanceType

Jimilian:
It would allow to get information about instance type from groovy scripts

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -98,7 +98,7 @@ public abstract class EC2AbstractSlave extends Slave {
     public List<EC2Tag> tags;
     public final String cloudName;
     public AMITypeData amiType;
-    private String instanceType = null;
+    private String instanceType;
 
     // Temporary stuff that is obtained live from EC2
     public transient String publicDNS;
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -23,12 +23,10 @@
  */
 package hudson.plugins.ec2;
 
-import edu.umd.cs.findbugs.annotations.CheckForNull;
 import hudson.Util;
 import hudson.model.Computer;
 import hudson.model.Descriptor;
 import hudson.model.Descriptor.FormException;
-import hudson.model.Hudson;
 import hudson.model.Node;
 import hudson.model.Slave;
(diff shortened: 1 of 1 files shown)


Jimilian:
@francisu, can you take a look, please?

thoulen:
Potential can increase the number of the call to the AWS API

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -152,6 +153,7 @@ public EC2AbstractSlave(String name, String instanceId, String description, Stri
         this.launchTimeout = launchTimeout;
         this.amiType = amiType;
         readResolve();
+        fetchLiveInstanceData(true);
```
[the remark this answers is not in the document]
Jimilian: It's a funny story. Jenkins allows to access object fields from groovy code. Also most of this fields are public and any of them can be used outside of this plugin.
Accessing fields directly gives unique opportunity to reduce the load and latency by not requesting AWS. To make this behavior possible somebody should initialize all this properties first.

Jimilian:
`fetchLiveInstanceData`will be called pretty soon anyway. The calling it in constructor only removes potential "race condition" when somebody smart called object field before it was initialized by `fetchLiveInstanceData`.
Also keep in mind that `fetchLiveInstanceData` throttles number of queries.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -560,6 +564,11 @@ public String getPrivateDNS() {
         return privateDNS;
     }
 
+    public String getInstanceType() {
+        fetchLiveInstanceData(false);
+        return instanceType;
+    }
```
batmat: I do not think this is needed if the goal is for getting it from Groovy scripts.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -98,6 +98,7 @@
     public List<EC2Tag> tags;
     public final String cloudName;
     public AMITypeData amiType;
+    private String instanceType = null;
```
batmat: Per consistency with the above, and readability, initializing variables in Java is unneeded FWIW. What's written above is what happens anyway.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
Jimilian: I think it's better to have it - mainly for the consistency and to make it exposable for other plugins.

PR #261, opened 2018-01-31
Add option to run t2 unlimited instances

kozzztya:
Hello. We faced a problem when t2 instances have too few credits and they can't gain them in the stopped state. Recently AWS added a new feature t2 unlimited that can solve our problem with credits. It would be great to use such feature in the plugin. I'm open to your suggestions what can be improved in this PR.


julienduchesne:
Fixed in #286

PR #266, opened 2018-03-07
Added support for assuming an IAM role

tylersmith34:
This allows for cross account slave creation without requiring IAM users.  The new settings allow a Jenkins instance to provide an IAM role for the master to assume when creating a slave/agent.


tylersmith34:
I missed #227 when I made this, this is essentially a re-do of that PR.  I'm going to close mine.

PR #267, opened 2018-03-07
[JENKINS-47681] Don't count slaves we just provisioned multiple times

marcmac:
When provisioning several on demand slaves, provisionOndemand() can return the same slave multiple times while it is launching.
Keep a list of already provisioned slaves, and exclude them from the available pool when asking for more slaves.


OliverPereira:
I would like to apply this change over PR252 and test it as this PR will probably fix a similar issue we are facing atm.

marcmac:
@OliverPereira I'm using this internally, but have not had time to track down the test failures that blew up my pull request.  I plan to have someone address this in the next few weeks, and can then get this integrated.

PR #270, opened 2018-03-21
JENKINS-38311: Don't start up EC2 slaves that are turned off when Jenkins starts up

datallah:
Currently when jenkins is restarted, all EC2 slaves are started regardless of whether the EC2 instances are currently running.

In the case where there are a large number of slaves, this can have a non-trivial cost.

This change makes it so that when jenkins starts up, EC2 slaves are only started if the EC2 instance is running (or starting to run).

This fixes https://issues.jenkins-ci.org/browse/JENKINS-37552

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -144,7 +144,7 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> {
     }
 
     /**
-     * Called when a new {@link Computer} object is introduced (such as when Hudson started, or when
+     * Called when a new {@link EC2Computer} object is introduced (such as when Hudson started, or when
      * a new agent is added.)
      *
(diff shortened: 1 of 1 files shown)


datallah:
It's unclear to me why the PR-merge check is failing on the Windows host - the error message doesn't have any useful content output.
The tests run successfully for me.

datallah:
Based on a suggestion from the IRC channel, I closed and re-opened the PR to re-trigger the checks and it passed successfully.

ruZZil:
Anyone maintaining this project? This is pretty critical. Every time Jenkins restarts, all our slaves are started up.  This issue is probably raking in a small fortune for Amazon, is that the point?

aabrahamian:
@ruZZil I ended up making a custom build with this change and it has worked amazingly. I recommend doing that. 

If a custom build is not possible, you can do what I used to do before I implemented the fix:  go into AWS and disable the `StartInstances` IAM permission when restarting Jenkins. After it is fully up, you can reenable the permission. It is a pain and certainly the code change is better in every way, but it can help you from spending a small fortune on restart.
Cheers!

PR #272, opened 2018-04-02
Improving debug messaging for WinConnector. 

ke5aux:
Ran into this issue when attempting to troubleshoot connection issues between master and windows slave. Instead of an ICMP packet, the "ping" method opens a TCP socket, and attempts to do an SMB file transfer.


ke5aux:
Something is up with the builder. The code change was a comment string.

PR #252, opened 2017-12-12
JENKINS-25832: Launch multiple slaves in parallel

Jimilian:
This commit changes the behaviour of core functionality: raising new
nodes.

It implements very simple algorithm:
- if NodeProvisioner requested X executors, start to raise
X/N nodes (N is number of executors on each node).
- if there is any orphant nodes (~not known by Jenkins at all), take
them in account (and start if it's neccessary)
- if there is not enough orphant nodes - raise needed nodes in parallel
- give them back to NodeProvisioner in RUNNING state

Also I failed to stop myself from extracting duplicated code to separate methods... sorry, for making review harder.

I attached the last version of master + this PR to this message. So, if someone wants to give a try, it's enough to download it and rename back to `.hpi` (GitHub does not allow `hpi` file extension)
[ec2-1.38-SNAPSHOT-Katysha-Edition.zip](https://github.com/jenkinsci/ec2-plugin/files/1551270/ec2-1.38-SNAPSHOT-Katysha-Edition.zip)

PR #274, opened 2018-04-06
Tag slave instances to their jenkins master

dblanchette:
This replaces PR #271 that had the wrong starting branch.

In the case where you have multiple Jenkins masters using the same configuration (same AMI for multiple clouds), the instance cap is shared between all instances and clouds. The workaround is to add an additional tag with the Jenkins url to have this instance cap unique per Jenkins master.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -81,6 +81,7 @@ import javax.servlet.ServletException;
 
 import hudson.model.TaskListener;
 import jenkins.model.Jenkins;
+import jenkins.model.JenkinsLocationConfiguration;
 
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.HttpResponse;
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -383,6 +389,9 @@ private int countCurrentEC2Slaves(SlaveTemplate template) throws AmazonClientExc
             filters.add(new Filter("launch.image-id", values));
         }
 
+        // The instances must match the jenkins server url
+        filters.add(new Filter("tag:" + EC2Tag.TAG_NAME_JENKINS_SERVER_URL + "=" + jenkinsServerUrl));
```
Jimilian: if user specified `EC2Tag.TAG_NAME_JENKINS_SERVER_URL` manually, this line will break instance cap calculation.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -475,6 +484,15 @@ private int countCurrentEC2Slaves(SlaveTemplate template) throws AmazonClientExc
         return n;
     }
 
+    private boolean isEc2ProvisionedJenkinsSlave(List<Tag> tags, String serverUrl) {
```
Jimilian: imho, it would be better to use `getCustomInstanceTags` here and compare list of tags from instance with expected tags.

Jimilian:
Even so it creates a very big conflict with my PR (https://github.com/jenkinsci/ec2-plugin/pull/252), it would be very good to have something like this in place.

Few things on top of review:
- it would be nice to update instanceCap documentation to refer to the new state.
- jenkins server url can be empty in some installations (it's not mandatory field)

At the end, in current state PR should work if user does not:
- re-use jenkins url
- override `EC2Tag.TAG_NAME_JENKINS_SERVER_URL` on his own

PR #275, opened 2018-04-09
Terminate spot instances properly

thaiphv:
This change ensures that a spot instance is always terminated regardless the result of the cancellation of the spot instance request.


amansukhaniEBSCO:
Hey, I had a question, not related to this PR 
Is there a way to find out when all the builds on an offline slave (Ec2) are finished, and then delete that node automatically? I have seen a configuration in the Jenkins Node Part that says offline a node when there is 1GB of free space left, but it does not delete the node after all the builds on it are finished.

Is there a job I can run to do this? or a Groovy DSL script?

PR #276, opened 2018-04-09
Start agent in cygwin friendly way

Jimilian:
Unfortunately, cygwin and Java are not best friends. So, JVM under cygwin is not
able to resolve '/tmp/slave.jar` properly. To avoid this issue let's
change directory first and execute local file.

After that user can use Windows node in absolutely same way as Unix node.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
@@ -190,9 +190,8 @@ public class EC2UnixLauncher extends EC2ComputerLauncher {
             scp.put(Jenkins.getInstance().getJnlpJars("slave.jar").readFully(), "slave.jar", tmpDir);
 
             String jvmopts = computer.getNode().jvmopts;
-            String prefix = computer.getSlaveCommandPrefix();
-            String launchString = prefix + " java " + (jvmopts != null ? jvmopts : "") + " -jar " + tmpDir + "/slave.jar";
-           // launchString = launchString.trim();
+            String command = "cd " + tmpDir + " && java " + (jvmopts != null ? jvmopts : "") + " -jar slave.jar";
+            String launchString = computer.getSlaveCommandPrefix() + "\"" + command + "\"";
 
             SlaveTemplate slaveTemplate = computer.getSlaveTemplate();
 


Jimilian:
@francisu please, take a look.

jsteinich:
As a workaround, you can set the override temp directory to a windows path.

michelzanini:
I am not 100% sure about this, but since I upgraded to 1.40 version of the plugin with this change, I can't provision Amazon Linux slaves anymore.

I suspect this change is the problem because reverting to 1.39 works and also I tested this:

```
[developer@jenkins-061c194ad17c91e0d ~]$ ssh -i key -p 999 ip-10-10-10-10.eu-west-1.compute.internal java -jar /tmp/slave.jar
<===[JENKINS REMOTING CAPACITY]===>rO0ABXNyABpodWRzb24ucmVtb3RpbmcuQ2FwYWJpbGl0eQAAAAAAAAABAgABSgAEbWFza3hwAAAAAAAAAf4=
```

^ Works when executing `java -jar /tmp/slave.jar`

```
[developer@jenkins-061c194ad17c91e0d ~]$ ssh -i key -p 999 ip-10-10-10-10.eu-west-1.compute.internal cd /tmp && java -jar slave.jar
Error: Unable to access jarfile slave.jar
```

^ Does NOT work when executing `cd /tmp && java -jar slave.jar`

michelzanini:
@thoulen Can we revert this or at least add a check for isUnix().
I am going to create a bug on Jira as blocker.

michelzanini:
By the way, it works if we add a escape char like:

```
[developer@jenkins-061c194ad17c91e0d ~]$ ssh -i key -p 999 ip-10-10-10-10.eu-west-1.compute.internal 'cd /tmp && java -jar slave.jar'
<===[JENKINS REMOTING CAPACITY]===>rO0ABXNyABpodWRzb24ucmVtb3RpbmcuQ2FwYWJpbGl0eQAAAAAAAAABAgABSgAEbWFza3hwAAAAAAAAAf4=
```

Please note the `'` at `'cd /tmp && java -jar slave.jar'`


Please read:
https://unix.stackexchange.com/questions/80821/why-does-cd-command-not-work-via-ssh

thoulen:
Thanks for investigation, please open the ticket.
I am preparing the fix.

Br
FM

michelzanini:
Ticket open https://issues.jenkins-ci.org/browse/JENKINS-53876

michelzanini:
Yes exactly, after java -jar the slave crashes and dies. @jseiser

thaiphv:
I'm not sure if this is the root cause, but after upgrading to 1.40, all of my Ubuntu slaves started using `/tmp` as the workspace root directory. After downgrading the plugin to 1.39, all the slaves used the home directory of the `jenkins` user as the workspace root again.

thoulen:
The change has been reverted, the 1.40.1 is fixing the problem.
But it should not have affected the workspace directory that should be
drive by 'Remote FS root', do you have 'Remote FS root' defined in your
configuration ?

thaiphv:
@thoulen No I don't set 'Remote FS root'. I'll try 1.40.1 anyway.

PR #277, opened 2018-04-12
[JENKINS-42530] Add support of switching role

EAnushan:
Merged changes from master and resolved conflicts.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
@@ -23,6 +23,8 @@
  */
 package hudson.plugins.ec2;
 
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClientBuilder;
 import hudson.Extension;
 import hudson.Util;
 import hudson.model.Failure;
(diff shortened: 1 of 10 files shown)


sestary:
We have tested it and are not having any issues with this MR.

jones2026:
Thank you @EAnushan for fixing these merge conflicts, anything else I can help with to get this merged?

EAnushan:
Done.

stevenbaker:
Some feedback from the developer community at large here now that this change is live ...

After updating to the most recent version of the EC2 plugin, the change in this PR broke some automation that we use to configure our Jenkins instances -- and due to the nature of the change, I suspect we're not alone.

The error was caused by the AmazonEC2Cloud constructor signature changing.

It used to be this:
```
public AmazonEC2Cloud(String cloudName, boolean useInstanceProfileForCredentials, 
                      String credentialsId, String region, String privateKey, 
                      String instanceCapStr, List<? extends SlaveTemplate> templates)
```
 
So the following chunk of Groovy code worked just fine:
```
AmazonEC2Cloud ec2Cloud = new AmazonEC2Cloud(cloudName, useInstanceProfile, 
                          credentialsId, region, privateKey, 
                          instanceCapStr, slaveTemplates)
```

But as part of this PR, commit https://github.com/jenkinsci/ec2-plugin/pull/277/commits/1e0a49d145c7499514c934c57b72a7f5b5762546 introduced a few new Strings to the the AmazonEC2Cloud constructor:
```
public AmazonEC2Cloud(String cloudName, boolean useInstanceProfileForCredentials, 
                      String credentialsId, String region, String privateKey, 
                      String instanceCapStr, List<? extends SlaveTemplate> templates,
                      String roleArn, String roleSessionName)
```

As soon as that one-and-only constructor changed, any Groovy code that calls into this plugin (such as the snippet above) now fails with the following exception:

```
ERROR: Unexpected exception occurred while performing groovy command.
groovy.lang.GroovyRuntimeException: Could not find matching constructor for: 
hudson.plugins.ec2.AmazonEC2Cloud(java.lang.String, java.lang.Boolean, 
java.lang.String, java.lang.String, java.lang.String, 
java.lang.String, java.util.ArrayList)
```

So, thinking through this a bit ...

The new role functionality is nice, but it was ticketed as an Improvement:

https://issues.jenkins-ci.org/browse/JENKINS-42530

Since it's merely an improvement, did this have to be a breaking change?  It seems like this could have been done in a way that was more backward-compatible?  (Or am I fundamentally misunderstanding the nature of this change?) 

For example, note how SlaveTemplate has added new functionality over time while keeping old constructor signatures around for backward compatibility:
https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L227-L279

That style of adding optional functionality is *fantastic* and keeps existing scripts from breaking. 🎉   

In the end, we'll make the changes on our side to our Groovy scripts.  We'll add the new role parameters to the AmazonEC2Cloud constructor to get things working again.  It's all good.

But just telling the story here in the hopes that it doesn't have to happen again ... thanks!

EAnushan:
Sorry about that @stevenbaker. I definitely agree with you. If I had made this change, I'd introduce two constructors to avoid backwards incompatibility.

I wasn't really sure what the extent of these changes were, so I hadn't noticed. I rebased the original authors changes on master since the original author didn't seem to be active anymore.

PR #278, opened 2018-04-20
[FIXED JENKINS-48161]

alexbramham:
Changed synchronized methods from object-level lock to class-level lock.
Fixes a deadlock issue when provisioning across different clouds.


alexanderrtaylor:
@alexbramham looks like you have some conflicting files that need to be adjusted

PR #279, opened 2018-05-04
Fix EC2RetentionStrategy attempt to stop a node on idle that's already disconnected

Ruenzuo:
If the EC2 cloud configuration to stop instances on idle is enabled the retention strategy will attempt to `idleTimeout()` an already disconnected node.

Why is this a problem? The `check` method runs every minute for every build node in a Jenkins instance, this can potentially cause API rate limit for the resource `StopInstances` in the AWS API.


julienduchesne:
Doing this,

julienduchesne:
With this change, offline slaves will never be terminated due to idle timeout

thoulen:
Fixed in the PR-311

PR #281, opened 2018-05-24
JENKINS-51526: Clone the AMI's Root Block Device instead of making a new one from scratch

datallah:
This makes sure that settings like the type (gp2/standard) are not lost


datallah:
Reopening to trigger checks again - seems to be a spurious failure.

datallah:
Reopening again to triggers checks - the Linux checks worked last time, but still not the Windows checks.

datallah:
This is a small fix that is ready to go - is there something I can do to move it forward?

gilbahat:
bump... any good reason this isn't being merged? this is *really* hurting us.

PR #283, opened 2018-05-29
Govcloud Support

ericgerling:
Create a region specific ec2 URL if the region is available. If the plugin is running inside of Govcloud, modified the way tags are applied to support Govcloud restrictions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,6 +216,11 @@ THE SOFTWARE.
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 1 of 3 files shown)


julienduchesne:
@ericgerling. Tag on create is now supported. Is this PR still needed? https://aws.amazon.com/about-aws/whats-new/2018/07/tag-on-create-for-amazon-ec2-and-amazon-ebs-is-now-available-in-the-aws-govcloud-region/

PR #284, opened 2018-05-30
[JENKINS-47681] Don't count slaves we just provisioned multiple times

jonathanjtan:
Rerunning #267 to see if we can isolate the commit in the continuous integration run for debugging purposes. If this isn't best practice/there's a better way to do this, please let me know!


julienduchesne:
Copy of #267

PR #285, opened 2018-06-08
CBR-431: Temporarily Offline EC2 Nodes Shouldn't Get Cleaned Up

awei1001:


awei1001:
Opened too early. Will reopen when ready.

PR #288, opened 2018-06-19
JENKINS-52012: Fix counting of slave caps with spot instances

perrinz:
When checking caps, only count spot instance requests with tag matching our description.


julienduchesne:
The correct JIRA is https://issues.jenkins-ci.org/browse/JENKINS-52021

PR #289, opened 2018-06-22
Bump to latest parent pom

batmat:
@francisu @ndeloof as the current people declared in the permission file as releasers.

cc @reviewbybees for more eyes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@ THE SOFTWARE.
     <parent>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>plugin</artifactId>
-      <version>2.11</version>
+      <version>2.37</version>
     </parent>
 
     <artifactId>ec2</artifactId>
@@ -56,6 +56,7 @@ THE SOFTWARE.
     <properties>
         <powermock.version>1.6.3</powermock.version>
         <jenkins.version>1.651.3</jenkins.version>
+        <java.level>7</java.level>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -124,12 +125,12 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
-            <version>4.5.2</version>
+            <version>4.5.3</version>
```
dwnusbaum: Better to use the [apache-httpcomponents-client-4-api](https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin) plugin instead if possible.

batmat: @dwnusbaum agreed, but here my only goal was to get past the RequireUpperBounds failure. IMO what you're saying is definitely the way to go, but should be a dedicated ticket.

dwnusbaum: @batmat Agreed, I think it makes sense to keep this PR minimal and handle migrating to the API plugin separately.

batmat:
@francisu any possibility of a merge & release?

If you do not have time anymore for this plugin, that is fine too. Just please answer here with a comment, so that we can mark the plugin to be adopted.

Thanks a lot!

pom.xml
oleg-nenashev: Agreed

pom.xml
```diff
@@ -207,14 +208,6 @@ THE SOFTWARE.
...
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
-                </configuration>
-            </plugin>
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>findbugs-maven-plugin</artifactId>
```
oleg-nenashev: can be replaced by a `findbugs.failOnError` property so that there are reports during the build

batmat:
@francisu you've done an awesome work for a long time here. The community is very grateful for your efforts! 
If you cannot anymore work on this, that is fine. This is life and happens to the bests :-). Maybe time will come again :-). 
I am going to mark the plugin as up for adoption then. Thank you again

ndeloof:
not sure why I'm set as a releaser. Definitively not maintaining this plugin

PR #290, opened 2018-06-24
Convert global findbugs skip to specific per class

batmat:
This change is to stop adding more FindBugs "violations" in other classes.

The idea is to also make fixing those more feasible progressively:
one can now remove a single "Match" entry from the exclusion file,
contribute a single fix for the related FindBugs errors, and move on.

I.e. one does not have to fix it all at once when removing "skip".

@francisu as the current maintainer.
cc @reviewbybees for more eyes.

Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -219,7 +219,7 @@ THE SOFTWARE.
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>findbugs-maven-plugin</artifactId>
                 <configuration>
-                    <skip>true</skip>
+                    <excludeFilterFile>findbugs-exclude.xml</excludeFilterFile>
                 </configuration>
             </plugin>
         </plugins>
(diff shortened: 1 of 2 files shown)


batmat:
I actually did create https://issues.jenkins-ci.org/browse/JENKINS-52208 a few minutes ago.

pom.xml
```diff
@@ -219,7 +219,7 @@ THE SOFTWARE.
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>findbugs-maven-plugin</artifactId>
                 <configuration>
-                    <skip>true</skip>
+                    <excludeFilterFile>findbugs-exclude.xml</excludeFilterFile>
```
oleg-nenashev: See https://github.com/jenkinsci/ec2-plugin/pull/289#discussion_r198444287

PR #293, opened 2018-07-03
[JENKINS-52319] [JENKINS-52334] compatibility with configuration-as-code

ndeloof:


thoulen:
Can you create a util function for that to replace that 👍   
this.mode = mode != null ? mode : Node.Mode.NORMAL;

like is done for the :
this.labels = Util.fixNull(labelString);

batmat:
Issue is: addition to the core is going to anyway take quite some time to percolate until here.

`org.apache.commons.lang3.ObjectUtils#defaultIfNull` ?

ndeloof:
ec2 plugin depends on an old core version which doesn't include  commons-lang3, so this would result in adding an explicit cast. Not sure this offers better readability vs the proposed ternary operation

batmat:
Yes, my bad, meant to post the same or only once as I did in https://github.com/jenkinsci/jenkins/pull/3656#issuecomment-425432744. This method is also in`commons-lang` AFAICT `org.apache.commons.lang.ObjectUtils#defaultIfNull`.

ndeloof:
right, but do we prefer:
```java
this.mode = (Node.Mode) ObjectUtils.defaultIfNull( mode, Node.Mode.NORMAL );
```
over:
```java
this.mode = mode != null ? mode : Node.Mode.NORMAL;
```

it makes more sense on java 8 thanks to static import and parameterized method return type
```java
this.mode = defaultIfNull( mode, Node.Mode.NORMAL );
```
... but as I said I don't plan to change ec2-plugin platform requirements in this PR

PR #294, opened 2018-07-05
introduce a ProvisionerStrategy to provision a node without delay

ndeloof:
the standard `NodeProvisioner.Strategy` has been implemented to avoid over-provisioning, mostly designed at a time EC2 instances where billed by hours, which means we didn't wanted to get an extra one while jenkins might soon have available executors.
As EC2 is now billed by the minute we can just provision a new node when we miss one without delay, in worse case it will stay for few minutes then be terminated.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
@@ -42,6 +42,7 @@ import jenkins.model.Jenkins;
 
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerResponse;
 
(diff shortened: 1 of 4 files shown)


steven-foster:
Thanks! this + #252 could be a great combo, since #252 does not play well with the default provisionerstrategy at all

ydubreuil:
Another reason to slow down provisioning back then was that boot time was quite high, up to few minutes. On tests I did with latest m5 instances, I can have a ready to serve build agent in just 25 seconds! Risks of over-provisioning is really low now.

ndeloof:
I can't get why CI is failing
`Tests run: 59, Failures: 0, Errors: 0`
but `SurefireBooterForkException : Error occurred in starting fork, check output in log`

dwnusbaum:
@ndeloof I've seen that error intermittently for this plugin's CI builds (check some older builds of master), but I was never able to reproduce it locally on my Windows machine. Maybe something is flaky with the dependencies and is fixed by #289?

francisu:
The windows build on CI has been bad for months now, and it just started happening one day for no reason that I could find. It worked locally for me, so I suspect this is some kind of environment issue. The Linux build has always been fine, and so that's what I used.

OliverPereira:
I combined this PR with #252 and it is working as expected.

No over provisioning of slaves now.

ndeloof:
I'm not even sure #252 is required with those changes as ec2 cloud won't wait long before it starts new nodes, so the benefit for #252 is pretty limited (not saying it is wrong)

dhallam:
Is there any update on a release for #294 (with or without #252) ?

thoulen:
I am going to prepare the release this week and I will merge it,  I am already using the PR-252, that results already in most of the cases over provisioning. I believe the PR-294 could results in more  over provisioning and hit some AWS limits (API calls), but with the option flag to turn on and off the feature and the CAP the people should be able to avoid problems.

PR #295, opened 2018-07-23
[JENKINS-35708] Round-robin subnet selection

DavidFluck:
[JENKINS-35708]

Allow users to supply multiple subnets for launching EC2 instances into
instead of just one. This solves the issue where you need to launch a
large amount of somewhat less available instances (e.g. GPU instances)
and need to use multiple availability zones to avoid exhausting that
instance type.

I tried to avoid breaking any backwards compatibility. I slightly
modified the behavior of the subnet ID configuration field: if you
provide a space-separated list of subnet IDs, the code will cycle
through them in order beginning with the first one. If you have one ID
it will still technically cycle through them, but the end result is that
you only launch instances into that single subnet as before, preserving
the original behavior.

I also added help text for the subnet ID field.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -537,7 +537,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
             if (getAssociatePublicIp()) {
                 net.setSubnetId(subnetId);
             } else {
-                reRequest.setSubnetId(subnetId);
+                riRequest.setSubnetId(subnetId);
             }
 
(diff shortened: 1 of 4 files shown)


jvz:
Can you rebase this from master? This plugin got some new maintainers recently and we're looking to clear up the backlog. Thanks!

DavidFluck:
Absolutely!

DavidFluck:
I've also confirmed that after the rebase the plugin changes still work as intended.

DavidFluck:
Ah yes. I'm looking at the tests and there should be a way to create three spot and on-demand instances and make sure their subnets are cycling through properly. I can do that today or tomorrow.

PR #301, opened 2018-10-01
Fix some findbugs bugs

julienduchesne:


src/main/java/hudson/plugins/ec2/win/WinConnection.java
```diff
@@ -119,6 +119,10 @@ private static String urlEncode(String value) {
         }
     }
 
+    private static void setLogStreamLevel(int level) {
+        LogStream.level = 6;
```
steven-foster: hardcoded value instead of argument? :)

PR #303, opened 2018-10-02
Upgrade baseline versions of Jenkins, Java, Apache HttpClient, and BouncyCastle

jvz:
Instead of relying directly on a common dependency, this switches to the
Jenkins plugin that bundles HttpClient et al. in an easily upgradeable
Jenkins plugin. This also upgrades the base level of Java required to
1.8 as dependencies (including Jenkins) now require it.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,7 +112,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>bouncycastle-api</artifactId>
-            <version>1.648.3</version>
+            <version>2.17</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -58,7 +58,7 @@ THE SOFTWARE.
         <changelist>-SNAPSHOT</changelist>
         <powermock.version>1.6.3</powermock.version>
         <jenkins.version>1.651.3</jenkins.version>
-        <java.level>7</java.level>
+        <java.level>8</java.level>
```
dwnusbaum: If you require Java 8 you should bump the Jenkins baseline to 2.60.3, because the 2.60 LTS line is the first version that requires Java 8.

dwnusbaum: Beat @oleg-nenashev by one second 😄

oleg-nenashev: I was reviewing from phone, so sorry for being too slow :D

PR #305, opened 2018-10-02
Update to AWS SDK 1.11.403

jvz:
@reviewbybees


julienduchesne:
We could add M5 types like here: https://github.com/jenkinsci/ec2-plugin/pull/242/files and close it

PR #310, opened 2018-10-16
Tag slave instances to their jenkins master

julienduchesne:
Replaces https://github.com/jenkinsci/ec2-plugin/pull/274 (from my company)

In the case where you have multiple Jenkins masters using the same configuration (same AMI for multiple clouds), the instance cap is shared between all instances and clouds. The workaround is to add an additional tag with the Jenkins url to have this instance cap unique per Jenkins master.


imod:
@julienduchesne could this be the reason for: https://issues.jenkins-ci.org/browse/JENKINS-54266 ?

julienduchesne:
Seems like it. Using the instance identity's public key would be better indeed IMO

batmat:
Fix proposed as https://github.com/jenkinsci/ec2-plugin/pull/314

batmat:
Apart from this, I think it would be great for next times if a JIRA entry can be requested for non trivial changes, so that we can a clear problem definition upstream to refer to.
Here for instance, I _think_ I get it from the PR description, but I am not totally sure. Thanks!

PR #311, opened 2018-10-16
FIX JENKINS-54071

thoulen:
FIX JENKINS-54071

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -320,7 +320,7 @@ public abstract class EC2Cloud extends Cloud {
         StringWriter sw = new StringWriter();
         StreamTaskListener listener = new StreamTaskListener(sw);
         EC2AbstractSlave node = t.attach(id, listener);
-        Jenkins.getActiveInstance().addNode(node);
+        Jenkins.getInstance().addNode(node);
 
         rsp.sendRedirect2(req.getContextPath() + "/computer/" + node.getNodeName());
     }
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -320,7 +320,7 @@ public void doAttach(StaplerRequest req, StaplerResponse rsp, @QueryParameter St
         StringWriter sw = new StringWriter();
         StreamTaskListener listener = new StreamTaskListener(sw);
         EC2AbstractSlave node = t.attach(id, listener);
-        Jenkins.getActiveInstance().addNode(node);
+        Jenkins.getInstance().addNode(node);
```
jvz: Should use `Jenkins.get()` to avoid NPE.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -643,10 +646,10 @@ private void wakeOrphansUp(AmazonEC2 ec2, List<Instance> orphans) {
 
         if (!instances.isEmpty()) {
             StartInstancesRequest siRequest = new StartInstancesRequest(instances);
-            StartInstancesResult siResult = ec2.startInstances(siRequest);
-
+            StartInstancesResult  siResult = ec2.startInstances(siRequest);
```
julienduchesne: added whitespace

src/main/java/hudson/plugins/ec2/EC2Cloud.java
thoulen: At the moment the code is mainly using the getInstance(), and the getActiveInstance() is deprecated. If we want to move to jenkins.get() we should handle in a separate "global clean up"commit

thoulen: yes, I am going to submit the patch for the problem reported by the findbugs

jvz: It's amusing because `Jenkins.getInstance()` is a nullable method, but it wasn't an issue until they broke up the methods.

PR #314, opened 2018-10-26
[JENKINS-54266] Warn if no server URL and default to instance-identity instead

batmat:
https://issues.jenkins-ci.org/browse/JENKINS-54266

(Also filed https://github.com/jenkinsci/instance-identity-module/pull/13 but cannot use it for now at it needs to be merged, released, and become embedded in the minimum Jenkins baseline, hence in a pretty far future.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,6 +130,12 @@ THE SOFTWARE.
             <version>4.5.5-3.0</version>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.modules</groupId>
+            <artifactId>instance-identity</artifactId>
+            <version>2.2</version>
+            <scope>provided</scope><!-- comes from the core -->
+        </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


batmat:
FindBugs is drunk AFAICT, fixing the build still, a green build is not just a nice to have.

thoulen:
I would rather not put tags if not available.. but is fine

batmat:
I think I overlooked the fact this was not the only location to change: https://github.com/jenkinsci/ec2-plugin/commit/701758bfb9eb238f762d4e15df1e0941492def25#diff-f2115e33148d3db7c133fe014ad9dfddR505

Anyway, going to propose a fix, and this time I will double-check I do not see the issue anymore. 
Not such a big issue, though, in probably 99% of cases since Jenkins by default sets the server url by Javascript when opening /configure, so any instance out there where admins did save at least once there won't ever hit this, but well it's just not good for reputation in general ;-).

PR #315, opened 2018-10-27
Reincrementalify and update to latest parent pom

batmat:
This will get the [>= 3.18 parent pom](https://github.com/jenkinsci/plugin-pom/blob/master/CHANGELOG.md#318) so that the pom.xml is automatically _reincrementalified_ on normal releases and this way the EC2 plugin still gets its incremental releases published at http://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/ec2/ as expected :)

I had filed https://issues.jenkins-ci.org/browse/INFRA-1825 before realizing the issue might be on the EC2 plugin side. Hence this PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,11 +30,11 @@ THE SOFTWARE.
     <parent>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>plugin</artifactId>
-      <version>3.15</version>
+      <version>3.25</version>
     </parent>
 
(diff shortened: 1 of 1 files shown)


batmat:
cool, this seems to fix the issue, as the sha of this PR got published:
http://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/ec2/1.42-rc793.29f87d3fb1d2/

PR #317, opened 2018-10-29
FIX-JENKINS-54315 AWS API's eventual consistency

thoulen:


src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -524,7 +524,7 @@ protected void clearLiveInstancedata() throws AmazonClientException {
      * clearLiveInstancedata if needed
      */
     protected void pushLiveInstancedata() throws AmazonClientException {
-        Instance inst = CloudHelper.getInstance(getInstanceId(), getCloud());
+        Instance inst = CloudHelper.getInstanceWithRetry(getInstanceId(), getCloud());
```
jvz: Looks like you'll need to propagate or handle the `InterruptedException` here and elsewhere.

PR #318, opened 2018-11-05
[JENKINS-53858] Remove synchronized from EC2Cloud.connect() to fix deadlock

r4v5:
There are multiple codepaths that can call connect(). Some codepaths
have already acquired the hudson.model.Queue lock, and others don't, so
the synchronized keyword here can lead to deadlocks where one thread is
holding the Queue lock and trying to lock the EC2Cloud, and another
thread is holding the EC2Cloud lock and trying to lock the Queue.

With the keyword removed, there's the possibility of multiple EC2
connections occurring, but I've yet to see this cause any observable
problems in reality, and the deadlock issue seems to have stopped
entirely.

We also move to a single purpose-built ReentrantLock per EC2Cloud
instance, since that codepath was one of the most common holders of the
EC2Cloud lock.

cc @thoulen

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -59,6 +59,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


thoulen:
I am going to stress test the patch, in the past we have several problem on double connection. 
Eventually we can handle with lock file in the node, as was already proposed some time ago.

r4v5:
Thanks. I originally tried using a lock like the one around the instance-counting on the connect, but on a Jenkins restart it would presumably deserialize the class instead of calling the constructor, so the lock was null, and I couldn’t figure out a way to deserialize a lock in a threadsafe manner without...some other lock anyway.

What are the problems if there are duplicate connections hanging around? We have a staging environment where I’m testing this change, I just don’t know how that would manifest and would like to know what to look for.

thoulen:
Jenkins will attempt to connect twice and one of the connections will end up in a exception.
After some stress seems fine, I will prepare a snapshot for more extended testing

thoulen:
Can you check the ticket:
https://issues.jenkins-ci.org/browse/JENKINS-53858
we some problem (I don't know why i didn't see in my env)

r4v5:
I'm also not sure why it didn't cause problems in mine, but I can put up a PR with the fix.

PR #319, opened 2018-11-07
Do not tag instances if no serverUrl

batmat:
https://issues.jenkins-ci.org/browse/JENKINS-54266

This is probably more sensible than my previous attempt on #314, and in the end matches what [@thoulen was saying would probably have been preferrable](https://github.com/jenkinsci/ec2-plugin/pull/314#issuecomment-433432132): we just do not tag in case there's no serverUrl.

It's also a followup of #310, for inter-tracking purpose.

*Anyway* the number of production instances out there without a `serverUrl` is probably close to 0, but well.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -46,8 +46,6 @@ import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
-import java.nio.charset.Charset;
-import java.security.interfaces.RSAPublicKey;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


batmat:
Tested by building a local package and manual upload. Going to do a second check on a freshly provisioned setup as soon as ci.jenkins.io builds the incremental release for this PR.

batmat:
Just confirmed this works on a fresh instance. Sorry for the mishaps on #314.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -377,8 +367,8 @@ private int countCurrentEC2Slaves(SlaveTemplate template) throws AmazonClientExc
             jenkinsServerUrl = jenkinsLocation.getUrl();
 
         if (jenkinsServerUrl == null) {
-            LOGGER.log(Level.WARNING, "No Jenkins server URL specified, using instance-identity instead");
-            jenkinsServerUrl = getInstanceIdentityEncodedPublicKey();
```
batmat: not trying anymore to be smart for no good reason, just warn and **not** crash if serverUrl is unspecified.

imod:
any chance we can get a release with this soon?

thoulen:
We should release in the beginning of next week, i have two pending task the pr 321 and fix the CAP

PR #321, opened 2018-11-08
JENKINS-53858: Fix NPE on startup due to uninitialized lock

r4v5:
Also catches NPEs on startup from the node health checking thread firing
before everything's set up, which was what I saw on my production
deployment. Oddly, my staging environment was satisfied with just #318.


Cervator:
Heya all - any chance this PR (and the older related one) makes it into a release soonish? :-)

PR #322, opened 2018-11-09
Spot slave readiness

thomas-brx:
This PR solves two issues with spot requests:

1) The plugin would sometimes get confused and return an instance-id for a completely different spot instance request. This was due to `spotInstanceRequestId` not being set before `getInstanceId` was called.
2) The plugin would sometimes report `INFO: Launching instance: (null)`, this would happen when the spot instance request was still pending. This in turn lead to unhandled exceptions when it could not resolve the private/public IP/name of the null-instance

The solution for the first issue is to extract the two methods `readResolve` and `fetchLiveInstanceData` from the constructor of `EC2AbstractSlave`, to a (fluent) init-method instead. This guarantees that `spotInstanceRequestId` is set in the `EC2SpotSlave` object when `getInstanceId` is first called (it's called in `fetchLiveInstanceData`).
This has minor impact on the `SlaveTemplate`, the factory method just has to add `.init()` after the `new EC2...()`

The solution for the second issue is to implement a `EC2Readiness` interface. This `EC2SpotSlave` implements the interface, and will signal `ready` when an instance is actually available. The interface also has a method for fetching the current status, which is set to the status of the spot instance request.

The `EC2UnixLauncher` is responsible for the handling of readiness.

This is the output in the log when starting a new spot instance (notice the first 2 INFO lines):

```
Nov 09, 2018 2:33:56 PM hudson.plugins.ec2.EC2Cloud
INFO: Node still not ready. Current status: Your Spot request has been submitted for review, and is pending evaluation.
Nov 09, 2018 2:33:57 PM hudson.plugins.ec2.EC2Cloud
INFO: Node still not ready. Current status: Your Spot request is pending fulfillment.
Nov 09, 2018 2:33:58 PM hudson.plugins.ec2.EC2Cloud
INFO: Launching instance: i-0e13beef03ba3aeb3
Nov 09, 2018 2:33:58 PM hudson.plugins.ec2.EC2Cloud
INFO: bootstrap()
Nov 09, 2018 2:33:58 PM hudson.plugins.ec2.EC2Cloud
INFO: Getting keypair...
Nov 09, 2018 2:33:58 PM hudson.plugins.ec2.EC2Cloud
INFO: Using key: jenkins
```


tommywo:
I've tested the fix and it works. Thanks @thomas-brx

nullify005:
This is really really needed in order to stop Jenkins from creating more spot requests than it really needs to get an instance up & running.

It's a must have

PR #223, opened 2016-11-29
Support maximum builds on slave

ariaviran:
Added support to set a maximum number of builds on a slave before it is
terminated in a setting called 'Maximum Total Uses'.

Resolves JENKINS-8618
Related to PR #162

PR #323, opened 2018-11-11
[JENKINS-8618] Support maximum builds on slave

autogun:
This is a rebased and conflict-free PR based on @ariaviran's original [PR](https://github.com/jenkinsci/ec2-plugin/pull/223) to support maximum builds on slave.

This code is running in our environment for more than a year (based on ver. 1.37 of the plugin) and it's doing very good job.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -97,6 +97,7 @@ public abstract class EC2AbstractSlave extends Slave {
     public List<EC2Tag> tags;
     public final String cloudName;
     public AMITypeData amiType;
+    public String maxTotalUses;
 
     // Temporary stuff that is obtained live from EC2
     public transient String publicDNS;
@@ -126,7 +127,7 @@ public abstract class EC2AbstractSlave extends Slave {
 
     public static final String TEST_ZONE = "testZone";
 
-    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, String maxTotalUses)
             throws FormException, IOException {
 
         super(name, "", remoteFS, numExecutors, mode, labelString, launcher, retentionStrategy, nodeProperties);
@@ -145,6 +146,7 @@ public abstract class EC2AbstractSlave extends Slave {
         this.cloudName = cloudName;
         this.launchTimeout = launchTimeout;
         this.amiType = amiType;
+        this.maxTotalUses = maxTotalUses;
         readResolve();
     }
 
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -96,7 +96,7 @@ public abstract class EC2AbstractSlave extends Slave {
     public List<EC2Tag> tags;
     public final String cloudName;
     public AMITypeData amiType;
-    public String maxTotalUses;
+    public int maxTotalUses;
     private String instanceType;
 
     // Temporary stuff that is obtained live from EC2
@@ -134,10 +134,10 @@ public abstract class EC2AbstractSlave extends Slave {
 
     public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
             throws FormException, IOException {
-        this(name, instanceId, description, remoteFS, numExecutors, mode, labelString, launcher, retentionStrategy, initScript, tmpDir, nodeProperties, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, tags, cloudName, usePrivateDnsName, useDedicatedTenancy, launchTimeout, amiType, "-1");
+        this(name, instanceId, description, remoteFS, numExecutors, mode, labelString, launcher, retentionStrategy, initScript, tmpDir, nodeProperties, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, tags, cloudName, usePrivateDnsName, useDedicatedTenancy, launchTimeout, amiType, -1);
     }
 
-    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, String maxTotalUses)
+    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, int maxTotalUses)
             throws FormException, IOException {
 
         super(name, remoteFS, launcher);
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -132,6 +132,11 @@ public abstract class EC2AbstractSlave extends Slave {
 
     public static final String TEST_ZONE = "testZone";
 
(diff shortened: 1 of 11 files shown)


src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -96,6 +96,7 @@
     public List<EC2Tag> tags;
     public final String cloudName;
     public AMITypeData amiType;
+    public String maxTotalUses;
```
jvz: Shouldn't this be an int?

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -131,7 +132,7 @@
 
     public static final String TEST_ZONE = "testZone";
 
-    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, String maxTotalUses)
```
jvz: In order to preserve backward compatibility, I think you should make a new constructor that this old one can delegate to.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -203,4 +206,31 @@ protected Object readResolve() {
...
+    }
+
+    public void taskCompletedWithProblems(Executor executor, Queue.Task task, long durationMS, Throwable problems) {
+        postJobAction(executor);
+    }
+
+    private void postJobAction(Executor executor){
+        EC2Computer computer = (EC2Computer) executor.getOwner();
+        EC2AbstractSlave slaveNode = computer.getNode();
+        int maxTotalUses = Integer.parseInt(slaveNode.maxTotalUses);
```
jvz: What should happen here if it's not a valid integer? If you use a field type of int (or Integer), Stapler should take care of conversions and error messages I think.

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -33,20 +33,20 @@
...
     }
 
     @DataBoundConstructor
-    public EC2OndemandSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2OndemandSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, int maxTotalUses)
```
jvz: Sorry, forgot to mention this class, too, should probably have a backward compatible constructor.

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -32,17 +32,17 @@
...
+        this(description + " (" + name + ")", spotInstanceRequestId, description, remoteFS, numExecutors, mode, initScript, tmpDir, labelString, Collections.emptyList(), remoteAdmin, jvmopts, idleTerminationMinutes, tags, cloudName, usePrivateDnsName, launchTimeout, amiType, maxTotalUses);
     }
 
     @DataBoundConstructor
-    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, int launchTimeout, AMITypeData amiType)
+    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, int launchTimeout, AMITypeData amiType, int maxTotalUses)
```
jvz: Looks like this class also needs a backward compatible constructor update.

awiddersheim:
I'm interested in making use of this feature but looking at the implementation I have some concerns. I guess the biggest of which is nothing immediately pops out at me that would prevent an instance from terminating or stopping when there are still active builds running. Perhaps I'm missing something though.

There is this call `computer.setAcceptingTasks(false);` but the instance is immediately terminated after that without waiting any amount of time for any currently running jobs to complete.

Also, since the max number of builds is determined _after_ jobs have completed, it seems possible that more than the configured number of builds may run.

Take for example the following scenario. I've configured `100` max builds for this executor and I'm at build `99`. The job `Foo` runs and takes `20` seconds followed immediately by job `Bar` running which only takes `10` seconds. Presumably, the job `Foo` would just be killed off when the instance is terminated.

Maybe it would make more sense to do the counting of max builds in `taskAccepted()`. Once the max has been reached, `computer.setAcceptingTasks(false)` can be called to no longer accept any new jobs. Then in the `postTask*` stages, if max has been reached, wait for the last build to complete and terminate. Perhaps with something like [this](https://javadoc.jenkins.io/hudson/model/Computer.html#countBusy--).

autogun:
@awiddersheim You're totally right. This is exactly what happens.
This never bitten us for the sole reason we set our agents with # of executors set to `1` and therefore overlooked.

See my last commit, following your suggestion, which fixes this.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -218,17 +230,15 @@ public void taskCompletedWithProblems(Executor executor, Queue.Task task, long d
...
         EC2Computer computer = (EC2Computer) executor.getOwner();
         EC2AbstractSlave slaveNode = computer.getNode();
-        int maxTotalUses = slaveNode.maxTotalUses;
-        if (maxTotalUses <= -1) {
-            return;
-        } else if (maxTotalUses <= 1) {
-            computer.setAcceptingTasks(false);
-            slaveNode.terminate();
+        // At this point, if agent is in suspended state and has 1 last executer running, it is safe to terminate.
+	    if (computer.countBusy() <= 1 && !computer.isAcceptingTasks()) {
```
awiddersheim: nit: Indentation looks wrong here.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -218,17 +230,15 @@ public void taskCompletedWithProblems(Executor executor, Queue.Task task, long d
...
+            LOGGER.info("Agent " + slaveNode.instanceId + " is terminated due to maxTotalUses (" + slaveNode.maxTotalUses + ")");
+	        slaveNode.terminate();
         } else {
-            slaveNode.maxTotalUses = slaveNode.maxTotalUses - 1;
+            LOGGER.info("Agent " + slaveNode.instanceId + " is still in use by more than one (" + computer.countBusy() + ") executers");
```
awiddersheim: This will log after _every_ task. May need to guard it better so it only emits when applicable.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -207,7 +207,19 @@ protected Object readResolve() {
     }
 
     public void taskAccepted(Executor executor, Queue.Task task) {
-        return;
+        EC2Computer computer = (EC2Computer) executor.getOwner();
+        EC2AbstractSlave slaveNode = computer.getNode();
+        int maxTotalUses = slaveNode.maxTotalUses;
+        if (maxTotalUses <= -1) {
+            LOGGER.info("maxTotalUses set to unlimited (" + slaveNode.maxTotalUses + ") for agent " + slaveNode.instanceId);
```
awiddersheim: I think it's debatable that this log message is of any use. At least at the `INFO` level.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -207,7 +207,19 @@ protected Object readResolve() {
...
+            return;
+        } else if (maxTotalUses <= 1) {
+            LOGGER.info("maxTotalUses drained - suspending agent " + slaveNode.instanceId);
+            computer.setAcceptingTasks(false);
+        } else {
+            slaveNode.maxTotalUses = slaveNode.maxTotalUses - 1;
+            LOGGER.info("maxTotalUses set to " + slaveNode.maxTotalUses + " for agent " + slaveNode.instanceId);
```
awiddersheim: This message is slight misleading in my opinion. Instead of `maxTotalUses set to`, maybe something like `slaveNode.instanceId + "  has " + slaveNode.maxTotalUses + " left"`

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
autogun: Adding bunch of those loggers basically helps during the tests... I don't mind removing it tho.
```
0.361 [id=111]       INFO    jenkins.InitReactorRunner$1#onAttained: Completed initialization
   0.369 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#taskAccepted: maxTotalUsesInstance has 3 builds left
   0.369 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#taskAccepted: maxTotalUsesInstance has 2 builds left
   0.370 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#taskAccepted: maxTotalUsesInstance has 1 builds left
   0.370 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#taskAccepted: maxTotalUses drained - suspending agent maxTotalUsesInstance
   0.370 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#postJobAction: Agent maxTotalUsesInstance is terminated due to maxTotalUses (1)
   0.370 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#taskAccepted: maxTotalUses drained - suspending agent maxTotalUsesInstance
   0.370 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#postJobAction: Agent maxTotalUsesInstance is terminated due to maxTotalUses (0)
   0.371 [id=94]        INFO    h.p.ec2.EC2RetentionStrategy#taskAccepted: maxTotalUses set to unlimited (-1) for agent maxTotalUsesInstance
   0.372 [id=94]        INFO    jenkins.model.Jenkins#cleanUp: Stopping Jenkins
```

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
autogun: Something like this?
```if (maxTotalUses <= 1) {
    LOGGER.info("Agent " + slaveNode.instanceId + " is still in use by more than one (" + computer.countBusy() + ") executers");
}```

awiddersheim: Yeah, I would only want to see this message when an instance is slated to be removed and the `maxTotalUses` feature is in use.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
awiddersheim: Right, I'd just adjust to `LOGGER.debug`.

stefanverhoeff:
@awiddersheim @thoulen this change would be very useful. What is stopping it from getting merged?

awiddersheim:
I'm just some guy without commit access to this repo. 🤷‍♂️

jvz:
No negative feedback over time, so I'm going to merge this. See related issue: https://issues.jenkins-ci.org/browse/JENKINS-8618

stefanverhoeff:
Thanks @jvz + @awiddersheim :-).

PR #324, opened 2018-11-13
FIX JENKINS-54041

thoulen:
Exclude stopped instance from the count of the CAP, the stopped instances are the first recovered, no problem for node leek

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -379,13 +379,16 @@ public abstract class EC2Cloud extends Cloud {
         Set<String> instanceIds = new HashSet<String>();
         String description = template != null ? template.description : null;
 
+        //FIXME convert to a filter query
         for (Reservation r : connect().describeInstances().getReservations()) {
             for (Instance i : r.getInstances()) {
                 if (isEc2ProvisionedAmiSlave(i.getTags(), description)
                     && isEc2ProvisionedJenkinsSlave(i.getTags(), jenkinsServerUrl)
                     && (template == null || template.getAmi().equals(i.getImageId()))) {
                     InstanceStateName stateName = InstanceStateName.fromValue(i.getState().getName());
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -379,13 +379,16 @@ private int countCurrentEC2Slaves(SlaveTemplate template) throws AmazonClientExc
...
         for (Reservation r : connect().describeInstances().getReservations()) {
             for (Instance i : r.getInstances()) {
                 if (isEc2ProvisionedAmiSlave(i.getTags(), description)
                     && isEc2ProvisionedJenkinsSlave(i.getTags(), jenkinsServerUrl)
                     && (template == null || template.getAmi().equals(i.getImageId()))) {
                     InstanceStateName stateName = InstanceStateName.fromValue(i.getState().getName());
-                    if (stateName != InstanceStateName.Terminated && stateName != InstanceStateName.ShuttingDown) {
+                    if (stateName != InstanceStateName.Terminated && 
+                        stateName != InstanceStateName.ShuttingDown && 
+                        stateName != InstanceStateName.Stopped ) {
```
batmat: I think it'd be great to introduce some small method to clarify to intent and make it easier to reason about/understand.
Here probably what we want in the end is like

```java
if ( isInstanceRunning( i ) {
... 
}
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -379,13 +379,16 @@ private int countCurrentEC2Slaves(SlaveTemplate template) throws AmazonClientExc
         Set<String> instanceIds = new HashSet<String>();
         String description = template != null ? template.description : null;
 
+        //FIXME convert to a filter query
```
batmat: :fireworks: :+1: these growing loops and ifs are becoming more and more scary :-).

thoulen: Yes ... we have to replace with a query on the AWS side, that will reduce the data transfer and the complexity of the if, I promise will be in the 1.43 :-)

PR #327, opened 2018-12-17
JENKINS-46294: Allow block reservations for spot instances

nullify005:
What it says on the tin, allows for the setting of a Spot Block when requesting a spot instance so that you get your instance for at least that duration.

https://issues.jenkins-ci.org/browse/JENKINS-46294


thoulen:
Can you resolve the conflicts ?

nullify005:
done, @thoulen

nullify005:
Something I'm not quite fond of now is the branching structure for spot instances introduced by the useBidPrice boolean.
It feels like there's code duplication here for not a lot of reason - however I'm not going to tackle that here

thoulen:
Thanks, for the work.
The spot instance support wasn't a first citizen in the plugin until now that start to be one of the most critical function. It needs a deep refactor, probably for the 1.50 I will adjust.

PR #328, opened 2018-12-19
Always use `getInstanceWithRetry` when using `CloudHelper`

facundovictor:
Whenever the EC2 cloud is accessed for getting an instance id using `CloudHelper`, it should retry
if the response is `InvalidInstanceID.NotFound`. Thus, it should just use `CloudHelper.getInstanceWithRetry`.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -332,7 +332,7 @@ public abstract class EC2AbstractSlave extends Slave {
      * @return instance in EC2.
      */
     public static Instance getInstance(String instanceId, EC2Cloud cloud) {
-        return CloudHelper.getInstance(instanceId, cloud);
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -348,9 +355,16 @@ void stop() {
             LOGGER.info("EC2 instance stop request sent for " + getInstanceId());
             toComputer().disconnect(null);
         } catch (AmazonClientException e) {
-            Instance i = CloudHelper.getInstance(getInstanceId(), getCloud());
-            LOGGER.log(Level.WARNING, "Failed to stop EC2 instance: " + getInstanceId() + " info: "
-                    + ((i != null) ? i : ""), e);
+            Instance i = null;
+            try {
```
thoulen: I thought that I replaced everywhere the call, but it seems is not.
I am not a fan of the try/Catch in a try/Catch, can you find a different solution ? (call the public getInstance)

PR #329, opened 2019-01-09
JENKINS-54329: allow user to choose slave connection strategy.

dgarzon:
This PR aims to fix the issues mentioned in: https://issues.jenkins-ci.org/browse/JENKINS-54329. Right now, the host used for connecting to the slaves is not really deterministic and restricts users as mentioned in the issue. I would like to get early feedback on this, but my plan is to also also remove the `associatePublicIp` field and use the `connectionStrategy` being used to better determine whether to associate the public IP or not. What do you guys think?

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -90,7 +90,6 @@ public abstract class EC2AbstractSlave extends Slave {
     public final String jvmopts; // e.g. -Xmx1g
     public final boolean stopOnTerminate;
     public final String idleTerminationMinutes;
-    public final boolean usePrivateDnsName;
     public final boolean useDedicatedTenancy;
     public boolean isConnected = false;
     public List<EC2Tag> tags;
@@ -121,8 +120,12 @@ public abstract class EC2AbstractSlave extends Slave {
     // Deprecated by the AMITypeData data structure
     @Deprecated
     protected transient int sshPort;
+
     @Deprecated
     public transient String rootCommandPrefix; // e.g. 'sudo'
+
+    @Deprecated
+    public transient boolean usePrivateDnsName;
 
     public transient String slaveCommandPrefix;
 
@@ -132,12 +135,7 @@ public abstract class EC2AbstractSlave extends Slave {
 
     public static final String TEST_ZONE = "testZone";
 
-    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
-            throws FormException, IOException {
-        this(name, instanceId, description, remoteFS, numExecutors, mode, labelString, launcher, retentionStrategy, initScript, tmpDir, nodeProperties, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, tags, cloudName, usePrivateDnsName, useDedicatedTenancy, launchTimeout, amiType, -1);
-    }
-
-    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, int maxTotalUses)
+    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
             throws FormException, IOException {
 
         super(name, remoteFS, launcher);
@@ -156,7 +154,7 @@ public abstract class EC2AbstractSlave extends Slave {
         this.stopOnTerminate = stopOnTerminate;
         this.idleTerminationMinutes = idleTerminationMinutes;
         this.tags = tags;
-        this.usePrivateDnsName = usePrivateDnsName;
+        this.usePrivateDnsName = connectionStrategy.equals(ConnectionStrategy.PRIVATE_DNS);
         this.useDedicatedTenancy = useDedicatedTenancy;
         this.cloudName = cloudName;
         this.launchTimeout = launchTimeout;
@@ -164,6 +162,12 @@ public abstract class EC2AbstractSlave extends Slave {
         this.maxTotalUses = maxTotalUses;
         readResolve();
         fetchLiveInstanceData(true);
+    }
+
+    @Deprecated
+    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+            throws FormException, IOException {
+        this(name, instanceId, description, remoteFS, numExecutors, mode, labelString, launcher, retentionStrategy, initScript, tmpDir, nodeProperties, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, tags, cloudName, useDedicatedTenancy, launchTimeout, amiType, ConnectionStrategy.backwardsCompatible(usePrivateDnsName, false, false), -1);
     }
 
     @Override
@@ -522,7 +526,6 @@ public abstract class EC2AbstractSlave extends Slave {
         createdTime = i.getLaunchTime().getTime();
         instanceType = i.getInstanceType();
 
-
         /*
          * Only fetch tags from live instance if tags are set. This check is required to mitigate a race condition
          * when fetchLiveInstanceData() is called before pushLiveInstancedata().
@@ -632,6 +635,7 @@ public abstract class EC2AbstractSlave extends Slave {
         return createdTime;
     }
 
+    @Deprecated
     public boolean getUsePrivateDnsName() {
         return usePrivateDnsName;
     }
--- a/src/main/java/hudson/plugins/ec2/EC2HostAddressProvider.java
+++ b/src/main/java/hudson/plugins/ec2/EC2HostAddressProvider.java
@@ -1,0 +1,56 @@
+package hudson.plugins.ec2;
+
+import com.amazonaws.services.ec2.model.Instance;
+import com.google.common.net.HostAndPort;
+import org.apache.commons.lang.StringUtils;
+
+import java.util.Optional;
+
+import static hudson.plugins.ec2.ConnectionStrategy.*;
+
+public class EC2HostAddressProvider {
+    public static String unix(Instance instance, ConnectionStrategy strategy) {
+        switch (strategy) {
+            case PUBLIC_DNS:
+                return filterNonEmpty(getPublicDnsName(instance)).orElse(getPublicIpAddress(instance));
+            case PUBLIC_IP:
+                return getPublicIpAddress(instance);
+            case PRIVATE_DNS:
+                return filterNonEmpty(getPrivateDnsName(instance)).orElse(getPrivateIpAddress(instance));
+            case PRIVATE_IP:
+                return getPrivateIpAddress(instance);
+            default:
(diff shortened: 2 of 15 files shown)


zimbatm:
Awesome work! Backward-compatible *and* includes tests.

Regarding `associatePublicIp` and `connectionStrategy`. They are related in the sense that a public IP `connectionStrategy` requires `associatePublicIp = true`. The reverse might not be true, the user might want to connect on the private IP but still allocate public IPs for users to SSH into the nodes.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -131,7 +130,7 @@
 
     public static final String TEST_ZONE = "testZone";
 
-    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2AbstractSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
```
jvz: Can you deprecate the old constructor and add a new constructor to maintain API compatibility? The old can delegate to this new one of course and drop the deprecated parameter.

src/main/java/hudson/plugins/ec2/EC2HostAddressProvider.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private static String getPrivateIpAddress(Instance instance) {
+        return instance.getPrivateIpAddress();
+    }
+
+    private static Optional<String> filterNonEmpty(String value) {
+        return Optional.ofNullable(value).filter(StringUtils::isNotEmpty);
+    }
+
+    public static class HostWithIpAddress {
```
jvz: You could potentially use `com.google.common.net.HostAndPort` here, or even `java.net.InetSocketAddress` if equality isn't too important (InetSocketAddress does some actual DNS lookups and such to compare different address/port combos).

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -33,20 +33,20 @@
...
+        this(description + " (" + instanceId + ")", instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, Collections.emptyList(), remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, false, launchTimeout, amiType);
     }
 
-    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
```
jvz: Same backward compatibility concern as in `EC2AbstractSlave`.

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -32,17 +32,17 @@
...
+        this(description + " (" + name + ")", spotInstanceRequestId, description, remoteFS, numExecutors, mode, initScript, tmpDir, labelString, Collections.emptyList(), remoteAdmin, jvmopts, idleTerminationMinutes, tags, cloudName, launchTimeout, amiType);
     }
 
     @DataBoundConstructor
-    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, int launchTimeout, AMITypeData amiType)
+    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType)
```
jvz: Same BC issues.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -161,10 +161,10 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
             InstanceType type, boolean ebsOptimized, String labelString, Node.Mode mode, String description, String initScript,
             String tmpDir, String userData, String numExecutors, String remoteAdmin, AMITypeData amiType, String jvmopts,
             boolean stopOnTerminate, String subnetId, List<EC2Tag> tags, String idleTerminationMinutes,
-            boolean usePrivateDnsName, String instanceCapStr, String iamInstanceProfile, boolean deleteRootOnTermination,
+            String instanceCapStr, String iamInstanceProfile, boolean deleteRootOnTermination,
             boolean useEphemeralDevices, boolean useDedicatedTenancy, String launchTimeoutStr, boolean associatePublicIp,
-            String customDeviceMapping, boolean connectBySSHProcess, boolean connectUsingPublicIp, boolean monitoring,
-            boolean t2Unlimited) {
+            String customDeviceMapping, boolean connectBySSHProcess, boolean monitoring,
```
jvz: And here, too. Maybe you can figure out how to support the old booleans by converting them into the enum as appropriate?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -383,10 +382,6 @@ public boolean getAssociatePublicIp() {
         return associatePublicIp;
     }
 
-    public boolean isConnectUsingPublicIp() {
```
jvz: You should be able to deprecate this and delegate to checking the enum here.

dgarzon:
Thanks for the review @jvz ! I will apply the requested changes tomorrow. Quick question: when you say deprecate, you mean to add the `@Deprecated` annotation to the `fields`, `methods` and `constructors` I removed / changed? Just making sure we are on the same page 👍

jvz:
> when you say deprecate, you mean to add the `@Deprecated` annotation to the `fields`, `methods` and `constructors` I removed / changed? Just making sure we are on the same page 👍

Yes, the annotation. You can also add a javadoc comment for it to describe the deprecation if it seems appropriate.

dgarzon:
@jvz pushed with your suggested changes. I also improved the algorithm for backward compatibility in `ConnectionStrategy.backwardsCompatible` to make better use of the three booleans. Please let me know what you think!

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -33,20 +33,26 @@
...
-        this(description + " (" + instanceId + ")", instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, Collections.emptyList(), remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, usePrivateDnsName, useDedicatedTenancy, launchTimeout, amiType);
+        this(description + " (" + instanceId + ")", instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, Collections.emptyList(), remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, useDedicatedTenancy, launchTimeout, amiType);
     }
 
     @DataBoundConstructor
-    public EC2OndemandSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
```
jvz: Where did you move this constructor to?

dgarzon: It is right above the new one, and is annotated with `@Deprecated`. The git-diff is confused because of the annotation I believe!

jvz:
Mostly looks good to merge. Just need to make sure anything that was already `public` before is still there now. For the public fields you removed, you should keep them there and mark the `@Deprecated`. I'm not sure if it's possible to translated the old booleans into their equivalent new options; if it is, that would be the best way to ensure backward compatibility.

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
jvz: Ah, ok.

dgarzon:
Got it @jvz! I was able to achieve full backwards compatibility by leaving the old stuff in place and just use the `ConnectionStrategy.backwardsCompatible` to determine the correct strategy. Some tests are now failing, but will keep debugging!

thoulen:
DAniel, can you rebase ?  I'd like to merge it for the 1.43

dgarzon:
@thoulen On it. Give me an hour or so!

dgarzon:
@thoulen I rebased, but cannot compile / run tests locally. There seem to be an issue with the current `pom.xml`:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (display-info) on project ec2: Execution display-info of goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce failed.: NullPoin
terException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (display-info) on project ec2: Execution display-info of goal org.apache.maven.plugins:maven-en
forcer-plugin:3.0.0-M1:enforce failed.

Caused by: java.lang.NullPointerException
    at org.apache.maven.plugins.enforcer.EnforceBytecodeVersion.isBadArtifact (EnforceBytecodeVersion.java:301)
    at org.apache.maven.plugins.enforcer.EnforceBytecodeVersion.checkDependencies (EnforceBytecodeVersion.java:210)
    at org.apache.maven.plugins.enforcer.EnforceBytecodeVersion.handleArtifacts (EnforceBytecodeVersion.java:136)
    at org.apache.maven.plugins.enforcer.AbstractResolveDependencies.execute (AbstractResolveDependencies.java:77)
    at org.apache.maven.plugins.enforcer.EnforceMojo.execute (EnforceMojo.java:202)
```

src/main/java/hudson/plugins/ec2/EC2HostAddressProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.amazonaws.services.ec2.model.Instance;
+import com.google.common.net.HostAndPort;
+import org.apache.commons.lang.StringUtils;
+
+import java.util.Optional;
+
+import static hudson.plugins.ec2.ConnectionStrategy.*;
+
+public class EC2HostAddressProvider {
```
jvz: You can make this class `final` with a `private` constructor since it's all static methods. Unless you intend to make this class overridable in general?

casz:
You constantly break the data binding in JCasC by having your databound constructors require so many parameters with no null point checking.
https://github.com/jenkinsci/configuration-as-code-plugin/issues/885

casz:
We should switch the databound constructors to only mandatory parameters and the rest of the parameters should be databound setters.

jvz:
@casz PRs welcome to help improve the backward compatibility situation going on here.

casz:
:sweat: I know, I sadly have plenty of work on my hands already

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -32,23 +32,24 @@
...
             throws FormException, IOException {
-        this(description + " (" + name + ")", spotInstanceRequestId, description, remoteFS, numExecutors, mode, initScript, tmpDir, labelString, Collections.emptyList(), remoteAdmin, jvmopts, idleTerminationMinutes, tags, cloudName, usePrivateDnsName, launchTimeout, amiType, maxTotalUses);
+        this(name, spotInstanceRequestId, description, remoteFS, numExecutors, mode, initScript, tmpDir, labelString, remoteAdmin, jvmopts, idleTerminationMinutes, tags, cloudName, false, launchTimeout, amiType);
     }
 
-    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, int launchTimeout, AMITypeData amiType)
```
varyvol: Isn't this constructor missing?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -158,15 +158,21 @@
     @Deprecated
     public transient String slaveCommandSuffix;
 
+    @Deprecated
+    public transient boolean usePrivateDnsName;
```
varyvol: May I ask why this was declared transient?

casz: Because it was deprecated?

varyvol: See https://github.com/jenkinsci/ec2-plugin/pull/400

PR #273, opened 2018-04-06
Add support for Spot Instances without bidding

dblanchette:
At re:Invent 2017, AWS announced it would support starting EC2 spot instances without setting a bid price. The bid price is automatically set at the current On-Demand price. You still pay the current market price, which is most often less than the On-Demand price.

Supporting this in the plugin decreases the amount of configuration needed by the user. There is only a checkbox to check.

2018-04-06_09h50_07

PR #331, opened 2019-01-29
Allow launching spot slaves without a bid price

julienduchesne:
Redo of https://github.com/jenkinsci/ec2-plugin/pull/273

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -536,7 +536,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
     /**
      * Provisions an On-demand EC2 slave by launching a new instance or starting a previously-stopped instance.
      */
-    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions, Boolean spotWithoutBidPrice)
+    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions, boolean spotWithoutBidPrice)
             throws IOException {
         AmazonEC2 ec2 = getParent().connect();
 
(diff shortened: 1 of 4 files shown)


thoulen:
Can you add a unit test ?

julienduchesne:
Of course

julienduchesne:
Done. Added a SlaveTemplate test

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -528,7 +528,16 @@ private void logProvisionInfo(String message) {
...
-    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions) throws AmazonClientException, IOException {
+    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions)
+            throws IOException {
+        return provisionOndemand(number, provisionOptions, false);
+    }
+
+    /**
+     * Provisions an On-demand EC2 slave by launching a new instance or starting a previously-stopped instance.
+     */
+    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions, Boolean spotWithoutBidPrice)
```
```suggestion
    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions, boolean spotWithoutBidPrice)
```

Otherwise you'll have a potential NPE on line 549.

src/main/java/hudson/plugins/ec2/SpotConfiguration.java
```diff
@@ -3,29 +3,29 @@
 import org.kohsuke.stapler.DataBoundConstructor;
 
 public final class SpotConfiguration {
+    public final Boolean useBidPrice;
```
jvz: Can you make this a `boolean` instead? If `boolean` doesn't work properly with `@DataBoundConstructor`, you can still use `Boolean` in the constructor args and unbox it there (convert `null` to whatever the default setting should be).

src/main/java/hudson/plugins/ec2/SpotConfiguration.java
```diff
@@ -34,11 +34,11 @@ public static String normalizeBid(String bid) {
 
             /* The specified bid price cannot be less than 0.001 */
             if (spotPrice < 0.001) {
-                return null;
+                return "";
```
jvz: This doesn't match the javadoc anymore. Which part is outdated?

julienduchesne:
Thanks for the comments @jvz. Fixed the issues

PR #332, opened 2019-01-30
Spot instances: Fallback to on-demand when there is no capacity

julienduchesne:
Follow up to #331. When there is no spot capacity (either for spot instances with a bid price or without one), fallback to on-demand instances

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -537,7 +537,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
     /**
      * Provisions an On-demand EC2 slave by launching a new instance or starting a previously-stopped instance.
      */
-    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions, Boolean spotWithoutBidPrice, Boolean fallbackSpotToOndemand)
+    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions, boolean spotWithoutBidPrice, boolean fallbackSpotToOndemand)
             throws IOException {
         AmazonEC2 ec2 = getParent().connect();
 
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -528,7 +529,16 @@ private void logProvisionInfo(String message) {
...
-    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions) throws AmazonClientException, IOException {
+    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions)
+            throws IOException {
+        return provisionOndemand(number, provisionOptions, false, false);
+    }
+
+    /**
+     * Provisions an On-demand EC2 slave by launching a new instance or starting a previously-stopped instance.
+     */
+    private List<EC2AbstractSlave> provisionOndemand(int number, EnumSet<ProvisionOptions> provisionOptions, Boolean spotWithoutBidPrice, Boolean fallbackSpotToOndemand)
```
jvz: Same boolean boxing concerns as in #331.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -929,6 +961,25 @@ private void setupCustomDeviceMapping(List<BlockDeviceMapping> deviceMappings) {
                 }
                 String slaveName = spotInstReq.getSpotInstanceRequestId();
 
+                if (spotConfig.fallbackToOndemand) {
+                    for (int i = 0; i < 2 && spotInstReq.getStatus().getCode().equals("pending-evaluation"); i++) {
```
jvz: I guess it's not exactly required, but jenkins-core somewhat recently added a `Retrier<T>` class which might be useful here. Might be worth looking at? I'm mostly concerned about `Thread.sleep` calls for testing purposes.

julienduchesne: hmm, the Retrier is very nice. However, it requires us to bump the jenkins version to at least 2.152. And the Retrier is marked `NoExternalUse` which crashes the build (unless there's a way around that).

jvz: Oh right. NoExternalUse is for avoiding creating new public APIs that can never change again. Ah well.

PR #336, opened 2019-04-01
Use secretTextarea for private keys

jvz:
This provides a more secure UI component for entering multiline secret
values.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,11 @@ THE SOFTWARE.
             <version>4.5.5-3.0</version>
         </dependency>
         <dependency>
+            <groupId>io.jenkins.temp.jelly</groupId>
+            <artifactId>multiline-secrets-ui</artifactId>
+            <version>1.0</version>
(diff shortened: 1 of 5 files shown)


stefanverhoeff:
Thanks! I've noticed before that the secret was visible with the "global config readonly" view, which is supposed to be a safe view of the config. This should fix that.

jvz:
> 🐝
> 
> Perhaps seems interesting to create a ticket?

We're pretty disorganized in this plugin about that. PRs or JIRA issues both work for the changelog.

pom.xml
```diff
@@ -149,6 +149,11 @@ THE SOFTWARE.
             <artifactId>apache-httpcomponents-client-4-api</artifactId>
             <version>4.5.5-3.0</version>
         </dependency>
+        <dependency>
+            <groupId>io.jenkins.temp.jelly</groupId>
```
jglick: TODO minimum baseline to remove

PR #341, opened 2019-04-26
Add plugin description

res0nance:
Add a plugin description which will show up in the plugin manager


res0nance:
Channel closed exception needs to be retriggered

PR #343, opened 2019-05-04
[JENKINS-25106] Support minimumNoInstances

erikhakansson:
PR not finished yet. Just publishing to allow commenting on it.


erikhakansson:
I would like some tips on how to properly test this. Any way to mock calls against ec2 api?

erikhakansson:
Finished the PR now. Should work fine!

erikhakansson:
@thoulen sorry to ping you, I just saw you had some recent commits so I guess you're one of the maintainers.

Do you think you could have a look at this PR?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -751,7 +793,11 @@ private static AmazonWebServicesCredentials getCredentials(@Nullable String cred
 
     private AmazonEC2 reconnectToEc2() throws IOException {
         synchronized(this) {
-            connection = connect(createCredentialsProvider(), getEc2EndpointUrl());
+            if (AmazonEC2Cloud.isTestMode()) {
```
snallami: I am not maintainer or owner of this plugin. Just happen to see this and adding my review comments.

I am not a big fan of mixing mock code in original code. Some suggestions here https://8thlight.com/blog/dariusz-pasciak/2014/12/19/dont-mix-test-code-with-production-code.html

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -784,12 +830,22 @@ public AmazonEC2 connect() throws AmazonClientException {
      * @return {@link AmazonEC2} client
      */
     public static AmazonEC2 connect(AWSCredentialsProvider credentialsProvider, URL endpoint) {
+        if (AmazonEC2Cloud.isTestMode()) {
```
snallami: Same comment like above abt mixing test code and production code

src/main/java/hudson/plugins/ec2/EC2Cloud.java
erikhakansson: Well, no, generally you don't want that, but as you can see in the code this methodology is already used, and frankly I'm not sure Jenkins supports injecting and/or replacing (test) classes in a running Jenkins instance.

That blog post deals mostly with unit testing.

jvz: We should clean up this test mode stuff, but I'd be happy seeing it done in a separate PR since it's both pervasive already in this plugin, and fixing it here would make the diff gigantic.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -784,12 +830,22 @@ public AmazonEC2 connect() throws AmazonClientException {
...
+            throw new IllegalStateException("This method should not be used in test mode");
+        }
         awsCredentialsProvider = credentialsProvider;
         AmazonEC2 client = new AmazonEC2Client(credentialsProvider, createClientConfiguration(endpoint.getHost()));
         client.setEndpoint(endpoint.toString());
         return client;
     }
 
+    public synchronized static AmazonEC2 mockConnect(AWSCredentialsProvider credentialsProvider, URL endpoint, EC2PrivateKey privateKey) {
```
jvz: 1. Test code should be either package-private or annotated with `@Restricted(NoExternalUse.class)`.
2. This does not seem to be a good way to introduce test code here anyways, though restricting its visibility will make it simpler to refactor.

src/main/java/hudson/plugins/ec2/util/MockAmazonEC2.java
```diff
@@ -0,0 +1,196 @@
...
+import hudson.plugins.ec2.AmazonEC2Cloud;
+import hudson.plugins.ec2.EC2PrivateKey;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class MockAmazonEC2 extends AmazonEC2Client {
```
jvz: Ok this definitely doesn't belong in the main source. To mock out a connection, you can use Mockito or PowerMockito to replace the connection instances being created at test time without adding test code at run time.

erikhakansson: @jvz well that's the issue really. How would you mock out a class that's loaded by an explicit `new` inside of a class loaded by the Jenkins plugin system in a running Jenkins instance?

Since the test is about testing that the functionality is properly triggered by Jenkins functionality I need to test it within a running Jenkins instance.

I *think* Jenkins instances have access to test classes when running with the JunitRunner, so I could move `MockAmazonEC2` to the `test` package, but then there'd be no compile-time knowledge of the class when choosing which implementation to load (i.e. real or test) so I'd have to use reflection which in my opinion is even dirtier.

So I'm really unsure how to test it in any other way.

erikhakansson:
@jvz maybe that refactoring you mentioned is actually necessary before this is really testable.

Perhaps some stuff can be made to use DI instead. As I understand it, Jenkins uses Guice. I've never used Guice specifically, but using any DI it should be possible to have a named injection, and perhaps that name can be read from a JVM parameter (or a default). Thus, when running tests we could just always pass a certain parameter that would ensure test (mock) classes would be loaded instead of real implementations. That way, test classes can remain firmly as test code and stay out of main source.

Thoughts on that?

jvz:
Any class that's annotated with `@Extension` is already wired into Guice, so you can `@Inject` there at least.

src/main/java/hudson/plugins/ec2/util/MockAmazonEC2.java
jvz: You might be able to use PowerMockito's static method mocking feature. Otherwise, the client factory should be refactored to be testable like you said.

erikhakansson:
@jvz 

I found this: https://stackoverflow.com/questions/7651980/google-guice-and-varying-injections-at-runtime (second answer)

Using that logic it might be possible to have a custom provider do something like:

```
@Override 
protected void configure() {
  String implementationClass = System.getProperty("my.implementation");
  if (implementationClass == null) {
    implementationClass = DEFAULT_IMPLEMENTATION_CLASS;
   }
   bind(MyInterface.class).to(Class.forName(implementationClass));
  }
```

Obviously with proper catches for ClassNotFound and such, but anyway, that way you can easily load a different implementation if running in test mode.

erikhakansson:
@jvz please see #379

erikhakansson:
Merged in changes from #379, refactored based on those changes, and squashed commits.

erikhakansson:
I.e, this depends on #379 being approved and merged.

naxhh:
Hi #379 has been merged! I was looking into implementing this feature so this is a +1 from us.
If there's something we can do to drive this forward..

erikhakansson:
@jvz Do you think you can have a look at this again now, if you have time? :)

@naxhh If you would try building and running it locally I think it would be great. If the maintainers get reports of successful real-world usage I'm sure they will feel safer accepting the PR. Also, real-world is usually better at exposing bugs than unit tests :)

src/main/java/hudson/plugins/ec2/util/EC2SlaveFactory.java
```diff
@@ -0,0 +1,32 @@
...
+            if (instance != null) {
+                throw new IllegalStateException("Multiple implementations of " + EC2SlaveFactory.class.getName()
+                        + " found. If overriding, please consider using ExtensionFilter");
+            }
+            instance = implementation;
+        }
+        return instance;
+    }
+
+    EC2OndemandSlave createOnDemandSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, String labelString, Node.Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
```
jvz: Would it make more sense to use some sort of config class object here? Having more than a couple method parameters in Java gets really hard to deal with.

src/test/java/hudson/plugins/ec2/EC2PrivateKeyTest.java
```diff
@@ -41,30 +42,8 @@
...
-                + "gYEAxAZSgXtfyCkJJSWJeQ44ra9/emBykuJzA4da21jOnm+qiA5n7kWWJVC5KgB/3RC8t1dKd81U\n"
-                + "DArRidvgaV5+PSlF+S541NxlriPgRfCFDbt4AkOpapHrczy2/jYfMU7Qyo616VKTZD3huU+JTK1I\n"
-                + "SEw24BaQH/LQY1pmcdns/QECgYAqkLcR6gukUryMIkCEvtycWQ493VzexWQfZBTEpXLfwciGHnxw\n"
-                + "b2dHx6vJpkclKEsacYNwZM/qv/54HMiaYry3fsOa0uCvco7+2kowDju3r3TRuWQxyLNxJd/2fCo8\n"
-                + "0cZ3kbJzHluG2igswL+F3zC1sFoCFtJLflnQJl+VO5HFKwKBgD5F/+paTIYc622xiDeCYsYfqnpq\n"
-                + "MpZJorRzNPGTxmv7Kg94kFh7h7zuccUcNn15iUpNRTwLUZpKArYcuU1bhnveBD4l+84XBii6mFjz\n"
-                + "9ontJin0nlHPk+AOmV8xt3yYD+wPAJy5MjUco7tS4Ix6bmvxcpZi2ZcHT1GwkiIzgKWE\n"
-                + "-----END RSA PRIVATE KEY-----");
+    private static EC2PrivateKey getPrivateKey() {
+        return new EC2PrivateKey(TestConstants.PRIVATE_KEY);
```
jvz: Not really in scope here, but this key could be replaced with a `KeyPairGenerator` instance that creates a new key for each test run.

jvz:
Looks like this is coming along nicely. @fcojfernandez can you or a teammate review?

src/main/java/hudson/plugins/ec2/util/EC2SlaveFactory.java
erikhakansson: Yeah I suppose. I just went with the lowest effort change to just retain the current parameters.

But how about a config object with a builder pattern that throws an exception for missing parameters?

src/test/java/hudson/plugins/ec2/EC2PrivateKeyTest.java
erikhakansson: Ah, I was not aware of the existence of that. :) 

Want me to implement it in this PR despite being out of scope?

jvz: That's up to you. I've updated a test [in jsch-plugin](https://github.com/jenkinsci/jsch-plugin/blob/master/src/test/java/org/jenkinsci/plugins/jsch/JSchSSHPublicKeyAuthenticatorTest.java) a while ago to do this. There are similar tests in ssh-credentials-plugin that hard code the keys, too. This only came to my attention before due to some sort of static analysis complaining about embedding private keys in the source code.

src/main/java/hudson/plugins/ec2/util/EC2SlaveFactory.java
jvz: That sounds like a great way to approach it!

jvz:
By the way, DevOps World/Jenkins World is going on this week, so review might be delayed.

src/main/java/hudson/plugins/ec2/util/EC2SlaveFactory.java
erikhakansson: I pushed a change here.

I decided against throwing Exceptions since I don't really know if some of those parameters should in fact be nullable.

src/test/java/hudson/plugins/ec2/EC2PrivateKeyTest.java
erikhakansson: I've managed to solve this locally for most tests.

However, the tests in EC2PrivateKeyTest specifically are a bit cumbersome because they rely on comparison against a hard-coded key fingerprint.

I could of course calculate the fingerprint of the generated key, but that generation is basically what the test already tests. So I'd more or less be replicating the production code in the test, and then testing whether my test replication functions the same as the production code.

I could of course revert the `EC2PrivateKeyTest.java` and retain the hard-coded key there, and just use the generated key for the rest of the tests, but then we'll still have the hard-coded key.

Suggestion?

jvz: Sounds like you'd be better off using the hard-coded key in that specific test at least.

erikhakansson: I pushed a change with that implemented now!

I also squashed the commits.

erikhakansson:
Is it possible to force retrigger the CI build? 

It failed for some network reason before it could even check out.

jvz:
Yeah, just add another empty commit or change the commit somehow. Or you could close this, wait a minute, then open it again which should also trigger CI again.

erikhakansson:
@fcojfernandez sorry I haven't replied to you. I haven't been able to look at your comments yet. I hope to get some time during the week!

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -148,7 +157,7 @@ private long internalCheck(EC2Computer computer) {
             // * Already Terminated
             // * We use stop-on-terminate and the instance is currently stopped or stopping
             if (InstanceState.TERMINATED.equals(state)
-                  || computer.getSlaveTemplate().stopOnTerminate && (InstanceState.STOPPED.equals(state) || InstanceState.STOPPING.equals(state))) {
+                  || slaveTemplate.stopOnTerminate && (InstanceState.STOPPED.equals(state) || InstanceState.STOPPING.equals(state))) {
```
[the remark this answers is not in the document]
erikhakansson: You're absolutely right it should be there too. It wasn't initially and I just changed to re-use the variable from above and didn't think about null-checks at all. Functionally it should make no difference from today since `computer.getSlaveTemplate()` is what's used today as well. 

Nevertheless I'll add a null-check.

erikhakansson:
@fcojfernandez null check implemented and pushed :)

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -170,7 +174,7 @@
     public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, String securityGroups, String remoteFS,
             InstanceType type, boolean ebsOptimized, String labelString, Node.Mode mode, String description, String initScript,
             String tmpDir, String userData, String numExecutors, String remoteAdmin, AMITypeData amiType, String jvmopts,
-            boolean stopOnTerminate, String subnetId, List<EC2Tag> tags, String idleTerminationMinutes,
+            boolean stopOnTerminate, String subnetId, List<EC2Tag> tags, String idleTerminationMinutes, String minimumNumberOfInstancesStr,
```
jvz: Are you able to use `int` or `Integer` here?

erikhakansson: It should be possible. I just followed the pattern of existing "int-type" parameters like `numExecutors` and `idleTerminationMinutes` where a String is parsed instead. 

Not sure what the reasoning behind those are since I'm not too familiar with the Jenkins gui code. And at any rate, form validation occurs before posting.

erikhakansson: Should I change to `int`?

jvz: Yeah, just stick to int where you can here as long as it works. The other variables are probably legacy.

res0nance: Could you fix this bit @erikhakansson

src/main/java/hudson/plugins/ec2/util/MinimumInstanceChecker.java
```diff
@@ -0,0 +1,42 @@
...
+import jenkins.model.Jenkins;
+
+import javax.annotation.Nonnull;
+import java.util.Arrays;
+import java.util.Objects;
+
+public class MinimumInstanceChecker {
+
+  public static int countCurrentNumberOfSlaves(@Nonnull SlaveTemplate slaveTemplate) {
+    return (int) Arrays.stream(Jenkins.get().getComputers()).filter(computer -> {
```
res0nance: Nit: I'd prefer if it was 4 spaces like the rest of the project

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
erikhakansson: Sorry, I haven't had time to sit with this for a while but I'm looking into it now!

erikhakansson: Done :)

src/main/java/hudson/plugins/ec2/util/MinimumInstanceChecker.java
erikhakansson: I agree. Just the default settings from my IDE that came through. Fixed!

fcojfernandez:
Relaunching the CI

erikhakansson:
Seems like unrelated errors on CI side. But for some reason it doesn't rebuild when I close and open

erikhakansson:
I can't figure this out.
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fec2-plugin/detail/PR-343/35/pipeline/11

The build obviously works, but on Windows the CI fails to clean up the workspace after.

erikhakansson:
Ok I'm even more confused now. I rebased against master and added a commit disabling my new tests, just to see what might cause the workdir to be undeletable. Now CI says 

```
[2019-09-20T16:01:01.848Z] Merging remotes/origin/master commit bc2c4f09236e40a64ac66ce86f4fabdda8387b76 into PR head commit e5383223c2d5f18ce03ae2cf99b9882c6f2f8856

Cannot checkout from unborn branch
```

erikhakansson:
I created a new PR (#401) with the exact same changes and it builds (although it's currently stalled at a later step, waiting for a new slave).

Should I just close this and continue on the other one?

jvz:
Yeah, let's continue on the new PR. It seems as though this branch's git history might have caused an issue in CI.

erikhakansson:
Yeah seems that way. I'm closing this.

erikhakansson:
I had a spelling mistake in the branch name anyway :)

PR #344, opened 2019-05-06
[For Discussion] SMB2 Support

res0nance:
This PR implements SMB2/3 support by leveraging SMBj. Unfortunately SMBj doesn't support SMB1 therefore this removes support for SMB1 also. Maybe it should be made backwards compatible or we should move off the old protocol.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
@@ -52,6 +52,7 @@ public class EC2WindowsLauncher extends EC2ComputerLauncher {
                 logger.println("Executing init script");
                 OutputStream init = connection.putFile(tmpDir + "init.bat");
                 init.write(initScript.getBytes("utf-8"));
+                init.close();
 
                 WindowsProcess initProcess = connection.execute("cmd /c " + tmpDir + "init.bat");
                 IOUtils.copy(initProcess.getStdout(), logger);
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
```diff
@@ -64,11 +65,13 @@ protected void launchScript(EC2Computer computer, TaskListener listener) throws
 
                 OutputStream initGuard = connection.putFile(tmpDir + ".jenkins-init");
                 initGuard.write("init ran".getBytes(StandardCharsets.UTF_8));
+                initGuard.close();
```
jvz: It would make more sense to change these code snippets to use try-with-resources like:

```
try (OutputStream o = ...) {
  ...
}
```

src/main/java/hudson/plugins/ec2/win/WinConnection.java
```diff
@@ -40,10 +47,8 @@ public WinConnection(String host, String username, String password) {
         this.host = host;
         this.username = username;
         this.password = password;
-        this.authentication = new NtlmPasswordAuthentication(null, username, password);
-        if(log.isLoggable(Level.FINE)){
-            setLogStreamLevel(6);
-        }
+        this.smbclient = new SMBClient();
+        this.authentication = new AuthenticationContext(username, password.toCharArray(), null);;
```
jvz:
```suggestion
        this.authentication = new AuthenticationContext(username, password.toCharArray(), null);
```

sean-zou:
+1

thoulen:
> This PR implements SMB2/3 support by leveraging SMBj. Unfortunately SMBj doesn't support SMB1 therefore this removes support for SMB1 also. Maybe it should be made backwards compatible or we should move off the old protocol.

The SMB1 cannot be supported it is insecure, also Microsoft mark as deprecated and not supported in many cases

res0nance:
I just tested this on Win Server 2016 and 2019 and seems to work well

res0nance:
Added a spotbugs filter for JDK11 tests regarding bug in spotbugs/spotbugs#756

PR #345, opened 2019-05-06
[JENKINS-52208] Fix findbugs in repository

res0nance:
Fixing the findbugs in the repository
Some changes would likely need exceptions thrown as I'm not sure if the function should ever be in that state.


thoulen:
Please can you rebase ?

PR #347, opened 2019-05-08
Add default to the new fields

thoulen:
Add default for maxTotalUses and for the connectionStrategy


thoulen:
> 🍜 assuming it's impractical to add tests

Good point, we are thinking to use a complete different structure to store the config, that should also give us a simple way to add tests

PR #348, opened 2019-05-09
[JENKINS-56907] Query AMI for platform before checking spot price

res0nance:


res0nance:
I'm not sure how to write a good test for this since it requires an aws call

PR #349, opened 2019-05-09
[JENKINS-53858] fix implementation of double-check-locking

ndeloof:


thoulen:
> _No description provided._

Adrress Bug : JENKINS-53858

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -759,20 +759,22 @@ private static AmazonWebServicesCredentials getCredentials(@Nullable String cred
     public AmazonEC2 connect() throws AmazonClientException {
         try {
             if (connection == null) {
-                connection = connect(createCredentialsProvider(), getEc2EndpointUrl());
+                synchronized (this) {
```
```suggestion
                synchronized (this) {
                    if (connection == null)
```

Otherwise, you still end up with multiple threads calling the method.

PR #350, opened 2019-05-09
Set timeout

sean-zou:


thoulen:
Can you explain the use case ? and open a ticket for that.
If we overwrite a system default , it is better to have in the configuration

res0nance:
Closing this due to no replies

PR #352, opened 2019-05-15
Add remoting workdir

res0nance:
This PR will set by default workdir to be the `remoteFS`

### DIFF
--- a/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
@@ -73,8 +73,9 @@ public class EC2WindowsLauncher extends EC2ComputerLauncher {
             logger.println("remoting.jar sent remotely. Bootstrapping it");
 
             final String jvmopts = node.jvmopts;
+            final String remoteFS = node.getRemoteFS();
             final WindowsProcess process = connection.execute("java " + (jvmopts != null ? jvmopts : "") + " -jar "
-                    + tmpDir + AGENT_JAR, 86400);
+                    + tmpDir + AGENT_JAR + " -workDir " + remoteFS, 86400);
             computer.setChannel(process.getStdout(), process.getStdin(), logger, new Listener() {
                 @Override
                 public void onClosed(Channel channel, IOException cause) {
--- a/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
@@ -74,8 +74,9 @@ public class EC2WindowsLauncher extends EC2ComputerLauncher {
 
             final String jvmopts = node.jvmopts;
             final String remoteFS = node.getRemoteFS();
-            final WindowsProcess process = connection.execute("java " + (jvmopts != null ? jvmopts : "") + " -jar "
-                    + tmpDir + AGENT_JAR + " -workDir " + remoteFS, 86400);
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
```diff
@@ -73,8 +73,9 @@ protected void launchScript(EC2Computer computer, TaskListener listener) throws
             logger.println("remoting.jar sent remotely. Bootstrapping it");
 
             final String jvmopts = node.jvmopts;
+            final String remoteFS = node.getRemoteFS();
             final WindowsProcess process = connection.execute("java " + (jvmopts != null ? jvmopts : "") + " -jar "
-                    + tmpDir + AGENT_JAR, 86400);
+                    + tmpDir + AGENT_JAR + " -workDir " + remoteFS, 86400);
```
jvz: Doesn't this parameter need to be escaped/quoted?

res0nance: Doesn't seem to need to be escaped in my testing it simply accepted it. Although it should require escaping in the case of paths with spaces

res0nance:
@jvz i added quotes around remoteFS it works except you have to avoid the trailing `\` or the jar will reject the workDir

src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
```diff
@@ -73,8 +73,10 @@ protected void launchScript(EC2Computer computer, TaskListener listener) throws
             logger.println("remoting.jar sent remotely. Bootstrapping it");
 
             final String jvmopts = node.jvmopts;
-            final WindowsProcess process = connection.execute("java " + (jvmopts != null ? jvmopts : "") + " -jar "
-                    + tmpDir + AGENT_JAR, 86400);
+            final String remoteFS = node.getRemoteFS();
+            final String launchString = "java " + (jvmopts != null ? jvmopts : "") + " -jar " + tmpDir + AGENT_JAR + " -workDir \"" + remoteFS + "\"";
```
jvz: You can use [WindowsUtil](https://github.com/jenkinsci/jenkins/blob/master/core/src/test/java/hudson/os/WindowsUtil.java) to properly quote arguments for both Windows and `cmd.exe` arguments (i.e., for use with `cmd.exe /c COMMAND` like with how you execute a builtin like `mklink`).

res0nance: I'd love to use it but unfortunately because its in the `/test` path of core its unusable here. Unless i fork that bit or have it moved into core. But even then core needs to be bumped again to use it.

jvz: Oh crap, I forgot it was only a test class. Feel free to copy/paste the class with a TODO comment of moving it to jenkins-core at some point.

I wrote the initial utility class while rewriting some flaky IO-related tests as well as one regarding junction points in Windows (fairly similar to symbolic links but with Windows-specific filesystem features), but I bet we can find more improper quoting of Windows command line arguments throughout Jenkins that could benefit from it.

jvz:
Got a merge conflict now as well.

res0nance:
I resolved the conflict and added a fix for the case where the tmp directory does not have a trailing slash.
I also moved `WindowsUtil` in and remove the assert and replaced it with an exception

zbynek:
Seems to trigger https://issues.jenkins-ci.org/browse/JENKINS-58163

PR #355, opened 2019-05-27
fix for java.lang.UnsupportedOperationException

tommywo:
related to https://jenkins.io/redirect/class-filter/


src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -107,6 +107,9 @@ public long check(EC2Computer c) {
             return 1;
         } else {
             try {
+                if (this.clock == null){
+                    this.clock = Clock.systemUTC();
```
res0nance: This bit should be in `readResolve()` as the field is `transient` now

res0nance:
I see in this class that there is another constructor and I think it is only used in tests. That lies outside the scope of this PR but is there a way to mark it only for tests @jvz because if the passed in `Clock` is null it will cause NPEs

PR #356, opened 2019-05-28
Use filter query when describing instances

res0nance:
- Get the API to do the filtering instead of doing it ourselves

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -386,7 +386,7 @@ public abstract class EC2Cloud extends Cloud {
         Set<String> instanceIds = new HashSet<String>();
         String description = template != null ? template.description : null;
 
-        List<Filter> filters = getGenericFilters(jenkinsServerUrl);
+        List<Filter> filters = getGenericFilters(jenkinsServerUrl, template);
         List<String> values = new ArrayList<>();
         values.add("pending");
         values.add("running");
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -386,42 +386,29 @@ private int countCurrentEC2Slaves(SlaveTemplate template) throws AmazonClientExc
         Set<String> instanceIds = new HashSet<String>();
         String description = template != null ? template.description : null;
 
-        //FIXME convert to a filter query
-        for (Reservation r : connect().describeInstances().getReservations()) {
+        List<Filter> filters = getGenericFilters(jenkinsServerUrl, template);
+        filters.add(new Filter("instance-state-name").withValues("running", "pending", "stopping"));
+        if (template != null) {
+            filters.add(new Filter("image-id").withValues(template.getAmi()));
```
res0nance: Slightly worried that this may cause issues if someone changes the `ami_id`

res0nance:
@jvz I think I'm done with this just turning it into the equivalent call letting the aws api deal with as much of the filtering as possible

jvz:
Can you either add a test or describe how to manually test this?

res0nance:
Behaviour doesn't change it has the exact behaviour as before as for automated testing we need a way to mock the aws api for that

res0nance:
@thoulen Can you take a look at this? This should help reduce the amount of data we pull from aws api and offload some of the filtering logic back to aws.

thoulen:
I am testing the patch

res0nance:
Added a fix to take care of a possible miscount when nexttoken is available

res0nance:
This PR also includes a bugfix where instance counting will be wrong if there are more than 1000 instances because we ignore the next token

res0nance:
Tested this and my instance caps are still properly respected so I'm merging it.

PR #296, opened 2018-07-27
Discard stale connections

JamesXNelson:
When a new instance is not created regularly, it can start to fail regularly:
```
    com.amazonaws.services.ec2.model.AmazonEC2Exception: Request has expired. (Service: AmazonEC2; Status Code: 400; Error Code: RequestExpired; Request ID: d6e3016d-bf8c-4374-8e3e-fe9b969c1734)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleErrorResponse(AmazonHttpClient.java:1632)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeOneRequest(AmazonHttpClient.java:1304)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeHelper(AmazonHttpClient.java:1058)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.doExecute(AmazonHttpClient.java:743)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeWithTimer(AmazonHttpClient.java:717)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.execute(AmazonHttpClient.java:699)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.access$500(AmazonHttpClient.java:667)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutionBuilderImpl.execute(AmazonHttpClient.java:649)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:513)
	at com.amazonaws.services.ec2.AmazonEC2Client.doInvoke(AmazonEC2Client.java:16445)
	at com.amazonaws.services.ec2.AmazonEC2Client.invoke(AmazonEC2Client.java:16421)
	at com.amazonaws.services.ec2.AmazonEC2Client.executeDescribeInstances(AmazonEC2Client.java:8104)
	at com.amazonaws.services.ec2.AmazonEC2Client.describeInstances(AmazonEC2Client.java:8079)
	at com.amazonaws.services.ec2.AmazonEC2Client.describeInstances(AmazonEC2Client.java:8116)
	at hudson.plugins.ec2.EC2Cloud.countCurrentEC2Slaves(EC2Cloud.java:363)

```

This change simply tests that a connection actually works, or else tries to create a new one.

PR #358, opened 2019-05-30
Fix JCasC backwards compatibility

res0nance:
- Fix backwards compatibility with JCasC configurations broken by #329 
- Add tests for JCasC


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -206,15 +206,19 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
         this.tags = tags;
         this.idleTerminationMinutes = idleTerminationMinutes;
         this.associatePublicIp = associatePublicIp;
-        this.connectionStrategy = connectionStrategy;
+        if (connectionStrategy == null) {
```
jvz: You could change all this line to just `this.connectionStrategy = connectionStrategy == null ? ConnectionStrategy.PRIVATE_IP : connectionStrategy;`

src/test/java/hudson/plugins/ec2/ConfigurationAsCodeTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Rule
+    public JenkinsConfiguredWithCodeRule j = new JenkinsConfiguredWithCodeRule();
+
+    @Test
+    @ConfiguredWithCode("EC2CloudEmpty.yml")
+    public void testEmptyConfig() throws Exception {
+        final AmazonEC2Cloud ec2Cloud = (AmazonEC2Cloud) Jenkins.get().getCloud("ec2-");
+        assertNotNull(ec2Cloud);
+        assertEquals(ec2Cloud.getTemplates().size(), 0);
```
jvz:
```suggestion
        assertEquals(0, ec2Cloud.getTemplates().size());
```

These assertions are usually ordered `assertEquals(expected, actual)`.

res0nance:
@casz are there any other cases of JCasC we should test for?

casz:
I don't actively use ec2 😓 But let me ask our gitter users 👍

casz:
Here is one exception log and yaml file from user:
https://gist.github.com/Shaps/efb4aed61955cd8301e2ae268d90dee6
https://gist.github.com/Shaps/be5503ec902d4ce0dbe4958d163b6902

timja:
Model export test docs:
https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#model-export-test

res0nance:
Thanks for your input guys I'll add an export test in a separate PR. I'll look into that yml failure as well

devblueray:
https://gist.github.com/devblueray/6980c4dfd3235f9ed5c2a71d1f46c4d5

Here's a gist of the error I am getting with the EC2 Plugin

res0nance:
@casz I got the yaml you linked working and added a MRE test case for it. Wondering if this was the right way to solve it.

@devblueray Thanks for the report do you have a `yml` file that will reproduce this issue.

Overall this PR is getting too big. I'll implement all the export stuff in a separate PR

devblueray:
I don't, this was from the export.  I am attempting to write the relevant config sections by hand to see if they work.

res0nance:
Alright I'll add an export test in a separate PR to deal with export issues. This PR is primarily to deal with the import issues

oleg-nenashev:
@res0nance Hi. What would be the release ETA?

res0nance:
Hey @oleg-nenashev I'm not a maintainer so I cant release it but #349, #296 and #360 probably should be merged before release. @thoulen does the releases so he would be able to speak about that.

PR #359, opened 2019-05-31
[JENKINS-57215] Fix stale uptime in EC2RetentionStrategy

datallah:
In my logs I have lots of cases where the uptime isn't correct in the EC2RetentionStrategy checking.

This fixes the uptime value to be not stale.

 * Additionally, add an early check to prevent the rest of the process if the instance is already in the target state (and remove the partial check from one code path)

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -124,6 +124,10 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
             // * We use stop-on-terminate and the instance is currently stopped or stopping
             if (InstanceState.TERMINATED.equals(state)
                   || computer.getSlaveTemplate().stopOnTerminate && (InstanceState.STOPPED.equals(state) || InstanceState.STOPPING.equals(state))) {
+                if (computer.isOnline()) {
+                    LOGGER.info("External Stop of " + computer.getName() + " detected - disconnecting. instance status" + state.toString());
+                    computer.disconnect(null);
+                }
                 return 1;
             }
             
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -110,14 +110,23 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
             InstanceState state;
 
             try {
+                state = computer.getState(); //Get State before Uptime because getState will refresh the cached EC2 info
                 uptime = computer.getUptime();
-                state  = computer.getState();
             } catch (AmazonClientException | InterruptedException e) {
                 // We'll just retry next time we test for idleness.
                 LOGGER.fine("Exception while checking host uptime for " + computer.getName()
                         + ", will retry next check. Exception: " + e);
                 return 1;
             }
+            
+            //Don't bother checking anything else if the instance is already in the desired state:
+            // * Already Terminated
+            // * We use stop-on-terminate and the instance is currently stopped or stopping
+            if (InstanceState.TERMINATED.equals(state)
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -110,14 +110,23 @@ private long internalCheck(EC2Computer computer) {
             InstanceState state;
 
             try {
+                state = computer.getState(); //Get State before Uptime because getState will refresh the cached EC2 info
                 uptime = computer.getUptime();
-                state  = computer.getState();
```
res0nance: I'm not getting this bit, unless `getUptime` somehow the instance changes its launch time. This should not change anything. The change is still harmless though.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -110,14 +110,23 @@ private long internalCheck(EC2Computer computer) {
...
                 LOGGER.fine("Exception while checking host uptime for " + computer.getName()
                         + ", will retry next check. Exception: " + e);
                 return 1;
             }
+            
+            //Don't bother checking anything else if the instance is already in the desired state:
+            // * Already Terminated
+            // * We use stop-on-terminate and the instance is currently stopped or stopping
+            if (InstanceState.TERMINATED.equals(state)
+                  || computer.getSlaveTemplate().stopOnTerminate && (InstanceState.STOPPED.equals(state) || InstanceState.STOPPING.equals(state))) {
```
res0nance: I'll defer this to others but I don't think this does the right thing in case the instance was stopped by external forces

datallah: Can you clarify what you are concerned about?

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
datallah: Yes, that's exactly right - getState() causes the stale EC2 instance "description" (including launch time) to be updated from AWS.

The instance description is cached unless explicitly flushed by a call to getState() or get updateInstanceDescription().  There are several cases where the cached value may not be accurate, of particular importance is when an instance is being launched.  When the instance is launched, a start is requested, but the instance description isn't flushed until we try to connect via ssh or WinRM (which is after the launch timeout).

What ends up happening as a result is that if the EC2RetentionStrategy happens between the time that a instance is initially launched and the launch timeout, it may appear to be idle based on the stale previous idle time and have the idle timeout triggered.  By switching the order so getState() refreshes the instance description this situation is avoided.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
res0nance: What if someone terminated an instance from AWS console. It now wont call `idletimeout` in those cases

datallah: Aha, it's true that this will be a slight behavior change in the case of an external EC2 instance state change.

In the case of an external Termination, the difference in behavior is that the node will not be removed from jenkins by this process, but the `EC2SlaveMonitor` thread will do so instead, so that should be OK (perhaps with slightly different timing, but I don't think that should be problematic).

Where there's a slightly more problematic change is the case of an external Stop - there doesn't appear to be an explicit "disconnect" unless you use `hudson.slaves.ConnectionActivityMonitor` (which I forgot isn't enabled by default).
I think it's probably worth a tweak to add an explicit disconnection for this case (and actually this can help us detect disconnections due to external stopping faster).  I will make an update for this.


It seems that

res0nance:
Relevant JIRA: JENKINS-57215

dawn-minion:
@jvz or @thoulen  Is it possible to get a new release made so this fix can be used? We keep hitting this issue, so would be a big help. Thanks!

PR #363, opened 2019-06-12
[JENKINS-53858] fix implementation of double-check-locking

res0nance:
- Rebase and supersede #349


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -163,7 +163,7 @@
 
     private transient KeyPair usableKeyPair;
 
-    protected transient AmazonEC2 connection;
+    protected volatile transient AmazonEC2 connection;
```
```suggestion
    protected transient volatile AmazonEC2 connection;
```

More common word ordering in Java.

PR #365, opened 2019-06-19
Reduce lock hotness

res0nance:
- Reduce the lock scope to when we need to modify the connection
- This change is aimed at reducing the time we hold the `Queue.lock` and `NodeProvisioner.provisioningLock` causing UI performance issues.


res0nance:
I had no way around this, any suggestions for improvement overall that we
can take up at a later time?

PR #367, opened 2019-06-20
fixing JENKINS-53952: wait for spot instances to get an instance id

mdodsworth:
Wait for the spot slave instance to get its instanceId set before attempting to bootstrap the node


src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -366,6 +366,13 @@ private Connection connectToSsh(EC2Computer computer, TaskListener listener) thr
                 }
                 String host = getEC2HostAddress(computer);
 
+                if ((computer.getNode() instanceof EC2SpotSlave) && computer.getInstanceId() == null) {
```
res0nance:
```suggestion
                if ((node instanceof EC2SpotSlave) && computer.getInstanceId() == null) {
```
`getNode()` is called above

PR #369, opened 2019-06-24
Increment dependencies

res0nance:
- Increment SDK dependency for users to see the new instance types
- Increment JCasC version to always test against more recent versions


res0nance:
retriggering CI

PR #370, opened 2019-06-24
Add trilead as a dependency

res0nance:
- Trilead is going to be removed from core at some point we use it so adding it as a dependency


res0nance:
Waiting on jenkinsci/jenkins#4085 not sure if anything else is required after?

PR #371, opened 2019-06-24
[JENKINS-58160] Don't use Guava Beta APIs

Tblue:
The EC2 plug-in currently uses Guava Beta APIs. Using Guava's Beta APIs is [strongly discouraged](https://github.com/google/guava#important-warnings) as they can change at any time.

In fact, this issue is blocking me from finishing [JENKINS-56839](https://issues.jenkins-ci.org/browse/JENKINS-56839): Jenkins core provides Guava 11, but a test dependency I am using pulls in Guava 20. While a bit messy, this wouldn't usually be an issue due to [Guava's API stability guarantee](https://github.com/google/guava#important-warnings); however, because the EC2 plug-in is using Guava Beta APIs that work differently in Guava 20 than they did in Guava 11, my tests for [JENKINS-56839](https://issues.jenkins-ci.org/browse/JENKINS-56839) break.

## Changes

I changed the following classes so that they don't rely on Guava Beta APIs anymore:

- `EC2HostAddressProvider` uses the [`HostAndPort`](https://guava.dev/releases/11.0.1/api/docs/com/google/common/net/HostAndPort.html) Beta API, but for no clear reason: The IP addresses that are parsed seem always to be plain IPv4 addresses without a port. Use a plain `String` instead.
- `WindowsProcess` uses the [`Closeables`](https://guava.dev/releases/11.0.1/api/docs/com/google/common/io/Closeables.html) Beta API to close resources quietly, without throwing an exception. What the `Closeables.closeQuietly()` method does is very easy to re-implement ourselves, so do just that.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2HostAddressProvider.java
+++ b/src/main/java/hudson/plugins/ec2/EC2HostAddressProvider.java
@@ -1,7 +1,6 @@
 package hudson.plugins.ec2;
 
 import com.amazonaws.services.ec2.model.Instance;
-import com.google.common.net.HostAndPort;
 import org.apache.commons.lang.StringUtils;
 
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


jvz:
The fact that Guava hasn't made any real efforts toward maintaining backward compatibility makes it impossible to upgrade Guava in Jenkins Core from my understanding. This is certainly welcome to make things simpler for tests, though, since you can use whatever version of Guava you want there.

Tblue:
@jvz:

> The fact that Guava hasn't made any real efforts toward maintaining backward compatibility makes it impossible to upgrade Guava in Jenkins Core from my understanding.

Indeed. I was originally hoping that this would allow us to use e. g. the latest Guava version (27), but it turns out that Guava 21 _did_ break compatibility by removing non-beta APIs (supposedly for the last time, though), which breaks Jenkins itself, so we are limited to at most Guava 20. Still, better than nothing.

Thanks for the approval!

jvz:
For some libraries that maintain backward compatibility, check out commons.apache.org. These do not cause classpath conflicts because they rename the packages if there's ever a backward incompatible change.

PR #375, opened 2019-06-27
[JENKINS-56839] Allow disabling T2/T3 Unlimited Mode for T3 instances

Tblue:
Quoting the [JIRA issue](https://issues.jenkins-ci.org/browse/JENKINS-56839):

> AWS' new T3 instance type is set to run with T2/T3 Unlimited enabled.
>
> The way that the current EC2 plugin is designed, we cannot create T3 slaves which have Unlimited disabled. The EC2 plugin has a checkbox to enable/disable. However, it uses the default if this is unchecked.
>
> The EC2 plugin should probably take that setting exactly and use it to set the value. Thus, to use T3 instances with bursting, the Jenkins admin must manually check the box.
>
> Alternatively, you could have a drop-down with options like: 1. Enabled 2. Disabled 3. Instance-type default

This pull request implements the latter suggestion. Instead of a simple checkbox for Unlimited Mode, there now is a drop-down, allowing the administrator to choose whether to explicitly enable/disable Unlimited Mode, or whether to use the AWS default for the chosen instance type (by not specifying any preference for Unlimited Mode when creating instances).

The old boolean setting is migrated to preserve existing configurations.


Tblue:
It seems like the Windows node used to run checks for pull requests is unreliable:

> [2019-06-27T10:37:27.401Z] win2012-467eb0 was marked offline: Connection was broken: java.util.concurrent.TimeoutException: Ping started at 1561631607399 hasn't completed by 1561631847400

Is there any way for me to trigger a re-build? I logged into `ci.jenkins.io`, but don't seem to have the necessary permissions.

res0nance:
Close and open PR should trigger a rebuild

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -318,6 +327,15 @@ public EC2Cloud getParent() {
         return parent;
     }
 
+    public BurstableUnlimitedMode getBurstableUnlimitedMode() {
+        if (burstableUnlimitedMode == null) {
+            // For backward-compatibility.
+            return t2Unlimited ? BurstableUnlimitedMode.ENABLED : BurstableUnlimitedMode.DEFAULT;
+        }
+
+        return burstableUnlimitedMode;
```
res0nance: Yous should resolve this bit in `readResolve`

thoulen:
I would prefer to use a simple solution, that is to set the value all the time and not to change the interface

Tblue:
@thoulen, all right, I can do that.

Tblue:
@thoulen: While implementing your suggestion, testing showed that AWS allows specifying a CPU credit option (i. e. the Unlimited Mode option) _only_ for T2/T3 instances. Specifying _any_ CPU credit option for other instance types, no matter the value of the option, leads to AWS rejecting the API request: The only solution is to not specify any CPU credit option at all in this case.

So it looks like we do need the "Use Instance Type Default" option after all so that people can properly use instance types other than T2/T3.

@res0nance: I'm not familar with JCasC, but shouldn't things continue working as-is as long as `burstableUnlimitedMode` is unset in the JCasC configuration file? I will see if I can add a test for that. I will also look into your other suggestions -- thanks!

thoulen:
Thanks for the try

Tblue:
The build is failing because https://repo.azure.jenkins.io/ is down. :(

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -175,7 +181,8 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
             String instanceCapStr, String iamInstanceProfile, boolean deleteRootOnTermination,
             boolean useEphemeralDevices, boolean useDedicatedTenancy, String launchTimeoutStr, boolean associatePublicIp,
             String customDeviceMapping, boolean connectBySSHProcess, boolean monitoring,
-            boolean t2Unlimited, ConnectionStrategy connectionStrategy, int maxTotalUses) {
+            boolean t2Unlimited, BurstableUnlimitedMode burstableUnlimitedMode, ConnectionStrategy connectionStrategy,
```
Tblue: I wonder whether `t2Unlimited` should maybe be removed completely from this constructor and be made into a (deprecated) `DataBoundSetter` instead (to make it clear that `t2Unlimited` is deprecated; its presence in the `DataBoundConstructor` signature might be confusing).

Tblue:
@res0nance: I addresses your suggestions; please have a look.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
Tblue: I just realized that this would break the migration of `t2Unlimited`, unless we move that from `readResolve()` back to e. g. `getBurstableUnlimitedMode()`... So let's leave things like they are for now.

res0nance: You can do it this way, it removes the Deprecated field from the ctor to make it less confusing.
https://github.com/jenkinsci/ec2-plugin/blob/597af44906321da78a92fd355355b77f5751a679/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L423

Tblue: Ah, right. I will look into that.

PR #377, opened 2019-07-06
Added minimum uphours feature

snallami:
Added minimum uphours feature so that agent stays up for specified minimum hours before considering for termination. Once agent stayed for minimum number of hours other rules (like idle termination time) will get evaluated.

**Why this feature is important?**

Main use cases.

1. Avoid agent spin up time. Windows agents in AWS takes somewhere around 10 to 20 minutes to spin up and without minimum uptime feature we see that builds are waiting longer for agent to be available. Linus agents spin up time is faster though compared to windows.

2. Warm up ephemeral agents with caches(maven or npm or any other) so that actual developer builds  can be faster and with the help of minimum uphours we can retain agent for 8 to 10 hours which helps in faster builds during development hours or working hours.

3. Reserve capacity upfront for pre planned work.

4. No need to depend on static agent for cache warmup.

And this is optional feature

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
@@ -32,29 +32,29 @@ public final class EC2OndemandSlave extends EC2AbstractSlave {
     private static final Logger LOGGER = Logger.getLogger(EC2OndemandSlave.class.getName());
 
     @Deprecated
-    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType)
+    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, String minUpHours)
             throws FormException, IOException {
-        this(instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, false, launchTimeout, amiType);
+        this(instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, false, launchTimeout, amiType, minUpHours);
     }
 
     @Deprecated
-    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, String minUpHours)
             throws FormException, IOException {
-        this(instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, false, useDedicatedTenancy, launchTimeout, amiType);
+        this(instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, false, useDedicatedTenancy, launchTimeout, amiType, minUpHours);
     }
 
     @Deprecated
-    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, String minUpHours)
             throws FormException, IOException {
-        this(description + " (" + instanceId + ")", instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, Collections.emptyList(), remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, useDedicatedTenancy, launchTimeout, amiType, ConnectionStrategy.backwardsCompatible(usePrivateDnsName, false, false), -1);
+        this(description + " (" + instanceId + ")", instanceId, description, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, Collections.emptyList(), remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, useDedicatedTenancy, launchTimeout, amiType, ConnectionStrategy.backwardsCompatible(usePrivateDnsName, false, false), -1, minUpHours);
     }
 
     @DataBoundConstructor
-    public EC2OndemandSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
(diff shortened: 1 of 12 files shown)


src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -161,6 +177,12 @@ private long internalCheck(EC2Computer computer) {
             if (computer.isOffline() && uptime < TimeUnit.MINUTES.toMillis(STARTUP_TIMEOUT)) {
                 return 1;
             }
+            
+            // Check min number of hours instance must be up - if stayed up at least that minimum.
+            long minUpHoursInMills = TimeUnit.MILLISECONDS.toHours(minUpHours);
+            if (minUpHoursInMills > uptime) {
```
res0nance: If the minuphours is 0 the instance never terminates?

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -32,29 +32,29 @@
     private static final Logger LOGGER = Logger.getLogger(EC2OndemandSlave.class.getName());
 
     @Deprecated
-    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType)
+    public EC2OndemandSlave(String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, String minUpHours)
```
res0nance: Should not add the new option to deprecated constructors

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -415,6 +418,10 @@ public String getCurrentSubnetId() {
     public boolean getAssociatePublicIp() {
         return associatePublicIp;
     }
+    
+    public String getMinUpHours() {
+ 		return minUpHours;
+ 	}
```
res0nance: Mind the indentation

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -33,23 +33,23 @@
     private final String spotInstanceRequestId;
 
     @Deprecated
-    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType)
+    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, String minUpHours)
```
res0nance: Same here on deprecated constructors

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -175,7 +177,7 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
             String instanceCapStr, String iamInstanceProfile, boolean deleteRootOnTermination,
             boolean useEphemeralDevices, boolean useDedicatedTenancy, String launchTimeoutStr, boolean associatePublicIp,
             String customDeviceMapping, boolean connectBySSHProcess, boolean monitoring,
-            boolean t2Unlimited, ConnectionStrategy connectionStrategy, int maxTotalUses) {
+            boolean t2Unlimited, ConnectionStrategy connectionStrategy, int maxTotalUses, String minUpHours) {
```
res0nance: Modifying the ctor this way also breaks backwards compat with groovy scripts

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
snallami: let me see if i can bypass them

snallami:
For some odd reason, I am not able to comment on review for minuphours=0 inline. Adding comment here.

If minuphours=0 then code evaluates other rule like idle termination time. See code @ 
https://gist.github.com/snallami/d917ec049e5d80e4206e2ce0cd2ef895

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
snallami: Do you want me to add one more overloaded constructor?

snallami:
Thanks for review. I updated code as per review comments. Adding notes here as well.

1) If minuphours=0 then code evaluates other rule like idle termination time. See code @
https://gist.github.com/snallami/d917ec049e5d80e4206e2ce0cd2ef895

**Need clarification**

2) I am not clear about what are the next steps about review comment "Modifying the ctor this way also breaks backwards compat with groovy scripts". 

Let me know if you want me to add an overloaded constructor or modify groovy scripts.

3) I will look into better way to add tests since check method always return 1

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
res0nance: Add another and deprecate the older one, is how it usually is handled. The older constructors should not take in new params and should provide a safe default. i.e the old behaviour.

res0nance:
A good start to tests is an old testcase that would otherwise timeout should not timeout any longer.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
snallami: hmmm , not sure why my eclipse is messing up formatting only for this line. Let me use text editor in next commit

Jenkinsfile
```diff
@@ -1 +0,0 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
```
snallami: Not sure why this got deleted in my commit. let me submit a new commit

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -73,7 +76,7 @@
                     String.valueOf(STARTUP_TIME_DEFAULT_VALUE)), STARTUP_TIME_DEFAULT_VALUE);
 
     @DataBoundConstructor
-    public EC2RetentionStrategy(String idleTerminationMinutes) {
+    public EC2RetentionStrategy(String idleTerminationMinutes, String minUpHours) {
```
jvz: You need to keep the old constructors and mark them `@Deprecated` in order to preserve backward compatibility as this class is public.

Also, you should use an argument type of either `int` (for a required value) or `Integer` (for a nullable one). I'd suggest `OptionalInt` as an argument type for the latter, though I'm not sure if that actually works with `@DataBoundConstructor`.

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -46,15 +46,15 @@ public EC2OndemandSlave(String instanceId, String description, String remoteFS,
...
     }
 
     @DataBoundConstructor
-    public EC2OndemandSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
+    public EC2OndemandSlave(String name, String instanceId, String description, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses, String minUpHours)
```
jvz: Same here; you need to create a new constructor and mark the old one `@Deprecated`.

I'll note that this sort of redundancy could be avoided by using a value object as the constructor argument rather than all of its attributes, but this would require a larger effort to fix at this point.

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -41,15 +41,15 @@ public EC2SpotSlave(String name, String spotInstanceRequestId, String descriptio
...
     }
 
     @DataBoundConstructor
-    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
+    public EC2SpotSlave(String name, String spotInstanceRequestId, String description, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses, String minUpHours)
```
jvz: Same here, need another constructor.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
jvz: Same here about the constructor. Same comment about the type (use `int` or `Integer`).

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -140,6 +140,8 @@
 
     public int nextSubnet;
 
+    private String minUpHours;
```
jvz: This should be `Integer` for a nullable value, or `int` for a non-null value. Type conversion is handled automatically for you at the point of `@DataBoundConstructor` and `@DataBoundSetter`, so you can maintain the real type throughout your code.

src/main/resources/hudson/plugins/ec2/SlaveTemplate/config.jelly
```diff
@@ -177,6 +177,10 @@ THE SOFTWARE.
     <f:entry title="${%Maximum Total Uses}" field="maxTotalUses">
       <f:textbox default="-1"/>
     </f:entry>
+    
+    	<f:entry title="Minimum uptime in hours" field="minUpHours">
+		<f:textbox default="0" />
```
jvz: This should use `<f:number>` instead. In fact, all the other textboxes in this form that are numeric should also use that.

src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-minUpHours.html
```diff
@@ -0,0 +1,6 @@
+<div>
+	<p> Retains ephemeral agents for specified hours before applying "idle termination minutes" rules.
```
jvz:
```suggestion
    <p>Retains ephemeral agents for specified hours before applying "idle termination minutes" rules.</p>
```

PR #360, opened 2019-06-03
Test case for the PR296

janario:
https://github.com/jenkinsci/ec2-plugin/pull/296

PR #378, opened 2019-07-15
Only check credentials once a minute

res0nance:
- Fix an issue introduced in #360 where calling `describeInstances` too often will eventually lead to locks being held for excessive time. This will cause `connect` to only check the credentials once a minute.
- #365 was attempting to mitigate this but the bottleneck changed to the `describeInstance` call

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -46,7 +46,6 @@ import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
-import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -66,6 +65,8 @@ import java.util.logging.SimpleFormatter;
 
 import javax.servlet.ServletException;
 
+import hudson.Extension;
+import hudson.model.PeriodicWork;
 import hudson.model.TaskListener;
 import hudson.util.ListBoxModel;
 import jenkins.model.Jenkins;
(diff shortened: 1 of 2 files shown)


res0nance:
I'll look into PeriodicWork thanks

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -975,4 +971,32 @@ public static void log(Logger logger, Level level, TaskListener listener, String
...
+        }
+
+        @Override
+        protected void doRun() throws IOException {
+            Jenkins instance = Jenkins.get();
+            if (instance.clouds != null) {
+                for (Cloud cloud : instance.clouds) {
+                    if (cloud instanceof EC2Cloud) {
+                        EC2Cloud ec2_cloud = (EC2Cloud) cloud;
+                        LOGGER.finer("Checking EC2 Connection on: " + ec2_cloud.getDisplayName());
```
```suggestion
                        LOGGER.finer(() -> "Checking EC2 Connection on: " + ec2_cloud.getDisplayName());
```

Lazy logging message construction for lower log levels.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -975,4 +971,32 @@ public static void log(Logger logger, Level level, TaskListener listener, String
...
+                        try {
+                            if(ec2_cloud.connection != null) {
+                                ec2_cloud.connection.describeInstances();
+                            }
+                        } catch (AmazonClientException e) {
+                            LOGGER.finer("Reconnecting to EC2 on: " + ec2_cloud.getDisplayName());
```
```suggestion
                            LOGGER.finer(() -> "Reconnecting to EC2 on: " + ec2_cloud.getDisplayName());
```

Same here.

res0nance:
I tested it with the additions manually and it seems to work fine

PR #379, opened 2019-07-17
Remove test code in production code

erikhakansson:
@jvz Please checkout this review, as per our discussion in #343 

I've broken out the EC2Cloud.connect() method to a factory that's injected with Guice and created a custom Guice module for binding that factory. That way, the actual implementation is overridable using a System property and tests can use a factory that mocks AmazonEC2.

I also refactored a bit of the existing production code and tests, getting rid of the isTestMode() in AmazonEC2Cloud and reduced the visibility of a few fields that were accessed directly from tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@ THE SOFTWARE.
         <jenkins.version>2.150.1</jenkins.version>
         <java.level>8</java.level>
         <jcasc.version>1.22</jcasc.version>
+        <aws-java-sdk.version>1.11.562</aws-java-sdk.version>
     </properties>
 
     <dependencies>
@@ -101,7 +102,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>aws-java-sdk</artifactId>
-            <version>1.11.562</version>
+            <version>${aws-java-sdk.version}</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -235,6 +236,17 @@ THE SOFTWARE.
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.amazonaws</groupId>
+                <artifactId>aws-java-sdk-bom</artifactId>
+                <version>${aws-java-sdk.version}</version>
+                <type>pom</type>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
     <repositories>
         <repository>
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -35,6 +35,7 @@ import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
 import com.cloudbees.plugins.credentials.domains.Domain;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import hudson.plugins.ec2.util.AmazonEC2Factory;
 import hudson.security.ACL;
 
 import java.io.BufferedReader;
@@ -83,7 +84,6 @@ import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.auth.InstanceProfileCredentialsProvider;
 import com.amazonaws.internal.StaticCredentialsProvider;
 import com.amazonaws.services.ec2.AmazonEC2;
-import com.amazonaws.services.ec2.AmazonEC2Client;
 import com.amazonaws.services.ec2.model.CreateKeyPairRequest;
 import com.amazonaws.services.ec2.model.DescribeSpotInstanceRequestsRequest;
 import com.amazonaws.services.ec2.model.Filter;
@@ -152,20 +152,18 @@ public abstract class EC2Cloud extends Cloud {
     @Deprecated
     private transient Secret secretKey;
 
-    protected final EC2PrivateKey privateKey;
+    private final EC2PrivateKey privateKey;
 
     /**
      * Upper bound on how many instances we may provision.
      */
-    public final int instanceCap;
+    private final int instanceCap;
 
     private final List<? extends SlaveTemplate> templates;
 
     private transient KeyPair usableKeyPair;
 
-    protected transient volatile AmazonEC2 connection;
-
-    private static AWSCredentialsProvider awsCredentialsProvider;
+    private transient volatile AmazonEC2 connection;
 
     protected EC2Cloud(String id, boolean useInstanceProfileForCredentials, String credentialsId, String privateKey,
             String instanceCapStr, List<? extends SlaveTemplate> templates, String roleArn, String roleSessionName) {
@@ -195,7 +193,7 @@ public abstract class EC2Cloud extends Cloud {
 
     public abstract URL getS3EndpointUrl() throws IOException;
 
-    protected Object readResolve() {
+    private Object readResolve() {
         this.slaveCountingLock = new ReentrantLock();
         for (SlaveTemplate t : templates)
             t.parent = this;
@@ -271,6 +269,10 @@ public abstract class EC2Cloud extends Cloud {
             return "";
         else
             return String.valueOf(instanceCap);
+    }
+
+    public int getInstanceCap() {
+        return instanceCap;
     }
 
     public List<SlaveTemplate> getTemplates() {
@@ -751,7 +753,7 @@ public abstract class EC2Cloud extends Cloud {
 
     private AmazonEC2 reconnectToEc2() throws IOException {
         synchronized(this) {
-            connection = connect(createCredentialsProvider(), getEc2EndpointUrl());
+            connection = AmazonEC2Factory.getInstance().connect(createCredentialsProvider(), getEc2EndpointUrl());
             return connection;
         }
     }
@@ -776,18 +778,6 @@ public abstract class EC2Cloud extends Cloud {
         } catch (IOException e) {
             throw new AmazonClientException("Failed to retrieve the endpoint", e);
         }
-    }
-
-    /***
-     * Connect to an EC2 instance.
-     *
-     * @return {@link AmazonEC2} client
-     */
-    public static AmazonEC2 connect(AWSCredentialsProvider credentialsProvider, URL endpoint) {
-        awsCredentialsProvider = credentialsProvider;
-        AmazonEC2 client = new AmazonEC2Client(credentialsProvider, createClientConfiguration(endpoint.getHost()));
-        client.setEndpoint(endpoint.toString());
-        return client;
     }
 
     public static ClientConfiguration createClientConfiguration(final String host) {
@@ -836,7 +826,7 @@ public abstract class EC2Cloud extends Cloud {
      * @param path String like "/bucketName/folder/folder/abc.txt" that represents the resource to request.
      */
     public URL buildPresignedURL(String path) throws AmazonClientException {
-        AWSCredentials credentials = awsCredentialsProvider.getCredentials();
+        AWSCredentials credentials = createCredentialsProvider().getCredentials();
         long expires = System.currentTimeMillis() + 60 * 60 * 1000;
         GeneratePresignedUrlRequest request = new GeneratePresignedUrlRequest(path, credentials.getAWSSecretKey());
         request.setExpiration(new Date(expires));
@@ -893,7 +883,7 @@ public abstract class EC2Cloud extends Cloud {
                 throws IOException, ServletException {
             try {
                 AWSCredentialsProvider credentialsProvider = createCredentialsProvider(useInstanceProfileForCredentials, credentialsId, roleArn, roleSessionName, region);
-                AmazonEC2 ec2 = connect(credentialsProvider, ec2endpoint);
+                AmazonEC2 ec2 = AmazonEC2Factory.getInstance().connect(credentialsProvider, ec2endpoint);
                 ec2.describeInstances();
 
                 if (privateKey == null)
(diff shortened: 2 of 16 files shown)


src/main/java/hudson/plugins/ec2/util/AmazonEC2Factory.java
```diff
@@ -0,0 +1,12 @@
+package hudson.plugins.ec2.util;
+
+import java.net.URL;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.services.ec2.AmazonEC2;
+
+public abstract class AmazonEC2Factory {
```
jvz: You can make this an `interface` to simplify this class a bit.

src/main/java/hudson/plugins/ec2/util/AmazonEC2Factory.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public abstract AmazonEC2 connect(AWSCredentialsProvider credentialsProvider, URL ec2Endpoint);
```
jvz: If we're really going to use Guice properly, we're going to want to make these values also injected. The AWSCredentialsProvider is one type, though you'll need to create an annotation to make the URL injectable (or maybe use `@Named`?).

Once you get to that level, this factory class is no longer needed and can be converted into the Guice module you added.

src/main/java/hudson/plugins/ec2/util/EC2GuiceModule.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Singleton;
+
+import hudson.Extension;
+
+@Extension
+public class EC2GuiceModule extends AbstractModule {
+    @Override
+    protected void configure() {
```
jvz: You're going to want to bind some real things instead of configuring the class name like this. See https://github.com/google/guice/wiki/LinkedBindings for some introductory info. Also:

* https://github.com/google/guice/wiki/ProvidesMethods for the method that returns an `AmazonEC2` instance
* https://github.com/google/guice/wiki/InstanceBindings for the URL value binding.

Or, if you want to really get fancy similar to what you're doing here, see https://github.com/google/guice/wiki/JustInTimeBindings

src/test/java/hudson/plugins/ec2/util/AmazonEC2FactoryMockImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import org.mockito.stubbing.Answer;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.services.ec2.AmazonEC2;
+import com.amazonaws.services.ec2.model.DescribeRegionsResult;
+import com.amazonaws.services.ec2.model.Region;
+
+import hudson.plugins.ec2.EC2Cloud;
+
+public class AmazonEC2FactoryMockImpl extends AmazonEC2Factory {
```
jvz: Then, once you've adapted the production Guice module class, you can specify a test version here to override the `AmazonEC2` instance returned.

src/main/java/hudson/plugins/ec2/util/AmazonEC2Factory.java
erikhakansson: I had it as an interface at first but it interfered with some Guice module provided by Jenkins that was scanning all potential @Injected classes and looking at their superclasses. So when annotating an interface field with @Inject, this module would (amongst other things) run .getSuperClass() on that, which returns null for an interface, and Object for a class, resulting in a NullPointerException.

I did some refactoring after that though, so I might have built the issue away without knowing it. I'll try it out.

src/main/java/hudson/plugins/ec2/util/AmazonEC2Factory.java
erikhakansson: Should be simple enough to implement, though I'm a bit unsure on how to work with `@Named` but I'm sure I'll figure it out.

src/main/java/hudson/plugins/ec2/util/EC2GuiceModule.java
erikhakansson: I'm not sure I follow entirely. I do understand the `@Provides` part, but won't that interfere with my ability to override actual implementation class?

As far as I understand it, `@Provides` does basically the same as `bind(MethodReturnType.class).to(ImplementationType.class)`. I will be able to get rid of the factory that way though.

However, the good part about having the Factory injectable, but not the AmazonEC2 implementation is that the `AmazonEC2Client` class is in the AWS SDK and has no default constructor but requires a lot of parameters. Thus, for the real implementation, I need to supply all that stuff in the `new AmazonEC2Client(.....)` but in the mock factory I can just do `Mockito.mock(AmazonEC2.class)`

erikhakansson: And if I move the instance creation to a `@Provides` method, that would require `Mockito.mock` in production code, right? Or am I misinterpreting something?

Is it possible to have two `@Provides` and give the test one priority when running tests?

erikhakansson: I may have found the answer in an old Guice issue:

https://github.com/google/guice/issues/1029

`Modules.override()` should be what I need. I'll just have a production module and a test module.

jvz: Right, you want two different Module classes here as they essentially correspond to a configuration.

erikhakansson:
@jvz after reading through Guice documentation a bit more, it really seems that a Factory with a `create()` method is the preferred method for doing what we're attempting here. 

Looking at https://github.com/google/guice/wiki/InstanceBindings for example, that seems to only apply to configuration fields that are not set in runtime.

However, both the `AWSCredentialsFactory` and `ec2EndpointURL` are constructed on retrieval within `EC2Cloud` which is not a Guice module (although a Guice object). Hence, I can't call `bind(String.class).annotatedWith(Names.named("ec2EndpointURL")).toInstance(getEc2EndpointUrl())` from within `EC2Cloud`.

One solution would be to `@Inject` the `EC2Cloud` instance in the `@Provides AmazonEC2` method and just use getters, but that would create a circular dependency since `EC2Cloud` depends on `AmazonEC2`.

I'm going to experiment some more, but I'm thinking the best way might be to retain the factory, but replace the explicit class name lookup with an overriding test Module like we discussed.

jvz:
Great investigative work so far! It seems you might be stepping into a rabbit hole of untying a mess of implicit dependencies here, though maybe it's just some simple refactoring. Keep me posted.

erikhakansson:
I'm afraid I've hit even more walls. It seems like `Modules.override` is specifically made for unit testing of modules. I'e, `Modules.override(new ModuleThatIsToBeTested()).with(new ModuleThatJustOverridesAFewBindingsForTestingPurposes())`. It creates a synthesized module with bindings in the first overridden by the second. If a module was already registered with `Guice.createInjector()` then this won't override.

And unfortunately all registration of modules happens dynamically at Jenkins startup (in `hudson.ExtensionFinder` specifically). I have no way of intercepting that.

I.e. I think that having two separate modules won't work.

jvz:
Aw, that's disappointing. Thanks for looking into this, though! I suppose we'll have to figure out a design pattern to use that abstracts this a bit.

Worst case scenario, you could make an extension point that's the connection factory which would be easier to override.

erikhakansson:
I'm going to have a look at this:

https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/ExtensionFilter.java

I haven't tried it out yet, but perhaps I can use it to filter out the production version of the factory when running tests. I.e. doing what you said about extension point and filter with this.

I believe I have to abandon Guice, though, because from my brief look at the code, it doesn't seem to work with Guice modules.

erikhakansson:
@jvz I've refactored and thrown out Guice in favour of `ExtensionPoint` and `ExtensionFilter` now and it works well.

Can you have a look now and see what you think?

erikhakansson:
Also, let me know if I should squash the commits 😄

pom.xml
```diff
@@ -236,6 +237,17 @@ THE SOFTWARE.
         </dependency>
     </dependencies>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.amazonaws</groupId>
+                <artifactId>aws-java-sdk-bom</artifactId>
+                <version>${aws-java-sdk.vesion}</version>
```
jvz: There's a typo here, and I'm not sure this dependencyManagement section is needed at all.

src/test/java/hudson/plugins/ec2/util/AmazonEC2FactoryExtensionFilter.java
```diff
@@ -0,0 +1,20 @@
...
+
+import hudson.Extension;
+import hudson.ExtensionComponent;
+import jenkins.ExtensionFilter;
+
+/**
+ * The sole purpose of this class is to filter out {@link AmazonEC2FactoryImpl} so as to avoid doing real calls to
+ * EC2 when running tests.
+ */
+@Extension
```
jvz:
```suggestion
@TestExtension
```

src/test/java/hudson/plugins/ec2/util/AmazonEC2FactoryMockImpl.java
```diff
@@ -0,0 +1,68 @@
...
+import org.mockito.stubbing.Answer;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.services.ec2.AmazonEC2;
+import com.amazonaws.services.ec2.model.DescribeRegionsResult;
+import com.amazonaws.services.ec2.model.Region;
+
+import hudson.plugins.ec2.EC2Cloud;
+
+@Extension
```
jvz:
```suggestion
@TestExtension
```

jvz: Also, by using this annotation, you might be able to make or use a more generic `ExtensionFilter` for test extensions or something like that.

pom.xml
erikhakansson: Ah, I see the typo. I'll fix it.

Regarding the dependencyManagement, it's a trick I learned from work. There's something weird with the `aws-java-sdk` (which is a transitive dependency through the `org.jenkins-ci.plugins:aws-java-sdk`) maven dependency that causes maven to (in some unknown circumstances) download a lot of old versions of the various dependencies. I.e, if you have version `1.11.562` of the dependency maven will still download e.g version `1.11.4xx` and a bunch of other versions. On an empty m2 repository this takes a loooot of time.

For some reason, adding the bom eliminates this problem. I have no idea why, and I have no idea why not everyone has the problem.

I can remove it if you want, but it shouldn't do any harm and it will reduce initial build time for some.

src/test/java/hudson/plugins/ec2/util/AmazonEC2FactoryExtensionFilter.java
erikhakansson: I just tried `@TestExtension` out, but Jenkins won't pick it up then.

https://wiki.jenkins.io/display/JENKINS/Unit+Test

seems to suggest that `@TestExtension` should be used as an inner class for a specific unit test. I.e. I think each test class that needs them would need their own inner class implementation for that to work.

pom.xml
jvz:
```suggestion
                <version>${aws-java-sdk.version}</version>
```

src/test/java/hudson/plugins/ec2/util/AmazonEC2FactoryExtensionFilter.java
jvz: Oh, I did not realize that. Ah well. An alternative would be yet another annotation, but I think that's just adding unnecessary code at that point. You can leave this as `@Extension`.

src/main/java/hudson/plugins/ec2/util/AmazonEC2Factory.java
```diff
@@ -0,0 +1,29 @@
+package hudson.plugins.ec2.util;
+
+import java.net.URL;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.services.ec2.AmazonEC2;
+import hudson.ExtensionPoint;
+import jenkins.model.Jenkins;
+
+public abstract class AmazonEC2Factory implements ExtensionPoint {
```
jvz: Just change this to an interface instead of an abstract class and we're good to go!

erikhakansson: Fixed now. Also did some cleanup and fixed my commit author since I did this work from two different computers.

erikhakansson: Should I squish the commits too?

erikhakansson: squash*

jvz: Yes, go ahead.

erikhakansson:
There we go, @jvz  😄

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -195,7 +193,7 @@ protected EC2Cloud(String id, boolean useInstanceProfileForCredentials, String c
 
     public abstract URL getS3EndpointUrl() throws IOException;
 
-    protected Object readResolve() {
+    private Object readResolve() {
```
jvz: As this class is abstract, I think this should remain `protected`.

erikhakansson: Alright, reverted that part

erikhakansson:
@jvz thanks for the help so far. What's the process now? I understand one more approval is needed? It would be cool to have this merged so I can get started with fixing the actual PR I wanted to push :)

jvz:
Yes, one more approval to merge. @thoulen typically reviews PRs in batches, or someone else might review this first.

erikhakansson:
Ok, good to know! Thanks!

PR #382, opened 2019-07-25
Allow defining node properties on template

jmendeth:
Proposal to allow defining node properties (environment variables, tool locations, etc.) that will be applied to created slaves.

`EC2AbstractSlave` and children already take a `nodeProperties` parameter, so we just need to add the field to `SlaveTemplate`.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -153,7 +153,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
     public int maxTotalUses;
 
-    public /* almost final */DescribableList<NodeProperty<?>, NodePropertyDescriptor> nodeProperties;
+    private /* lazily initialized */ DescribableList<NodeProperty<?>, NodePropertyDescriptor> nodeProperties;
 
     public int nextSubnet;
 
@@ -586,8 +586,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
     }
 
     public DescribableList<NodeProperty<?>, NodePropertyDescriptor> getNodeProperties() {
-        assert nodeProperties != null;
-    	return nodeProperties;
+    	return Objects.requireNonNull(nodeProperties);
     }
 
     public enum ProvisionOptions { ALLOW_CREATE, FORCE_CREATE }
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -1329,7 +1329,6 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
             maxTotalUses = -1;
         }
 
-         // 1.45 new parameters
         if (nodeProperties == null) {
             nodeProperties = new DescribableList<>(Saveable.NOOP);
         }
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -229,7 +229,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
         this.useDedicatedTenancy = useDedicatedTenancy;
         this.connectBySSHProcess = connectBySSHProcess;
         this.maxTotalUses = maxTotalUses;
-        this.nodeProperties = new DescribableList<NodeProperty<?>,NodePropertyDescriptor>(Saveable.NOOP, nodeProperties);
+        this.nodeProperties = new DescribableList<>(Saveable.NOOP, Util.fixNull(nodeProperties));
         this.monitoring = monitoring;
         this.nextSubnet = 0;
 
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -32,6 +32,7 @@ import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -67,6 +67,9 @@ import hudson.Util;
 import hudson.model.*;
 import hudson.model.Descriptor.FormException;
 import hudson.model.labels.LabelAtom;
+import hudson.slaves.NodeProperty;
+import hudson.slaves.NodePropertyDescriptor;
+import hudson.util.DescribableList;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 
@@ -150,6 +153,8 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
     public int maxTotalUses;
 
+    public /* almost final */DescribableList<NodeProperty<?>, NodePropertyDescriptor> nodeProperties;
+
     public int nextSubnet;
 
     public String currentSubnetId;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -138,6 +141,8 @@
 
     public int maxTotalUses;
 
+    public /* almost final */DescribableList<NodeProperty<?>, NodePropertyDescriptor> nodeProperties;
```
```suggestion
    private /* lazily initialized */ DescribableList<NodeProperty<?>, NodePropertyDescriptor> nodeProperties;
```

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -525,6 +547,15 @@ public int getMaxTotalUses() {
         return maxTotalUses;
     }
 
+    public List<NodePropertyDescriptor> getNodePropertyDescriptors() {
+        return NodePropertyDescriptor.for_(NodeProperty.all(), EC2AbstractSlave.class);
+    }
+
+    public DescribableList<NodeProperty<?>, NodePropertyDescriptor> getNodeProperties() {
+        assert nodeProperties != null;
+    	return nodeProperties;
```
```suggestion
    	return Objects.requireNonNull(nodeProperties);
```

Then you can remove the assertion above.

jmendeth:
Thanks for the quick review! I swear I had made that field private... Both changes committed and tested.

res0nance:
Retrigger-CI

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -525,6 +547,14 @@ public int getMaxTotalUses() {
         return maxTotalUses;
     }
 
+    public List<NodePropertyDescriptor> getNodePropertyDescriptors() {
+        return NodePropertyDescriptor.for_(NodeProperty.all(), EC2AbstractSlave.class);
+    }
+
+    public DescribableList<NodeProperty<?>, NodePropertyDescriptor> getNodeProperties() {
+    	return Objects.requireNonNull(nodeProperties);
```
res0nance: This line seems to not compile could this be fixed up?

jmendeth: It seems 9c26fdc32 broke it, I have rebased and fixed ^^

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -219,6 +225,7 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
         this.useDedicatedTenancy = useDedicatedTenancy;
         this.connectBySSHProcess = connectBySSHProcess;
         this.maxTotalUses = maxTotalUses;
+        this.nodeProperties = new DescribableList<NodeProperty<?>,NodePropertyDescriptor>(Saveable.NOOP, nodeProperties);
```
ro-kr: - the type parameters on the constructor are redundant, using <> should be sufficient.
- the nodeProperties argument can be null (e.g. in unit tests), this causes a NullPointerException in the constructor of the DescribableList. Wrapping the argument with Util.fixNull(...) should prevent this.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1298,6 +1329,11 @@ protected Object readResolve() {
             maxTotalUses = -1;
         }
 
+         // 1.45 new parameters
```
ro-kr: The version number in the comment is out of date. Maybe the just remove the comment because we cannot know when the PR will be released?

jmendeth: Removed!

jhansche:
@mildsunrise @res0nance Does this resolve https://issues.jenkins-ci.org/browse/JENKINS-36544?

The merge commit shows it is included in tags: `ec2-1.49.1` `ec2-1.49` `ec2-1.48` `ec2-1.47`. And according to the config.jelly change, this should be present between the "Maximum Total Uses" field, and the repeatable-delete button.

I'm running v1.49.1, but I do not see the Node Properties option where it looks like it should be. Am I missing something?

jhansche:
I see the Node Properties listed if I click Configure on a slave instance that was already launched. But I assumed this should appear in the "Configure Clouds" template -- and I don't see it in that template.

mildsunrise:
That's strange, I'm running v1.49.1 and I do see the node properties section at the bottom of the AMI definition in the advanced properties 🤔

jhansche:
OK, that worries me then 😶
I was poking around with the Script Console, and it _looks_ to me like everything should be right. But `Jenkins.instance.getCloud("ec2-$name").templates.each { it.nodeProperties }` is just an empty list. `nodePropertyDescriptors` contains the 4 NodeProperty describables that are available via core/plugins (including Env vars, which is what I'm trying to get working).

I also tried editing `config.xml` to add an env var to the template. When I then reload from disk, the same groovy script above _did_ show the Env vars NodeProperty on the template's nodeProperties list. But then when I go back to Configure Clouds, there's still nothing where the <descriptorList> is supposed to be.  If I click save at that point, it simply strips the nodeProperties from the template.

I don't think it should matter, but I am using a spot instance template.

I'm at the extent of my knowledge here, but it seems to be something to do with how stapler is generating the form, but I don't know why or how to figure out why.

Jenkins v2.219
Plugin v1.49.1
Amazon EC2 cloud
Spot instance slave template
Nothing that seemed out of the ordinary in jenkins.log
Anything else I can look for? 🤷‍♂️

jhansche:
@mildsunrise Any other suggestions that we can look for, to figure out why the Node Properties isn't appearing on my cloud configuration?  I added a screenshot to the JENKINS-36544 ticket to show what I see:
Screen Shot 2020-03-09 at 11 29 38 AM

mildsunrise:
@jhansche Nope, sorry :( I don't know how to debug those problems... the only thing I could suggest is to build this plugin from source and install it. You could make changes to the template and verify that they appear in the UI.

jhansche:
Thanks for the suggestion.

I ran the plugin locally, with the existing POM configuration, which uses jenkins version 2.150.1 by default. On that version of Jenkins, the "Configure Clouds" is still in the System Configuration ($JENKINS/configure) page, instead of the newer separate page. On that old page, I **do** see the Node Properties descriptor list.

Updating the jenkins version to 2.223 (and fixing a dependency version error with `slf4jVersion`), it is now moved into the newer $JENKINS/configureClouds URL, and on this screen, I no longer see the Node Properties descriptor. It looks like maybe there was a change in Jenkins that caused this to stop working?  

@mildsunrise What version of Jenkins are you running, such that you see it working properly?

jhansche:
The Configure Clouds section got moved to its own page as of https://github.com/jenkinsci/jenkins/pull/4339 with commit 77f1242 on Nov 16.

I confirmed that if I run the plugin against version 2.204, it works correctly and I see the Node Properties list.  Running against version 2.205 breaks it, and the Node Properties do not appear in the separate Configure Clouds screen.

Looks like this PR got approved on Nov 2, but didn't merge until a month later on Dec 1, after the Configure Clouds change happened.  Is it possible that this change assumed it was going to be in the /configure Configure System screen, and after that moves to its own screen, it no longer has access to those Node Properties for some reason?

jhansche:
Tracked this down further by debugging the config.jelly file:

```
<f:descriptorList title="${%Node Properties}" field="nodeProperties" descriptors="${it.getNodePropertyDescriptors()}" />
```
In version 2.204 and older, `$it` is an instance of `hudson.model.Hudson`.  On 2.205 and later, `$it` refers to `jenkins.model.GlobalCloudConfiguration`.

Looking into https://wiki.jenkins.io/display/JENKINS/Basic+guide+to+Jelly+usage+in+Jenkins#BasicguidetoJellyusageinJenkins-Otherpredefinedobjects I see that `${app}` is another way to access the Hudson instance, so changing that to:
```
<f:descriptorList title="${%Node Properties}" field="nodeProperties" descriptors="${app.getNodePropertyDescriptors()}" />
```
causes it to work properly on both older and newer Jenkins versions.

mildsunrise:
Wow, great work :o Yeah, that explains it... Thanks a lot!

jhansche:
New PR opened: https://github.com/jenkinsci/ec2-plugin/pull/440

PR #384, opened 2019-07-30
Ensure external facing streams are closed

res0nance:
- When process is destroyed close all external facing streams
This is partly to defend against any weird channel issues where it still thinks the streams are alive


res0nance:
Not sure if i should close everything just to be defensive

PR #386, opened 2019-08-10
Update pom.xml

thoulen:
Highlight the breaking change, to use the new feature of automatic generated password for  windows , the "ec2:GetPasswordData" has to be added to the IAM role


jvz:
Might want to note something about shared cloud config also being forward-incompatible (i.e., old ec2 plugin won't be able to parse new cloud config info).

PR #387, opened 2019-08-10
[JENKINS-58578] Force the setting of the private ip also for public subnet

thoulen:
JENKINS-58578 Force the setting of the private ip also for public subnet

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -947,11 +947,12 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
             launchSpecification.setKeyName(keyPair.getKeyName());
             launchSpecification.setInstanceType(type.toString());
 
+            net.setAssociatePublicIpAddress(false);
             if (getAssociatePublicIp()) {
                 net.setAssociatePublicIpAddress(true);
-                net.setDeviceIndex(0);
-                launchSpecification.withNetworkInterfaces(net);
             }
+            net.setDeviceIndex(0);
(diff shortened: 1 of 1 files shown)


andgoldin:
This should probably also be updated in the `provisionSpot` function as well: [Lines 950-954](https://github.com/jenkinsci/ec2-plugin/blob/f327c01986aa99033e4e514ceb01e656c485eb3c/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L950-L954)

jvz:
Is this unit testable?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -633,11 +633,12 @@ private void logProvisionInfo(String message) {
             }
         }
 
+        net.setAssociatePublicIpAddress(false);
```
res0nance:
```suggestion
        net.setAssociatePublicIpAddress(getAssociatePublicIp());
```
Will allow you to remove the if

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -946,11 +947,12 @@ private void setupCustomDeviceMapping(List<BlockDeviceMapping> deviceMappings) {
             launchSpecification.setKeyName(keyPair.getKeyName());
             launchSpecification.setInstanceType(type.toString());
 
+            net.setAssociatePublicIpAddress(false);
```
res0nance:
```suggestion
        net.setAssociatePublicIpAddress(getAssociatePublicIp());
```
Same here

varyvol:
Do you think this could be merged?

PR #389, opened 2019-08-20
Use non-deprecated code and remove unused imports

res0nance:
Removing all the deprecated code particularly in the WinRM bits

### DIFF
--- a/src/main/java/hudson/plugins/ec2/win/winrm/WinRMClient.java
+++ b/src/main/java/hudson/plugins/ec2/win/winrm/WinRMClient.java
@@ -22,11 +22,15 @@ import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.ParseException;
 import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.AuthSchemeProvider;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.ClientProtocolException;
 import org.apache.http.client.methods.HttpPost;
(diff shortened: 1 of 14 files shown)


res0nance:
I'm done with this, finally cleaned up most of the deprecated code

src/main/java/hudson/plugins/ec2/win/winrm/WinRMClient.java
```diff
@@ -176,15 +183,40 @@ private void setupHTTPClient() {
...
+                                                            .build();
+            builder.setDefaultAuthSchemeRegistry(authSchemeRegistry);
         }
-        httpclient.setCredentialsProvider(credsProvider);
-        httpclient.getParams().setParameter(CoreConnectionPNames.CONNECTION_TIMEOUT, 5000);
-        // httpclient.setHttpRequestRetryHandler(new WinRMRetryHandler());
+        if (useHTTPS) {
+            try {
+                SSLConnectionSocketFactory sslConnectionFactory = new SSLConnectionSocketFactory(
+                                                                  new SSLContextBuilder().loadTrustMaterial(null, new TrustSelfSignedStrategy()).build(),
```
jvz: Now I'm assuming this is a pure refactoring here, but this is not the greatest idea. There should probably be an admin-controlled config setting to disable hostname verification and whatnot.

res0nance: This is pure refactoring, we should enhance this from a security perspective outside of this PR.

res0nance:
I'll merge this tomorrow if no issues.

PR #398, opened 2019-09-17
[JENKINS-57795] Retry when creating planned nodes

res0nance:
Due to eventual consistency in EC2 API a node started from a stopped state can and will return a stopped state after just being started add a few retries to ensure that nodes do not just fall off due to that.

A second follow up to this might be to search for orphans periodically.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -643,6 +643,10 @@ public abstract class EC2Cloud extends Cloud {
         return new PlannedNode(t.getDisplayName(),
                 Computer.threadPoolForRemoting.submit(new Callable<Node>() {
                     public Node call() throws Exception {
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -681,6 +685,12 @@ public Node call() throws Exception {
                             }
 
                             if (!state.equals(InstanceStateName.Pending)) {
+                                if (attempts++ < 4) {
+                                    LOGGER.log(Level.FINE, "{0}. Node {1} is neither pending, neither running, it's {2}. retrying attempt: {3}",
+                                                new Object[]{t, slave.getNodeName(), state, attempts});
+                                    Thread.sleep(TimeUnit.SECONDS.toMillis(5));
```
jvz: Nit:

```suggestion
                                    TimeUnit.SECONDS.sleep(5);
```

res0nance:
Closing in favour of #399

PR #399, opened 2019-09-19
[JENKINS-57795] Retry when waking up orphans or creating new nodes 

sdhoka:
Root Cause Analysis:
The use of threadpool at times causes some threads to return old values of instance state when making the getState() call. I was able to replicate similar thing using a ruby script (preferred language) that started a stopped instance and used about 10 threads to poll the state method, few threads still returned stopped for a millisecond interval although the instance had already moved to pending. This causes some instances to be started on AWS but are never connected and also, never stopped/terminated
Solution:
1. Add a retry logic on receiving `stopped`  state from the `getstate()` method. The thread will wait for 5 seconds before retrying.  I have been testing this fix in our internal environment and a retry attempt of 1 seems to work perfectly. The retry count can be increased but that can keep the thread waiting for long.
2. Remove the `uptime <  idleMilliseconds` check from EC2RetentionStrategy.java. For planned stopped nodes that are started but not connected,  `idleMilliseconds`  will be counted from the time even before when it was last stopped and will be always greater than `uptime`. This prevents such faulty nodes from being terminated and also incurs AWS costs.  The condition is ideally true for instances in the launching phase that haven't been connected yet. However, this is already taken care of by the condition [here](https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java#L161) which protects termination of launching nodes from getting terminated before 30 minutes (STARTUP_TIMEOUT)

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -643,7 +643,7 @@ public abstract class EC2Cloud extends Cloud {
         return new PlannedNode(t.getDisplayName(),
                 Computer.threadPoolForRemoting.submit(new Callable<Node>() {
                     int retryCount     = 0;
-                    int DESCRIBE_LIMIT = 1;
+                    final int DESCRIBE_LIMIT = 2;
                     public Node call() throws Exception {
                         while (true) {
                             String instanceId = slave.getInstanceId();
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -642,6 +642,8 @@ public abstract class EC2Cloud extends Cloud {
     private PlannedNode createPlannedNode(final SlaveTemplate t, final EC2AbstractSlave slave) {
         return new PlannedNode(t.getDisplayName(),
                 Computer.threadPoolForRemoting.submit(new Callable<Node>() {
+                    int retryCount     = 0;
+                    int DESCRIBE_LIMIT = 1;
                     public Node call() throws Exception {
                         while (true) {
                             String instanceId = slave.getInstanceId();
@@ -681,9 +683,16 @@ public abstract class EC2Cloud extends Cloud {
                             }
 
                             if (!state.equals(InstanceStateName.Pending)) {
-                                LOGGER.log(Level.WARNING, "{0}. Node {1} is neither pending, neither running, it's {2}. Terminate provisioning",
-                                        new Object[]{t, state, slave.getNodeName()});
-                                return null;
+                                
+                                if (retryCount >= DESCRIBE_LIMIT){
+                                    LOGGER.log(Level.WARNING,"Instance {0} did not move to running after {1} attempts, terminating provisioning",
+                                        new Object[]{instanceId, retryCount});
+                                    return null;
+                                }
+
+                                LOGGER.log(Level.INFO, "Attempt {0}: {1}. Node {2} is neither pending, neither running, it''s {3}. Will try again after 5s",
+                                        new Object[]{retryCount, t, slave.getNodeName(), state});
+                                retryCount++;
                             }
 
                             Thread.sleep(5000);
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -643,7 +643,7 @@ public abstract class EC2Cloud extends Cloud {
         return new PlannedNode(t.getDisplayName(),
                 Computer.threadPoolForRemoting.submit(new Callable<Node>() {
                     int retryCount     = 0;
-                    final int DESCRIBE_LIMIT = 2;
+                    private static final int DESCRIBE_LIMIT = 2;
                     public Node call() throws Exception {
                         while (true) {
                             String instanceId = slave.getInstanceId();
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -164,11 +164,6 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
 
             final long idleMilliseconds = this.clock.millis() - computer.getIdleStartMilliseconds();
 
-            //Stopped instance restarted and Idletime has not be reset
-            if ( uptime <  idleMilliseconds) {
-                return 1;
-            }
-
 
             if (idleTerminationMinutes > 0) {
                 // TODO: really think about the right strategy here, see


sdhoka:
@res0nance 
I'd appreciate if you can review this PR.
Thanks,
Shubham

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -164,11 +164,6 @@ private long internalCheck(EC2Computer computer) {
 
             final long idleMilliseconds = this.clock.millis() - computer.getIdleStartMilliseconds();
 
-            //Stopped instance restarted and Idletime has not be reset
-            if ( uptime <  idleMilliseconds) {
```
res0nance: This was added to prevent computers with stale uptimes from terminating.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -642,6 +642,8 @@ else if (jenkinsInstance.isTerminating()) {
     private PlannedNode createPlannedNode(final SlaveTemplate t, final EC2AbstractSlave slave) {
         return new PlannedNode(t.getDisplayName(),
                 Computer.threadPoolForRemoting.submit(new Callable<Node>() {
+                    int retryCount     = 0;
+                    int DESCRIBE_LIMIT = 1;
```
res0nance: There probably should be more than 1 retry and it should be `final`

sdhoka: Thanks for catching that, just fixed it and increased the count to 2

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
sdhoka: Yes but [this](https://github.com/jenkinsci/ec2-plugin/pull/359) PR should have fixed the stale uptimes, right?

res0nance: Much like this PR that number could possibly be stale due to eventual consistency

sdhoka: I am not sure I understand, I am under the impression that #399 should permanently fix the stale uptime issue in `EC2RetentionStrategy.java` since `getstate()` will get the latest info from AWS and refresh uptime value. Could you please elaborate on the concern?

res0nance: I do not know how the EC2 API behaves in this case. Previously we had issues of stale uptimes causing nodes to terminate but those nodes were just spun but did throw the exception seen in this PR

res0nance: @datallah IIRC you added this bit, is it due to this?

sdhoka: @res0nance 
Hello Rihaan,
I can remove this block for now if you'd want to merge the other change and maybe discuss this separately. My only concern is that instances which are manually started on AWS will never be stopped nor connected to Jenkins by the ec2 plugin in that scenario. Since, for manually started nodes, `uptime` will always be less than `idleMilliseconds`

res0nance: getIdleMilliseconds returns when the node was started i.e when it was connected to jenkins. That number should never be further into the future than the uptime.

https://github.com/jenkinsci/jenkins/blob/6d6a0a665b4ac0864646d16b114bb78b5f21f9c1/core/src/main/java/hudson/model/Computer.java#L1058
https://github.com/jenkinsci/jenkins/blob/6d6a0a665b4ac0864646d16b114bb78b5f21f9c1/core/src/main/java/hudson/model/Executor.java#L887

datallah: @res0nance actually, I didn't add that bit, that has been like that for a while.

I agree with the sentiment that the check should no longer be necessary because of the fixes in #359.

sdhoka: I tried out an experiment on this scenario and was able to confirm that `uptime < idleMilliseconds` can be reproduced as follows
1. Consider a node that was stopped by ec2-plugin on idle timeout after finishing the build
2. For this node , `uptime >  idleMilliseconds` since uptime is counted from the time when the instance was last started but idle time is calculated after the connection process and build completes.
3. Now, manually start this stopped instance. Since ec2-plugin is not aware of this launch, ec2-plugin will never connect to the node and it will not come online. Idle time here will represent the time since the node got idle before it was stopped by the plugin while uptime will be the time since you just manually launched this machine on AWS. Thus, `uptime < idleMilliseconds`. The presence of this condition in `EC2RetentionStrategy.java` basically prevents the ec2-plugin from ever shutting down that manually started instance unless you manually initiate the connection procedure which in turn resets the idle time.
4. A similar issue occurs in `JENKINS-57795` where ec2-plugin starts the instance but never connects to it, this causes idlemilliseconds to never reset
5. Use the following script in the Jenkins script console to try this out
```groovy
package hudson.plugins.ec2;
import java.util.concurrent.TimeUnit;

import org.apache.commons.lang.math.NumberUtils;

Jenkins jenkins = Jenkins.getInstance()
int STARTUP_TIME_DEFAULT_VALUE = 30;
int STARTUP_TIMEOUT = NumberUtils.toInt(System.getProperty(EC2RetentionStrategy.class.getCanonicalName() + ".startupTimeout",String.valueOf(STARTUP_TIME_DEFAULT_VALUE)), STARTUP_TIME_DEFAULT_VALUE)

EC2Computer computer = jenkins.getComputer("Your node name in jenkins")

EC2RetentionStrategy ret = new EC2RetentionStrategy('-5')

idleTerminationMinutes = ret.idleTerminationMinutes

println computer.isIdle()
println computer.getState();
println computer.isOffline()

long uptime = computer.getUptime()
long idleMilliseconds = System.currentTimeMillis() - computer.getIdleStartMilliseconds()

println uptime < idleMilliseconds
```

res0nance: In light of this new discovery would you re-add that check?

sdhoka: No, I meant to explain that removing that check is necessary otherwise it will prevent some slaves from getting stopped even after idle timeout

res0nance: Sorry, I misunderstood

mirzmaster:
Thank you @sdhoka & @res0nance for your efforts towards resolving this issue, which I believe we've been running into as well in our environment.  When can we expect to see this in a release?

sdhoka:
@res0nance 
Thanks for getting this merged. Do you think that this fix might also be applicable for [ JENKINS-56036](https://issues.jenkins-ci.org/browse/JENKINS-56036)? Not sure if spot instances use the same logic

res0nance:
It should be that as well

res0nance:
@mirzmaster I'm hoping to release this week

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
thoulen: that was introduced because the startup time can be longer than idle

chkelly:
Appreciate all the work put into this PR as we believe this will solve issues properly launching build slaves with our Jenkins installs. 

Is there any sense of when the next release will be cut?

PR #401, opened 2019-09-20
 [JENKINS-25106] Support minimumNoInstances

erikhakansson:
I'm attempting to open an entirely new PR instead of #343 to see if my weird issues goes away.


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -126,6 +128,8 @@
 
     public int instanceCap;
 
+    public int minimumNumberOfInstances;
```
jvz: No need to add more public fields here. That's not a great practice in Java.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1289,6 +1356,16 @@ public boolean isUseHTTPS() {
         return amiType.isWindows() && ((WindowsData) amiType).isUseHTTPS();
     }
 
+    @Extension
+    public static final class OnSaveListener extends SaveableListener {
+        @Override
+        public void onChange(Saveable o, XmlFile file) {
+            if (o instanceof Hudson) {
```
```suggestion
            if (o instanceof Jenkins) {
```

src/main/java/hudson/plugins/ec2/util/EC2SlaveConfig.java
```diff
@@ -0,0 +1,285 @@
...
+
+        final String spotInstanceRequestId;
+
+        private Spot(SpotBuilder builder) {
+            super(builder);
+            this.spotInstanceRequestId = builder.spotInstanceRequestId;
+        }
+    }
+
+    private static abstract class Builder<B extends Builder, C extends EC2SlaveConfig> {
```
jvz:
```suggestion
    private static abstract class Builder<B extends Builder<B, C>, C extends EC2SlaveConfig> {
```

I think this would be a more appropriate generic bound here provided it still compiled. :)

src/main/java/hudson/plugins/ec2/util/EC2SlaveConfig.java
```diff
@@ -0,0 +1,285 @@
...
+
+import hudson.model.Node;
+import hudson.plugins.ec2.AMITypeData;
+import hudson.plugins.ec2.ConnectionStrategy;
+import hudson.plugins.ec2.EC2Tag;
+import hudson.slaves.NodeProperty;
+
+import java.util.List;
+
+public abstract class EC2SlaveConfig {
```
jvz: I'm just think out loud here, but this config with builder class pattern being developed here seems like a nice candidate for improving Jenkins' databinding feature to natively support this type of pattern.

jvz:
Do you think the new slave-related classes can use the updated term "agent"? We could refactor the old slave classes to use agent in a separate PR, but it'd be easier if we didn't introduce new code with the old name since it's a pain to change when people are already depending on the API.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
erikhakansson: Indeed. I probably just followed the existing pattern there.

erikhakansson: Commited a fix

src/main/java/hudson/plugins/ec2/util/EC2SlaveConfig.java
erikhakansson: I'm generally of fan of builder patterns as it shows the user of the class a very verbose representation of what's used in constructing the class. `withSomeProperty()` is a lot more easier to understand at a quick glance than Constructor parameter #5 :)

That being said, I don't have that much experience with Jenkins' databinding, or Jenkins at all (other than using it) so I don't feel qualified to judge if this pattern works everywhere else :)

erikhakansson:
> Do you think the new slave-related classes can use the updated term "agent"? We could refactor the old slave classes to use agent in a separate PR, but it'd be easier if we didn't introduce new code with the old name since it's a pain to change when people are already depending on the API.

Sure! Just `Slave` to `Agent` like that? Is that a Jenkins thing, that it's been changed to Agent?

jvz:
At some point, Jenkins changed the name from "slave" to "agent" officially, though fully migrating everything to the new name would break various existing APIs, so it's mostly been done through the UI and documentation. New code has been encouraged to use the new name.

Someone went through the trouble of renaming `Hudson` to `Jenkins` in a backwards compatible way, so I'm not sure why the main `Slave` classes haven't also been similarly renamed yet. The jar file is `agent.jar` nowadays, though Remoting supports a few filename variants.

src/main/java/hudson/plugins/ec2/util/EC2SlaveConfig.java
jvz: Yeah, I don't, either, though I do have extensive experience in the equivalent functionality in Log4j2, so it's an active interest of mine in general. :)

The gist of my idea at least would be using `@DataBoundSetter` on the `with` or `set` methods and a `build()` method (or a new annotation), though I'm not sure how that would fit in with Stapler and Jenkins APIs. Hopefully I'll remember to revisit this after figuring that out.

src/main/java/hudson/plugins/ec2/util/MinimumInstanceChecker.java
```diff
@@ -0,0 +1,43 @@
...
+import hudson.plugins.ec2.EC2Cloud;
+import hudson.plugins.ec2.EC2Computer;
+import hudson.plugins.ec2.SlaveTemplate;
+import jenkins.model.Jenkins;
+
+import javax.annotation.Nonnull;
+import java.util.Arrays;
+import java.util.Objects;
+
+public class MinimumInstanceChecker {
```
jvz: Unless this class is meant for public use, you can add `@Restricted(NoExternalUse.class)` to this (and anything else that's not meant for use outside of the plugin; it helps avoid restricting backward compatibility).

src/main/java/hudson/plugins/ec2/util/EC2SlaveConfig.java
erikhakansson: Haha yeah there's always the issue of remembering stuff.

src/main/java/hudson/plugins/ec2/util/EC2AgentFactory.java
```diff
@@ -0,0 +1,27 @@
...
+            if (instance != null) {
+                throw new IllegalStateException("Multiple implementations of " + EC2AgentFactory.class.getName()
+                        + " found. If overriding, please consider using ExtensionFilter");
+            }
+            instance = implementation;
+        }
+        return instance;
+    }
+
+    EC2OndemandSlave createOnDemandSlave(EC2AgentConfig.OnDemand config) throws Descriptor.FormException, IOException;
```
jvz: You can also rename the newly added methods.

src/test/java/hudson/plugins/ec2/util/AmazonEC2FactoryMockImpl.java
```diff
@@ -40,21 +50,98 @@ public static AmazonEC2 createAmazonEC2Mock() {
...
-    public static AmazonEC2 createAmazonEC2Mock(@Nullable Answer defaultAnswer) {
-        AmazonEC2 mock;
+    public static AmazonEC2Client createAmazonEC2Mock(@Nullable Answer defaultAnswer) {
+        AmazonEC2Client mock;
         if (defaultAnswer != null) {
-            mock = Mockito.mock(AmazonEC2.class, defaultAnswer);
+            mock = Mockito.mock(AmazonEC2Client.class, defaultAnswer);
         } else {
-            mock = Mockito.mock(AmazonEC2.class);
+            mock = Mockito.mock(AmazonEC2Client.class);
```
jvz: You can use static imports for `Mockito` methods.

src/main/java/hudson/plugins/ec2/util/MinimumInstanceChecker.java
erikhakansson: I added that. But I didn't do it for the factories, etc, since I figure someone might actually want to extend those.

src/test/java/hudson/plugins/ec2/util/AmazonEC2FactoryMockImpl.java
erikhakansson: Added static import now, but I actually prefer not to have them. With `Mockito.mock` it makes it obvious that it's Mockito and not e.g. PowerMock.

Similarly to `assertEquals`; are we talking AssertJ, Junit, something else? I just prefer the explicitness :)

erikhakansson:
> LGTM! Thanks for the great work!

Thank you! :)

erikhakansson:
Thanks :) Let's see if CI ever lets it through though :)

PR #402, opened 2019-09-27
[JENKINS-59562] Always use network interface for subnet

res0nance:
If you specify network interfaces you must not specify the subnet or security groups on the instance level.

Essentially what happens is when you specify a network interface all networking related configurations must be on the interface specification or the ec2 api will return an error.


res0nance:
I can test this manually on monday, I'm hoping the issue reporter would test this as well as I have updated the ticket.

res0nance:
Apparently security groups must be specified the same way as well. Changed that bit too

res0nance:
CI failures are due to a windows deletion issue

Kern3l:
Hey, will 1.46.1 be released in the plugin repository so we can have it working?

varyvol:
@Kern3l it should be available already.

PR #403, opened 2019-09-27
Allow 0 in form validation for minimum number of instances

erikhakansson:
I realized that 0 was not a valid number for minimum number of instances after I tried out the new release.

Only a form validation error, so it's still possible to continue despite having 0 as the value, but this fixes that issue.


varyvol:
Belated approval.

PR #404, opened 2019-09-29
Support limiting minimum instances to a time range

erikhakansson:
Follow-up to PR #401, this PR adds the option to limit minimum number of instances to a specific time range.

I got some wishes for that from work so that we don't have build agents up and running all weekend when no one is working.

image

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -130,6 +130,8 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
     private int minimumNumberOfInstances;
 
+    private MinimumNumberOfInstancesTimeRangeConfig minimumNumberOfInstancesTimeRangeConfig;
+
     public final boolean stopOnTerminate;
 
     private final List<EC2Tag> tags;
@@ -506,6 +508,15 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
     public int getMinimumNumberOfInstances() {
         return minimumNumberOfInstances;
+    }
+
+    public MinimumNumberOfInstancesTimeRangeConfig getMinimumNumberOfInstancesTimeRangeConfig() {
+        return minimumNumberOfInstancesTimeRangeConfig;
+    }
+
+    @DataBoundSetter
+    public void setMinimumNumberOfInstancesTimeRangeConfig(MinimumNumberOfInstancesTimeRangeConfig minimumNumberOfInstancesTimeRangeConfig) {
+        this.minimumNumberOfInstancesTimeRangeConfig = minimumNumberOfInstancesTimeRangeConfig;
     }
 
     public int getInstanceCap() {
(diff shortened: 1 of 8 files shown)


src/main/java/hudson/plugins/ec2/util/MinimumInstanceChecker.java
```diff
@@ -8,12 +9,18 @@
...
+import java.time.LocalDateTime;
+import java.time.LocalTime;
 import java.util.Arrays;
 import java.util.Objects;
 
 @Restricted(NoExternalUse.class)
 public class MinimumInstanceChecker {
 
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Needs to be overridden from tests")
+    public static Clock clock = Clock.systemDefaultZone();
```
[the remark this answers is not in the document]
erikhakansson: That would be a prettier way of setting the clock, but remember the discussions we had in #379 ?

I had huge issues in managing to override anything Guice-related in the test-context, mostly due to the fact that the Jenkins test harness manages the lifecycle on its own and doesn't really allow me to interfer from tests.

I might be wrong, but I think some change are necessary on the Jenkins core side to respect mock injects and stuff like that when running in the test harness.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -130,6 +130,8 @@
 
     private int minimumNumberOfInstances;
 
+    private MinimumNumberOfInstancesTimeRangeConfig minimumNumberOfInstancesTimeRangeConfig;
```
varyvol: What about old configs not having this property? Should you manage that in readResolve?

res0nance: IIUC There is a nullcheck for its usage

varyvol: Right!

src/main/java/hudson/plugins/ec2/util/MinimumInstanceChecker.java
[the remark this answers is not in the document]
res0nance: @jvz Are you ok with merging this in its current state?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
erikhakansson: Correct, there's a null check. And I did a quick check now and couldn't find anywhere that I had missed checking. :)

src/main/java/hudson/plugins/ec2/util/MinimumInstanceChecker.java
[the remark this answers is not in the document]
jvz: No objections to this area. I don't remember if I did a full code review of this PR, so you might want a second approval first before merging.

jvz:
I won't have time to review this for a couple days.

res0nance:
@jvz do you have the bandwidth to review this?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1503,6 +1514,38 @@ public FormValidation doCheckMinimumNumberOfInstances(@QueryParameter String val
...
+            try {
+                MinimumNumberOfInstancesTimeRangeConfig.validateLocalTimeString(value);
+                return FormValidation.ok();
+            } catch (IllegalArgumentException e) {
+                return FormValidation.error("Please enter value in format 'h:mm a' or 'HH:mm'");
+            }
+        }
+
+        // For some reason, all days will validate against this method so no need to repeat for each day.
+        public FormValidation doCheckMonday(@QueryParameter boolean monday,
```
varyvol: Maybe a different name if it's valid for all days?

erikhakansson: Ideally, yes, but it's some sort of Jelly magic based on the field names.

src/main/java/hudson/plugins/ec2/util/MinimumNumberOfInstancesTimeRangeConfig.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+public class MinimumNumberOfInstancesTimeRangeConfig {
+
```
ro-kr: The field names are excessively long due to the class name prefix. Is this prefix required?

src/main/resources/hudson/plugins/ec2/SlaveTemplate/config.jelly
```diff
@@ -142,6 +142,34 @@ THE SOFTWARE.
...
+              <f:checkbox field="friday" checked="${instance.minimumNumberOfInstancesTimeRangeConfig.minimumNoInstancesActiveTimeRangeDays.friday}"/>
+              <label>${%Friday}</label>
+              <f:checkbox field="saturday" checked="${instance.minimumNumberOfInstancesTimeRangeConfig.minimumNoInstancesActiveTimeRangeDays.saturday}"/>
+              <label>${%Saturday}</label>
+              <f:checkbox field="sunday" checked="${instance.minimumNumberOfInstancesTimeRangeConfig.minimumNoInstancesActiveTimeRangeDays.sunday}"/>
+              <label>${%Sunday}</label>
+          </f:entry>
+      </f:rowSet>
+    </f:optionalBlock>
+    <f:entry />
```
ro-kr: Is this empty entry element required?

erikhakansson: Not required, but it gets really cramped and ugly without it.

res0nance:
Merging tomorrow if no one has any objections

PR #410, opened 2019-10-10
Allow custom definition of tool locations and node properties of AWS EC2 slaves

GeorgeCimpoies:
- manually tested the changes in both our UAT & PROD environment, plugin loads & works as expected. 
- fixed failing unit tests.

Also, this solves https://issues.jenkins-ci.org/browse/JENKINS-51581

Would be great if we could have this in the next release

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -32,6 +32,7 @@ import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -41,6 +42,9 @@ import java.util.stream.Stream;
 import javax.servlet.ServletException;
 
 import hudson.plugins.ec2.util.*;
+import hudson.slaves.NodeProperty;
+import hudson.slaves.NodePropertyDescriptor;
+import hudson.util.DescribableList;
 
 import hudson.XmlFile;
 import hudson.model.listeners.SaveableListener;
@@ -154,6 +158,8 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
     public String currentSubnetId;
 
+    private /* lazily initialized */ DescribableList<NodeProperty<?>, NodePropertyDescriptor> nodeProperties;
+
     private transient/* almost final */Set<LabelAtom> labelSet;
 
     private transient/* almost final */Set<String> securityGroupSet;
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -187,7 +193,8 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
             String instanceCapStr, String iamInstanceProfile, boolean deleteRootOnTermination,
             boolean useEphemeralDevices, boolean useDedicatedTenancy, String launchTimeoutStr, boolean associatePublicIp,
             String customDeviceMapping, boolean connectBySSHProcess, boolean monitoring,
-            boolean t2Unlimited, ConnectionStrategy connectionStrategy, int maxTotalUses) {
+            boolean t2Unlimited, ConnectionStrategy connectionStrategy, int maxTotalUses,
```
jvz: This breaks backward compatibility. You can either add this as a setter method with `@DataBoundSetter`, or you'll have to deprecate the old constructor and add this new one just like the dozens below.

GeorgeCimpoies: Done

GeorgeCimpoies:
Thanks for the quick review! Hope it's all good :)

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -559,6 +583,14 @@ public int getMaxTotalUses() {
         return maxTotalUses;
     }
 
+    public List<NodePropertyDescriptor> getNodePropertyDescriptors() {
+        return NodePropertyDescriptor.for_(NodeProperty.all(), EC2AbstractSlave.class);
+    }
+
+    public DescribableList<NodeProperty<?>, NodePropertyDescriptor> getNodeProperties() {
+        return Objects.requireNonNull(nodeProperties);
```
jvz: Nit: it might make more sense to throw an `IllegalStateException` here instead of `NullPointerException`.

GeorgeCimpoies: Done! WDYT?

GeorgeCimpoies:
@jvz I'm not sure why the CI fails, could you please point me in the right direction?

jvz:
The test errors are saying something about breaking backward compatibility.

GeorgeCimpoies:
I suppose it won't be merged without 100% passing tests?

jvz:
Valid test errors, yes. If the test failures are incidental due to an infrastructural issue (happens from time to time), we can still merge if we can verify the tests on other machines.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -223,6 +230,7 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
         this.useDedicatedTenancy = useDedicatedTenancy;
         this.connectBySSHProcess = connectBySSHProcess;
         this.maxTotalUses = maxTotalUses;
+        this.nodeProperties = new DescribableList<NodeProperty<?>,NodePropertyDescriptor>(Saveable.NOOP, nodeProperties);
```
ro-kr: - the type parameters on the constructor are redundant

- the `nodeProperties` argument may be `null` (e.g. in unit tests), this causes a NullPointerException in the constructor of the DescribableList. Wrapping the argument with `Util.fixNull(...)` should prevent this.

GeorgeCimpoies: Done

GeorgeCimpoies:
@jvz @ro-kr OK to merge?

GeorgeCimpoies:
https://github.com/jenkinsci/ec2-plugin/commit/6bf5a84f85e5caa2febb9ed765f977ff3c9faf72 was merged recently so I suppose we can close this.

PR #412, opened 2019-10-29
Bump dependency

awiddersheim:
While looking at upgrading too `2.190.2` from `2.176.4` I saw the Trilead breakout and noticed it's already been accounted for. However, I did see this. Not sure if it's time yet for this bump or not but here you go anyway!


awiddersheim:
Also, see https://github.com/jenkinsci/ec2-plugin/commit/619f0d1806166ee25ce564def9d2eb3134702c4a for more details.

res0nance:
You can't do this unless the minimum baseline is `2.189` IIRC

PR #413, opened 2019-10-30
Add Per-AMI Spare Instance parameter

hughsaunders:
Currently there is a parameter for minimum number of instances for an AMI,
if you set it to n, there will always be at least n instances.

This PR adds minimum spare instances, if you set this to n, there will
always be n unsused instances waiting for new builds (unless the
instance cap is reached). This prevents new builds from having to wait
for provisioning, while also meaning that you don't need to set a high
absolute minimum.

Related: conjurinc/ops#401

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -36,6 +36,7 @@ import hudson.slaves.RetentionStrategy;
 import jenkins.model.Jenkins;
 
 import java.time.Clock;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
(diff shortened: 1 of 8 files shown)


hughsaunders:
Hi @res0nance, thanks for your review, I've fixed the ci issues, mind having another look?
Thanks!

erikhakansson:
LGTM! Great idea that I'll definitely use at work. 

@res0nance is that because I'm not a member of the organisation?

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -169,8 +166,22 @@ private long internalCheck(EC2Computer computer) {
...
 
             //on rare occasions, AWS may return fault instance which shows running in AWS console but can not be connected.
-            //need terminate such fault instance by {@link #STARTUP_TIMEOUT}
-            if (computer.isOffline() && uptime < TimeUnit.MINUTES.toMillis(STARTUP_TIMEOUT)) {
+            //need terminate such fault instance.
+            // An instance may also fail running user data scripts and
+            // need to be cleaned up.
+            if (computer.isOffline()){
+                EC2AbstractSlave node = computer.getNode();
+                if (Objects.isNull(node)){
```
jvz:
```suggestion
                if (node == null){
```

The `Objects::isNull` method is mostly useful as a function reference.

hughsaunders:
> Code changes look alright. Could you add a test for this feature?

Will Do

hughsaunders:
Added basic test

Jetersen:
Please for the love of JCasC and the next person touching this, split the databoundconstructor to only be MANDATORY fields and everything else should use databound setters.

https://issues.jenkins-ci.org/browse/JENKINS-57513
IMHO this should have never been merged.

PR #414, opened 2019-10-31
Handle DescribeInstances failures when creating new instance

dbnicholson:
Sometimes when launching a new instance, `DescribeInstances` does not report data within the amount of time that `getInstanceWithRetry` waits. However, the instance is created and eventually Jenkins can't create any more nodes because we're at maximum capacity.

This PR attempts 2 fixes:

1. Increase the timeout waiting for `DescribeInstances` to return data when creating a new instance

2. If no data could be found, send a termination request for the request so it doesn't pollute the capacity accounting.


dbnicholson:
BTW, I haven't actually run or built this code. I don't have a setup to build Jenkins plugins.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -594,6 +606,10 @@ private void fetchLiveInstanceData(boolean force) throws AmazonClientException {
         }
     }
 
+    private void fetchLiveInstanceData(boolean force) throws AmazonClientException {
+        return fetchLiveInstanceData(force, 25);
```
wjt:
```suggestion
        fetchLiveInstanceData(force, 25);
```

jvz:
Looks like you have a compiler error. You can build the plugin by running `mvn install`.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
dbnicholson: Indeed, thanks.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -162,7 +162,19 @@ public EC2AbstractSlave(String name, String instanceId, String templateDescripti
...
+        try {
+            // Wait up to 1 minute for the instance to show up
+            fetchLiveInstanceData(true, 60);
+        } catch (com.amazonaws.AmazonClientException e) {
+            /*
+             * If DescribeInstances didn't return any information about this
+             * instance, try to terminate it so that if it does come up later
+             * it doesn't affect capacity calculations.
+             */
+            LOGGER.log(Level.WARNING, "Failed to get instance data for new instance " + getInstanceId(), ", terminating");
```
res0nance:
```suggestion
            LOGGER.log(Level.WARNING, "Failed to get instance data for new instance " + getInstanceId() + ", terminating");
```
I'm assuming its a string

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -162,7 +162,19 @@ public EC2AbstractSlave(String name, String instanceId, String templateDescripti
         this.amiType = amiType;
         this.maxTotalUses = maxTotalUses;
         readResolve();
-        fetchLiveInstanceData(true);
+        try {
+            // Wait up to 1 minute for the instance to show up
+            fetchLiveInstanceData(true, 60);
```
res0nance: When does this get called? If its during a provision operation, having such a long timeout might not be ideal as during provisioning the queue lock is held which means during this 1 minute builds you start won't be queued up and builds can't be picked up by nodes.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
dbnicholson: Yes, thanks.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
dbnicholson: I'm not totally sure what Jenkins context this is called in, but I'm pretty sure it doesn't block other builds from queuing. That said, it already blocks for up to 25 seconds. This just extends it to 60 seconds.

We could try leaving out this commit and just having the commit that tries to terminate the instance if it couldn't get the information about the instance. That would hopefully keep it from getting deadlocked.

Yet another possible fix is for the EC2 plugin to keep track of which instances it launched and actively terminate any other instances it finds in EC2 with the same details. I didn't look into that, though.

res0nance: Could we keep the timeout to be the same and explore increasing it if it does not solve the issue completely

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -162,7 +162,19 @@ public EC2AbstractSlave(String name, String instanceId, String templateDescripti
...
+        } catch (com.amazonaws.AmazonClientException e) {
+            /*
+             * If DescribeInstances didn't return any information about this
+             * instance, try to terminate it so that if it does come up later
+             * it doesn't affect capacity calculations.
+             */
+            LOGGER.log(Level.WARNING, "Failed to get instance data for new instance " + getInstanceId() + ", terminating");
+            terminateInstance();
```
res0nance: @thoulen also made a point that this terminateInstance call can fail if the api gets throttled

PR #416, opened 2019-11-22
Update EC2Slave and SlaveTemplate help text

rdpa:
~Remove obsolete help text for help-stopOnTerminate.html. EC2 is now billed per second (min 60 seconds), so short stop timeouts are no longer counterproductive.~

Add clarification that short stop timeouts only matter for windows agents.


res0nance:
Windows is still billed by the hour so the text should be amended with that information

PR #429, opened 2020-02-11
Handle cases where SlaveTemplate has been removed

res0nance:
Recently we encoutered an issue where when a SlaveTemplate is removed but an old Instance that uses this slave template still exists it will throw an NPE.

This PR annotates the methods and reworks the code to handle such scenarios more gracefully.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
@@ -126,9 +126,14 @@ public class EC2UnixLauncher extends EC2ComputerLauncher {
         boolean successful = false;
         PrintStream logger = listener.getLogger();
         EC2AbstractSlave node = computer.getNode();
+        SlaveTemplate template = computer.getSlaveTemplate();
 
         if(node == null) {
             throw new IllegalStateException();
+        }
+
+        if (template == null) {
+            throw new IOException("Could not find corresponding slave template for " + computer.getDisplayName());
         }
 
         if (node instanceof EC2Readiness) {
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -411,9 +416,9 @@ public boolean verifyServerHostKey(String hostname, int port, String serverHostK
         }
     }
 
-    private String getEC2HostAddress(EC2Computer computer) throws InterruptedException {
+    private static String getEC2HostAddress(EC2Computer computer, SlaveTemplate template) throws InterruptedException {
```
fcojfernandez: Why static? I cannot see when it's needed

res0nance: I changed it to static because it doesn't need any value in the actual class itself so it should have been static from the start as its not required to be tied to any instance of the Launcher

fcojfernandez: your words make sense, but it looks weird to me having a static private method which is not called within a static block of code. Anyway, I'm not blocking the PR because of this. It's just matter of taste and I only wanted to know if there were other reasons. I will let the others give their opinion if applies and I'm approving the PR.

res0nance: Ahh for me I used to mostly write C++, so a static method has the benefit of not passing in an additional parameter(The this pointer). So the habit stayed with me.

res0nance:
> LGTM. The use of `IOException` could potentially be replaced with something more specific, though I can't think of anything besides maybe `IllegalStateException`.

Can't use IllegalState as its not caught outside and does not terminate the instance. It might also lead to a FD leak scenario(which seems to be a bug in core)

res0nance:
If there are no objections to merging it as is, I will be merging tomorrow

PR #430, opened 2020-02-21
Update Spare Instance parameter to consider idle executors

Graham42:
If we're using more than 1 executor per agent and the scheduling
strategy spreads builds across the existing agents, we could end up
scaling up many more agents than we need.


res0nance:
Could you add a test for this?

Graham42:
I don't see any existing tests for this area of the code (unless I've missed something) and I'm not invested enough in this change right now to write a new test suite from scratch. If someone else wants to take on tests for this feel free.

PR #433, opened 2020-02-27
Offload terminates to separate thread

res0nance:
Offload termination of agents so as to not stall `NodeProvisioner.update` when NodeProvisioner.update is called it will hold the queue lock. In order to reduce the hotness of this lock schedule all terminates to be off thread

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -30,6 +30,7 @@ import hudson.model.Descriptor.FormException;
 import hudson.model.Node;
 import hudson.model.Slave;
 import hudson.plugins.ec2.util.AmazonEC2Factory;
+import hudson.plugins.ec2.util.ResettableCountDownLatch;
 import hudson.slaves.NodeProperty;
 import hudson.slaves.ComputerLauncher;
 import hudson.slaves.RetentionStrategy;
@@ -41,13 +42,11 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javax.annotation.Nonnull;
-import javax.annotation.concurrent.GuardedBy;
 
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
@@ -114,9 +113,8 @@ public abstract class EC2AbstractSlave extends Slave {
     protected transient long lastFetchTime;
 
     /** Terminate was scheduled */
-    @GuardedBy("this")
     @Nonnull
-    protected transient volatile CountDownLatch terminateScheduled = new CountDownLatch(0);
+    protected transient ResettableCountDownLatch terminateScheduled = new ResettableCountDownLatch(1, false);
 
     /*
      * The time (in milliseconds) after which we will always re-fetch externally changeable EC2 data when we are asked
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -45,6 +45,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.annotation.concurrent.GuardedBy;
+
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
@@ -108,6 +110,10 @@ public abstract class EC2AbstractSlave extends Slave {
 
     /* The time at which we fetched the last instance data */
     protected transient long lastFetchTime;
+
+    /** Terminate was scheduled */
+    @GuardedBy("this")
+    protected transient volatile boolean terminateScheduled = false;
 
     /*
      * The time (in milliseconds) after which we will always re-fetch externally changeable EC2 data when we are asked
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -465,6 +465,11 @@ public abstract class EC2AbstractSlave extends Slave {
         return null;
     }
 
+    @Override
+    public boolean isAcceptingTasks() {
+        return !terminateScheduled;
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -459,6 +465,11 @@ public Node reconfigure(final StaplerRequest req, JSONObject form) throws FormEx
         return null;
     }
 
+    @Override
+    public boolean isAcceptingTasks() {
```
res0nance: From the javadocs this signals that the computer can't accept tasks so when terminating once out of the lock this node should not be able to receive a task

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
```diff
@@ -471,4 +472,11 @@ public void testRetentionStopsAfterActiveRangeEnds() throws Exception {
         assertEquals(1, computers.size());
         assertEquals(1, AmazonEC2FactoryMockImpl.instances.size());
     }
+
+    private static void checkRetentionStrategy(EC2RetentionStrategy rs, EC2Computer c) throws InterruptedException {
+        rs.check(c);
+        EC2AbstractSlave node = c.getNode();
+        while (node.terminateScheduled)
+            Thread.sleep(TimeUnit.SECONDS.toMillis(1));
```
jvz: At first, I was going to suggest mocking the `Clock` instance in the test to advance time, but then I realized this is a spinlock. To be a more reliable test that depends less on the passage of time (i.e., stateful), use of [CountDownLatch](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountDownLatch.html) which counts down at the end of the `Runnable` submitted would be more appropriate.

res0nance: Hmm, how do you propose that to work. Since the call happens inside terminate() I don't see a way of introducing it in tests.

jvz: If you want to make it repeatable, there's a CyclicBarrier class you can use to reset the counter each time. Basically, you count down on terminate, and in the unit test, you use await() on the latch or barrier.

jvz: As for testability, it'd be ok to expose as package-private for unit tests (I think there might be a `@VisibleForTesting` annotation or something similar you can use).

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -109,6 +113,11 @@
     /* The time at which we fetched the last instance data */
     protected transient long lastFetchTime;
 
+    /** Terminate was scheduled */
+    @GuardedBy("this")
+    @Nonnull
+    protected transient volatile CountDownLatch terminateScheduled = new CountDownLatch(0);
```
jvz: A volatile lock? This seems overly complicated now. Do you think [CyclicBarrier](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CyclicBarrier.html) would work better here? If you need to be able to reset the latch, that would work better with the barrier.

res0nance: CyclicBarrier makes this worse due to `BrokenBarrierExceptions` that can occur.

One thing i can do is move the nasty bits into a separate utility class and make a ResettableCountdownLatch

jvz: That sounds like a better approach. Go for it!

src/main/java/hudson/plugins/ec2/util/ResettableCountDownLatch.java
```diff
@@ -0,0 +1,43 @@
+package hudson.plugins.ec2.util;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class ResettableCountDownLatch {
```
jvz: Suggestion: add `@Restricted(NoExternalUse.class)` as this isn't really meant to be available to other plugins to use.

nigelarmstrong-bc:
This PR seems to have introduced a critical bug @res0nance https://issues.jenkins-ci.org/browse/JENKINS-62062

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -109,6 +112,10 @@
     /* The time at which we fetched the last instance data */
     protected transient long lastFetchTime;
 
+    /** Terminate was scheduled */
+    @Nonnull
+    protected transient ResettableCountDownLatch terminateScheduled = new ResettableCountDownLatch(1, false);
```
nigelarmstrong-bc, line 117: @res0nance My java isn't very good, but I think when these objects are deserialized `ResettableCountDownLatch` is getting set to `null` by default and not a new object.

nigelarmstrong-bc: I think you can fix this in `readResolve`

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -459,6 +466,11 @@ public Node reconfigure(final StaplerRequest req, JSONObject form) throws FormEx
         return null;
     }
 
+    @Override
+    public boolean isAcceptingTasks() {
```
roundtrip-rslobojan, line 470: The addition of this method has resulted in a critical bug (as noted by @nigelarmstrong-bc) - this method does not exist on serialized versions of this class created by previous plugins, and you get this exception in Jenkins on any access:

```
Caught exception evaluating: c.iconClassName in /jenkins/ajaxExecutors. Reason: java.lang.reflect.InvocationTargetException
java.lang.NullPointerException
	at hudson.plugins.ec2.EC2AbstractSlave.isAcceptingTasks(EC2AbstractSlave.java:471)
	at hudson.model.Computer.isAcceptingTasks(Computer.java:1612)
	at hudson.slaves.SlaveComputer.isAcceptingTasks(SlaveComputer.java:177)
	at hudson.model.Computer.getIconClassName(Computer.java:753)
Caused: java.lang.reflect.InvocationTargetException
```

roundtrip-rslobojan: cc @thoulen ^^

PR #435, opened 2020-02-28
Misc cleanups

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Cleanup some of the old code


src/main/java/hudson/plugins/ec2/win/winrm/request/OpenShellRequest.java
```diff
@@ -13,7 +13,7 @@
 
 import com.google.common.collect.ImmutableList;
 
-public class OpenShellRequest extends AbstractWinRMRequest implements WinRMRequest {
+public class OpenShellRequest extends AbstractWinRMRequest {
```
res0nance: AbstractWinRMRequest already implements WinRMRequest as such this can be removed

PR #438, opened 2020-03-09
Fix Javascript error with Node Properties that prevented config display

troymohl:
This fixes the bug reported in https://issues.jenkins-ci.org/browse/JENKINS-60890 that prevented configuration options from displaying.


mildsunrise:
Shouldn't we move the [method](https://github.com/jenkinsci/ec2-plugin/blob/6bf5a84f85e5caa2febb9ed765f977ff3c9faf72/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L585) to the descriptor class for this to work?

troymohl:
@mildsunrise I missed that. I will close this PR as https://github.com/jenkinsci/ec2-plugin/pull/440 fixes it.

PR #440, opened 2020-03-09
[JENKINS-36544] Fix Node Properties on Jenkins 2.205+

jhansche:
This is a follow-up fix for PR #382.

In 2.204, the "Configure Clouds" screen was on the main /configure system
configuration page. As a result, `${it}` referred to the global system
object, aka the Hudson instance.

In 2.205, PR jenkinsci/jenkins#4339 merged, which moves the Configure Clouds
section to a new page: /configureClouds. As a result of that change, `${it}`
no longer refers to the Hudson instance, but to a new
GlobalCloudConfiguration object. That object does not have the
`getNodePropertyDescriptors()` method, so it is not able to populate the
"Node Properties" block.

Switching that to `${app}` gives it a correct reference to the Hudson object,
and that allows it to show the Node Properties block on all Jenkins versions,
both before and after v2.205.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -616,10 +616,6 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
         return maxTotalUses;
     }
 
-    public List<NodePropertyDescriptor> getNodePropertyDescriptors() {
-        return NodePropertyDescriptor.for_(NodeProperty.all(), EC2AbstractSlave.class);
-    }
-
     public DescribableList<NodeProperty<?>, NodePropertyDescriptor> getNodeProperties() {
     	return Objects.requireNonNull(nodeProperties);
     }
@@ -1796,6 +1792,10 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
         public String getDefaultConnectionStrategy() {
             return ConnectionStrategy.PRIVATE_IP.toString();
+        }
+
+        public List<NodePropertyDescriptor> getNodePropertyDescriptors() {
+            return NodePropertyDescriptor.for_(NodeProperty.all(), EC2AbstractSlave.class);
         }
 
         public ListBoxModel doFillConnectionStrategyItems(@QueryParameter String connectionStrategy) {
--- a/src/main/resources/hudson/plugins/ec2/SlaveTemplate/config.jelly
+++ b/src/main/resources/hudson/plugins/ec2/SlaveTemplate/config.jelly
@@ -214,7 +214,7 @@ THE SOFTWARE.
       <f:textbox default="-1"/>
     </f:entry>
 
-    <f:descriptorList title="${%Node Properties}" field="nodeProperties" descriptors="${app.getNodePropertyDescriptors()}" />
+    <f:descriptorList title="${%Node Properties}" field="nodeProperties" descriptors="${descriptor.nodePropertyDescriptors}" />
 
   </f:advanced>
 
--- a/src/main/resources/hudson/plugins/ec2/SlaveTemplate/config.jelly
+++ b/src/main/resources/hudson/plugins/ec2/SlaveTemplate/config.jelly
@@ -214,7 +214,7 @@ THE SOFTWARE.
       <f:textbox default="-1"/>
     </f:entry>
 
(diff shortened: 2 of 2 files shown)


jhansche:
cc @res0nance @ro-kr @mildsunrise per original PR participants.

jhansche:
@fcojfernandez Of course! I tried to explain why it works, but much of the information is kind of spread out across here, the linked PR #382, the Jira ticket [JENKINS-36544](https://issues.jenkins-ci.org/browse/JENKINS-36544), as well as in the commit message. My mistake for not adding more screenshots.

The original PR was where I was trying to debug why it doesn't work for me. That PR conversation starts here: https://github.com/jenkinsci/ec2-plugin/pull/382#issuecomment-584312977. And specifically where I finally tracked it down to the problem in this comment: https://github.com/jenkinsci/ec2-plugin/pull/382#issuecomment-596750879

2.204 and earlier, the Configure Clouds section was on the main system configuration page ($JENKINS/configure), and on that page, `$it` is the global Jenkins/Hudson instance. After 2.205, that was moved to its own page, where `$it` is now an instance of a new class `GlobalCloudConfiguration` instead.  `GlobalCloudConfiguration` does not have the `getNodePropertyDescriptors()` method, which is why it did not work on 2.205 and later.

The reason you don't see the method on the `Jenkins` class is because it's inherited from:  `Hudson` < `Jenkins` < `AbstractCIBase` < `Node`, and the method is [`Node.getNodePropertyDescriptors()`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Node.java#L516).

That means the original PR added the `getNodePropertyDescriptors()` method into _this_ project unnecessarily, because that method was never getting called -- I can remove that as part of this fix as well, if you'd like? It was instead calling `$it.getNodePropertyDescriptors()` where `$it` is the global system configuration object (hudson.model.Hudson), which is the same as `$app.getNodePropertyDescriptors()`, so this change works in both <=2.204, and >=2.205.

In all the cases below, my test methodology is:
* Navigate to the cloud configuration screen (this moves to a new page >=2.205)
* Click "Add a new cloud"
* Click "Amazon EC2"
* Under the "AMIs" section, click "Add"
* At the bottom of the AMI block, expand "Advanced"
* Expect to see "Node Properties" block at the bottom of the block
   This fails on Jenkins >=2.205 without my change.

Using the jenkins.version that's defined in this project's pom.xml file (2.150.1), which is before 2.204 and has the Configure Clouds section on the main system configuration screen, this screenshot shows the Node Properties block working correctly with my change (note: it also worked without my change):

Screen Shot 2020-03-10 at 11 35 00 AM

Updating to version 2.205, where the Configure Clouds section moved to a new page. This fails with current master (screenshot taken from the attachment in Jira, was also shown in the original PR https://github.com/jenkinsci/ec2-plugin/pull/382#issuecomment-596602472):

Screen Shot 2020-03-09 at 11 29 38 AM

but works correctly with my change:

Screen Shot 2020-03-10 at 11 39 28 AM

The latest version I can find is Jenkins 2.225, which also works:

Screen Shot 2020-03-10 at 11 43 14 AM

I hope that helps explain what's going on. Let me know if there's anything else I can help clarify.

mildsunrise:
> That means the original PR added the `getNodePropertyDescriptors()` method into _this_ project unnecessarily, because that method was never getting called -- I can remove that as part of this fix as well, if you'd like?

True! That's an issue, we should call the method I added to `SlaveTemplate`, not the one on `Hudson`... Otherwise we're showing the node properties that apply to Jenkins itself, not the EC2 slave.

jhansche:
@mildsunrise if that was your intent, then it would need to be moved into the Descriptor instead, and called with `descriptors="${descriptor.getNodePropertyDescriptors()}"`.

But it's worth noting that the original change already had this behavior. So I really think it's a different fix, and shouldn't be combined into one. But I'll make the change if we just want it fixed.

troymohl:
@jhansche I submitted this PR with the change that used `descriptor.` https://github.com/jenkinsci/ec2-plugin/pull/438

mildsunrise:
Fully agree, that's a different fix, and yes it should be put in the descriptor (apologies, I was confused about Jenkins when I wrote the original PR). I was going to open a fix, but @troymohl did while I was writing the comment, thanks guys 😊💙

jhansche:
I went ahead and made the change here, because #438 alone is not enough to fix it. I tested the change, and it still works properly with the move to Descriptor.

PR #443, opened 2020-03-14
[JENKINS-59048] Export the spotBlockReservationDurationStr constructor parameter for JCasc

multani:
The Jenkins "Configuration as Code" plugin raises an error while trying
to export the `spotBlockReservationDurationStr` attribute from the
configuration, as this constructor attribute is not exposed in the
`SpotConfiguration` class.

Fix: [JENKINS-59048](https://issues.jenkins.io/browse/JENKINS-59048)


res0nance:
I've requested a review from the jcasc guys because I'm not sure if this is the right fix

src/main/java/hudson/plugins/ec2/SpotConfiguration.java
```diff
@@ -20,6 +20,19 @@
         }
     }
 
+    /**
+     * Export the spotBlockReservationDuration attribute for CasC plugin.
+     *
+     * @return The spotBlockReservationDuration attribute as a string.
+     */
+    public String getSpotBlockReservationDurationStr() {
+        if (spotBlockReservationDuration == Integer.MAX_VALUE) {
```
timja: why is this if statement required?

while this would probably work a couple of suggestions but feel free to ignore if maintainers just want the quick fix...

1. The current constructor isn't a 'correct' `DataBoundConstructor`, it has 'mandatory' fields that aren't used in databinding, i.e. `spotBlockReservationDurationStr` is bound to `spotBlockReservationDuration`

This could be solved by deprecating the existing constructor, fixing the parameters in a new constructor and updating any calling code / views.

2. It would be good to extend the test to verify that this works and it doesn't get broken in the future, here's the existing test: https://github.com/jenkinsci/ec2-plugin/blob/master/src/test/java/hudson/plugins/ec2/ConfigurationAsCodeTest.java

Jetersen: I kinda agree with @timja fix the databound constructor.

Make if follow java beans standards not even sure why it takes in strings when stapler and jelly have full support for numbers.

Jetersen:
Okay after looking into it my recommendation still stands.
Make SpotConfig describable.
Add a jelly config for it.
Change 
https://github.com/jenkinsci/ec2-plugin/blob/ad224d83077eaba6b60aeed818ae6afec70b989c/src/main/resources/hudson/plugins/ec2/SlaveTemplate/config.jelly#L61-L75
to use a 
```jelly
<f:optionalProperty title="${%Spot configuration}" field="spotConfig"/>
```

in the newly created `spotConfig` jelly
you can update it to
```jelly
  <f:optionalBlock name="spotConfig" title="Use Spot Instance" checked="${instance.spotConfig != null}">
    <f:entry title="${%Fallback to on-demand instances}" >
      <f:checkbox field="fallbackToOndemand" checked="${instance.spotConfig.fallbackToOndemand}" />
    </f:entry>
    <f:optionalBlock name="useBidPrice" title="Set bid price" inline="true" checked="${h.defaultToTrue(instance.spotConfig.useBidPrice)}">
      <f:validateButton title="${%Check Current Spot Price}" progress="${%Checking...}" method="currentSpotPrice" with="useInstanceProfileForCredentials,credentialsId,region,type,zone,roleArn,roleSessionName,ami" />
      <f:entry title="${%Spot Max Bid Price}" field="spotMaxBidPrice">
        <f:textbox />
      </f:entry>
    </f:optionalBlock>
    
    <f:entry title="${%Spot Block Reservation Duration}" field="spotBlockReservationDuration">
      <f:number />
    </f:entry>
  </f:optionalBlock>
```

src/main/java/hudson/plugins/ec2/SpotConfiguration.java
Jetersen: btw while your at it make the data bound constructor only have mandatory fields and use data bound setters for optional fields.
See https://issues.jenkins-ci.org/browse/JENKINS-57513

multani:
Thanks @res0nance, @timja and @Jetersen for the review and suggestions!

I'm still new to Jenkins plugin development, so I'll try to dig a bit more. If I sum up:

* I'll deprecate the constructor to correctly use mandatory and optional fields, adding getters/setters for optional fields (I need to dig into that)
* I'll update the Jelly, extract the SpotConfig and change it to use numbers instead of strings :+1: 
* Update the tests as well (I missed that!)

multani:
So, I think I addressed most of the comments:

* The previous constructor has been deprecated in favor of one with only the mandatory parameter.

  I went with only `useBidPrice` as the mandatory argument, as it drives the decision to use "spot instances" or "on-demand instances" if it's set or not...

  (TBH, the whole functionality from this class is a bit weird, as to use AWS EC2 Spot Instances, a bit price has to be set, but the UI allows to select spot instances without a bid price, which requests on-demand instances.
  Also, the bid price should be number, not a string, etc.

  Anyway, I put all these concerns aside as it's not really the purpose of the patch.)

  - `DataBoundDescriptor` should be better implemented now (AFAIU?)

  - I had to remove the `final` from the fields for them to be settable via the `DataBoundSetter`, I could also make them private (let me know, that will increase a bit the patch though.)


* I move the spot configuration in a dedicated Jelly file.

* I upgraded the current test to also test the loading/dumping of the spot
  configuration.


This changes a few things though:

* "Spot Block Reservation Duration" is now always set to 0 by default, which is an acceptable and should have the same behavior as the empty string before.

* Dumping the configuration via JCasC doesn't dump the default parameters: if "Spot Block Reservation Duration" and "Fallback to on-demand instances" have the default values, then they don't appear in the YAML file.

  I'm not sure about this one, as if the default values are changing later on, loading the as-of-today configuration in the then-plugin will produce a different final configuration.


[There's still a test failing in `testSpotConfigWithFallback`](https://github.com/jenkinsci/ec2-plugin/pull/443/commits/64af2a17256248860ee251820e6a2da3321928a6#diff-bbdfa1129e587852b313d8ae4e651d2eR199): `orig.spotConfig.useBidPrice` is `false` while `received.spotConfig.useBidPrice` is `true`... I haven't dig into that one yet.

src/main/java/hudson/plugins/ec2/SpotConfiguration.java
```diff
@@ -20,17 +25,39 @@
...
-            return "";
-        } else {
-            return String.valueOf(spotBlockReservationDuration);
-        }
+    @DataBoundConstructor
+    public SpotConfiguration(boolean useBidPrice) {
+        this.useBidPrice = useBidPrice;
+        this.spotMaxBidPrice = "";
+        this.fallbackToOndemand = false;
+        this.spotBlockReservationDuration = 0;
```
Jetersen: These are default for bool and int so no need to clarify.
```suggestion
```

src/main/resources/hudson/plugins/ec2/SpotConfiguration/config.jelly
```diff
@@ -0,0 +1,38 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form" >
+  <f:entry title="${%Fallback to on-demand instances}" >
+    <f:checkbox field="fallbackToOndemand" checked="${instance.spotConfig.fallbackToOndemand}" />
+  </f:entry>
+
+  <f:optionalBlock name="useBidPrice" title="Set bid price" inline="true" checked="${h.defaultToTrue(instance.spotConfig.useBidPrice)}">
```
Jetersen: Not sure this optional block makes sense... however I do not use the plugin.

Jetersen:
JCasC by design does not print default values when exporting.
If default's are changed it's because the plugin maintainer wanted that. So JCasC should not interfere.

I may be able to look into the test failure tomorrow if needed.

src/test/resources/hudson/plugins/ec2/UnixData.yml
```diff
@@ -12,6 +12,11 @@ jenkins:
             type: "T2Micro"
             remoteFS: "/home/ec2-user"
             mode: "NORMAL"
+            spotConfig:
+                fallbackToOndemand: true
```
timja: Formatting looks off, the amiType block below is using 2 spaces for indent

timja:
RE: making the fields private, that’s a breaking change and not recommended, the usual approach for that is creating a new field and migrating the old data.

Not worth it here imo

src/main/resources/hudson/plugins/ec2/SpotConfiguration/config.jelly
multani: Well, it doesn't really make sense to me, as I explained above:

* Either you selected "I want to request a Spot instance" and you *need* to specify a price (this is mandatory on the AWS' side)
* Or, you don't want to request Spot instances, in which case the spotConfig block is just not set.

I could make that change here, but I think it goes a bit out of scope of the initial issue, and it will require more change in the code to reflect that new "logic" (it's a bit convoluted at the moment IMO).
So, I'd prefer to stay at the same feature level for that PR :)

src/test/resources/hudson/plugins/ec2/UnixData.yml
multani: Oups, I compared the indentation with the "command prefix/suffixes" just below, I missed there was an extra key in between...

src/main/resources/hudson/plugins/ec2/SpotConfiguration/config.jelly
multani: > Not sure this optional block makes sense...

Actually, why do you think it doesn't make sense?
The failing test is actually on that attribute, so may be I misunderstood your comment :)

src/test/resources/hudson/plugins/ec2/UnixData.yml
```diff
@@ -12,6 +12,11 @@ jenkins:
             type: "T2Micro"
             remoteFS: "/home/ec2-user"
             mode: "NORMAL"
+            spotConfig:
```
res0nance: Can we ensure that this was set appropriately in https://github.com/jenkinsci/ec2-plugin/blob/master/src/test/java/hudson/plugins/ec2/ConfigurationAsCodeTest.java

src/main/java/hudson/plugins/ec2/SpotConfiguration.java
```diff
@@ -1,15 +1,20 @@
...
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
-public final class SpotConfiguration {
+public final class SpotConfiguration extends AbstractDescribableImpl<SpotConfiguration>  {
     public final boolean useBidPrice;
-    public final String spotMaxBidPrice;
-    public final boolean fallbackToOndemand;
-    public final int spotBlockReservationDuration;
+    public String spotMaxBidPrice;
```
res0nance: I dislike this change as this is now public and modifiable

timja: It doesn't appear to be used outside this plugin so you could probably make it private https://github.com/search?q=org%3Ajenkinsci+spotMaxBidPrice&type=Code

multani: I making them private, should we add a note or something in the changelog to say it may break compatibility?

src/test/resources/hudson/plugins/ec2/UnixData.yml
multani: I added a test when it's there and another one when it's not.

multani:
The test failure is actually a genuine bug: I tried manually to save a config where "Set bid price" was not checked, and it appears again as "checked" when reloading the page.

multani:
[Found the issue](https://github.com/jenkinsci/ec2-plugin/pull/443/commits/d313b7775290aea5e8dba7d8a6afac391383ddac) ( :crossed_fingers: ): I forgot to remove the accessing the `spotConfig` attribute when I moved the SpotConfiguration jelly file to a dedicated file.

multani:
Arf, apparently, the "Check Current Spot Price" button doesn't work anymore ... :man_facepalming:

multani:
I don't think the test failure is related to my branch, is it possible to restart the build?

timja:
> I don't think the test failure is related to my branch, is it possible to restart the build?

if you close and reopen the PR it'll trigger it again

multani:
Hi, is there anything else that I must do on this pull request?
AIFAK, the tests are now passing and I (should have!) addressed all the issues :)

jetersen:
it's up to the maintainers for the plugin to review and merge 😅 I was only pinged to review because of the JCasC side of things.

src/main/java/hudson/plugins/ec2/CloudHelper.java
```diff
@@ -66,4 +73,39 @@ static Instance getInstance(String instanceId, EC2Cloud cloud) throws AmazonClie
         }
         return instances.get(0);
     }
+
+    @CheckForNull
+    static Image getAmiImage(AmazonEC2 ec2, String ami) {
+        List<String> images = new LinkedList<>();
```
varyvol, line 79: You could maybe use Collections.singleton() or empty() to simplify this (even if you only moved this).

multani: @varyvol This would be better? (genuine question, my Java skills are rusty)

```java
        List<String> images = Collections.emptyList();
        images.add(ami);
        List<String> owners = Collections.emptyList();
        List<String> users = Collections.emptyList();
```

timja: @multani something like this:

```suggestion
        List<String> images = Collections.singletonList(ami);
```

varyvol: Precisely that, including @timja 's suggestion.

multani:
@thoulen Thanks a lot for the merge!

I wanted to updated [JENKINS-59048](https://issues.jenkins.io/browse/JENKINS-59048) to reflect the changes, but I keep getting HTTP 403 errors whatever I'm trying to do (watch the ticket, comment, etc.) :man_shrugging:

timja:
@multani update it on https://issues.jenkins-ci.org, the domain you're on doesn't work

PR #445, opened 2020-03-19
[JENKINS-56927] Implementation of credentials for private SSH keys used by EC2 plugin

djesionek:
This should fix the issue described in [JENKINS-56927](https://issues.jenkins-ci.org/browse/JENKINS-56927).

I tried to keep as much of the credential logic itself contained inside the _EC2Cloud_ class but still in a way that the private key itself is not being stored longer than necessary outside of the credential API.

This implementation also takes care of migrating the old plaintext ssh configuration to the credentials API. This was checked in a test environment and also the unit tests themselves are not modified more than necessary and some of them still relied on the old storage method and a migration is performed in each one. 
In my tests the privateKey was still included in plaintext in the CasC yaml when exporting, so I managed to find a quite hacky solution for that by returning _null_ in the getter and creating a static method resolving the credential of a _EC2Cloud_ object to a _EC2PrivateKey_ which is now used instead in all places. If there is a better way for that I have not found it yet...

All of this introduced one new dependency: _org.jenkins-ci.plugins:ssh-credentials:1.14_
As the _aws-credentials_ plugin has a different dependency to the _credentials_ package it also was fixed to be min version 2.1.17 to prevent dependency errors. I have not checked how low we could go here but I'm also not sure how important this might be here.

Eucalyptus implementations also were updated accordingly with this but could not be tested explicitly due to lack of a test environment on my side.


djesionek:
Hi @jvz and @res0nance ,
should I have assigned anyone from you maintainers as reviewer for this PR to be picked up or is something else missing?

jvz:
Ooh, this looks nice! I've added some people as reviewers to look through this.

Wadeck:
_(I will not have the time to review it, sorry)_

pom.xml
```diff
@@ -101,11 +101,21 @@ THE SOFTWARE.
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>credentials</artifactId>
+            <version>2.1.17</version>
```
jvz, line 107:
```suggestion
            <version>2.3.5</version>
```

This is fairly old otherwise.

pom.xml
```diff
@@ -101,11 +101,21 @@ THE SOFTWARE.
...
+        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>aws-credentials</artifactId>
             <version>1.11</version>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>ssh-credentials</artifactId>
+            <version>1.14</version>
```
jvz, line 117:
```suggestion
            <version>1.18.1</version>
```

Same.

src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -187,6 +196,16 @@ public ListBoxModel doFillRegionItems(
             return model;
         }
 
+        public ListBoxModel doFillSshKeysCredentialsIdItems(@QueryParameter String sshKeysCredentialsId) {
+
```
jvz, line 200: Needs a permission check for `Node/Configure` or admin.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -914,9 +984,22 @@ public FormValidation doCheckUseInstanceProfileForCredentials(@QueryParameter bo
             return FormValidation.ok();
         }
 
-        public FormValidation doCheckPrivateKey(@QueryParameter String value) throws IOException, ServletException {
+        public FormValidation doCheckSshKeysCredentialsId(@QueryParameter String value) throws IOException, ServletException {
+
```
jvz, line 988: Now that this method is looking up system credentials, it should have a permission check added.

src/main/java/hudson/plugins/ec2/Eucalyptus.java
```diff
@@ -71,11 +86,21 @@ public String getDisplayName() {
             return "Eucalyptus";
         }
 
+        public ListBoxModel doFillSshKeysCredentialsIdItems(@QueryParameter String sshKeysCredentialsId) {
+
```
jvz, line 90: This also needs a permission check.

src/main/resources/hudson/plugins/ec2/AmazonEC2Cloud/config-entries.jelly
```diff
@@ -36,8 +36,8 @@ THE SOFTWARE.
   <f:entry title="${%Region}" field="region">
     <f:select/>
   </f:entry>
-  <f:entry title="${%EC2 Key Pair's Private Key}" field="privateKey">
-    <s:secretTextarea value="${instance.privateKey.privateKeySecret}"/>
+  <f:entry field="sshKeysCredentialsId" title="${%EC2 Key Pair's Private Key}">
+    <c:select includeUser="true"/>
```
jvz, line 40:
```suggestion
    <c:select includeUser="false"/>
```

User-scoped credentials won't be relevant here as there's no build associated.

src/main/resources/hudson/plugins/ec2/Eucalyptus/config-entries.jelly
```diff
@@ -31,8 +31,8 @@ THE SOFTWARE.
   <f:entry title="${%Use EC2 instance profile to obtain credentials}" field="useInstanceProfileForCredentials">
     <f:checkbox />
   </f:entry>
-  <f:entry title="${%EC2 Key Pair's Private Key}" field="privateKey">
-    <s:secretTextarea value="${instance.privateKey.privateKeySecret}"/>
+  <f:entry field="sshKeysCredentialsId" title="${%EC2 Key Pair's Private Key}">
+    <c:select includeUser="true"/>
```
jvz, line 35:
```suggestion
    <c:select includeUser="false"/>
```

Same here.

djesionek:
@jvz how do you handle changing the dependency versions generally? After applying your changes and setting newer dep versions the enforcer requires me to also change trilead-api 1.0.3->1.0.5 and structs 1.17->1.20. The newer version of ssh-credentials you suggested also requires at least jenkins at version 2.190.1. 
Are there any considerations to make regarding backwards compatibility of any sorts or should they be just the newest version at the time of implementation of features?

jvz:
Jenkins 2.190.1 is fairly old as it is, so that's not a bad baseline now. The current LTS is in 2.222.x, so that's a few LTS releases behind already. I'll defer that to one of the other maintainers, though.

djesionek:
> Jenkins 2.190.1 is fairly old as it is, so that's not a bad baseline now. The current LTS is in 2.222.x, so that's a few LTS releases behind already. I'll defer that to one of the other maintainers, though.

I'm still trying to find the right combination of dependency versions as bumping the jenkins version results only in more dependency issues that I can not interpret at this point. So suggestions are much appreciated

jvz:
Perhaps we need to adopt a Jenkins [bom](https://github.com/jenkinsci/bom/)?

djesionek:
> Perhaps we need to adopt a Jenkins [bom](https://github.com/jenkinsci/bom/)?
> 
> Edit: which has apparently moved into [the core repo](https://github.com/jenkinsci/jenkins/blob/master/bom/pom.xml).

That might depend on whether we want to or will be able stay below Jenkins 2.195

djesionek:
@jvz you were right, the bom is apparently one solution. Maybe this helps preventing such hassle in future. Is this something I should include here in this PR probably with some cleanup (of the plugin versions no longer necessary in the pom) or leave the cleanup for another task?

jvz:
You can do the dependency update and bom in a separate PR.

djesionek:
> You can do the dependency update and bom in a separate PR.

Does it make sense then to revert the dependency changes here so the PR can be actually built and then reviewed independently to the dependency topic?

jvz:
> Does it make sense then to revert the dependency changes here so the PR can be actually built and then reviewed independently to the dependency topic?

Yeah, go ahead. Thanks!

pom.xml
```diff
@@ -134,7 +146,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>trilead-api</artifactId>
-            <version>1.0.3</version><!-- TODO: BUMP when trilead split is in LTS and is the baseline -->
+            <version>1.0.5</version><!-- TODO: BUMP when trilead split is in LTS and is the baseline -->
```
res0nance, line 149: That comment should be dropped

pom.xml
```diff
@@ -250,6 +262,13 @@ THE SOFTWARE.
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
+                <groupId>org.jenkins-ci.main</groupId>
+                <artifactId>jenkins-bom</artifactId>
```
res0nance, line 267: wasn't the whole point of bom to stop having to specify specific plugin versions?

src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -24,29 +24,32 @@
...
 import javax.annotation.Nullable;
 import javax.servlet.ServletException;
 
 import jenkins.model.Jenkins;
 
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.DataBoundSetter;
-import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.*;
```
res0nance, line 52: No star imports please

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -33,6 +34,9 @@
 import com.cloudbees.plugins.credentials.SystemCredentialsProvider;
 import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
 import com.cloudbees.plugins.credentials.domains.Domain;
+import com.cloudbees.plugins.credentials.domains.DomainRequirement;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.model.*;
```
res0nance, line 39: Same here

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -45,15 +49,7 @@
...
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.EnumSet;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
res0nance, line 52: Here as well

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -156,8 +143,8 @@
     @CheckForNull
     @Deprecated
     private transient Secret secretKey;
-
-    private final EC2PrivateKey privateKey;
+    @CheckForNull
+    private String sshKeysCredentialsId;
```
res0nance, line 147: the `privateKey` variable should not be removed it should be marked as transient

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -265,8 +304,13 @@ public String getCredentialsId() {
         return credentialsId;
     }
 
+    public String getSshKeysCredentialsId() {
```
res0nance, line 307:
```suggestion
    @CheckForNull
    public String getSshKeysCredentialsId() {
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -896,6 +940,33 @@ public static URL checkEndPoint(String url) throws FormValidation {
         }
     }
 
+    public static EC2PrivateKey resolvePrivateKey(EC2Cloud cloud){
```
res0nance, line 943:
```suggestion
    @CheckForNull
    public static EC2PrivateKey resolvePrivateKey(EC2Cloud cloud){
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -317,7 +361,7 @@ public SlaveTemplate getTemplate(Label label) {
      */
     public synchronized KeyPair getKeyPair() throws AmazonClientException, IOException {
         if (usableKeyPair == null)
-            usableKeyPair = privateKey.find(connect());
+            usableKeyPair = resolvePrivateKey(this).find(connect());
```
res0nance, line 364: This method can return null so this can result in an NPE

src/test/java/hudson/plugins/ec2/AmazonEC2CloudUnitTest.java
```diff
@@ -36,6 +39,10 @@
 * Unit tests related to {@link AmazonEC2Cloud}, but do not require a Jenkins instance.
 */
 public class AmazonEC2CloudUnitTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
```
res0nance, line 44: Why do we need jenkinsrule?

src/test/java/hudson/plugins/ec2/EC2PrivateKeyTest.java
```diff
@@ -41,6 +44,12 @@
     @Rule
     public JenkinsRule r = new JenkinsRule();
 
+    @Before
+    public void init(){
+        // Tests using the BouncyCastleProvider failed without that
+        Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
```
res0nance, line 50: All the tests fail without this?

pom.xml
jvz: Yes; the version overrides specified in pom.xml can be removed with this bom applied.

src/test/java/hudson/plugins/ec2/EC2PrivateKeyTest.java
jvz: Probably an issue in Java 9+? I read about an incompatibility with the module path and how JCA is designed, but I can't find an official bug about it.

src/test/java/hudson/plugins/ec2/AmazonEC2CloudUnitTest.java
djesionek: _AmazonEC2Cloud_ now relies on a call to SystemCredentialsProvider.getInstance() which is not available without a jenkinsrule. Is there another way to solve this?

src/test/java/hudson/plugins/ec2/EC2PrivateKeyTest.java
djesionek: For some reason this is not possible anymore on my current machine. I removed it for now.

src/test/java/hudson/plugins/ec2/AmazonEC2CloudUnitTest.java
res0nance: AFAIK, no. If it is needed thats fine then

pom.xml
djesionek: Removing the versions from all plugin deps does not work for me. I also couldn't find a reference that this should work in neither bom code ( core-bom and plugin-bom )

src/main/java/hudson/plugins/ec2/EC2Cloud.java
djesionek: @res0nance is there any reason to leave it marked as transient if it is not being used at all?
As far as I understand it it will still be passed through the constructor after a plugin update from a version where it was used and will be migrated to a credential. So this would be left as an unused variable.

jvz:
I don't maintain this plugin anymore, and I can't see what changed since my last review. Do you have a diff?

djesionek:
> I don't maintain this plugin anymore, and I can't see what changed since my last review. Do you have a diff?

The first commit after you requested changes was ab02e81 so you should be able to diff locally or on github from this commit on.

jvz:
Thanks for the tip! That made things easier to re-review.

djesionek:
@res0nance is there anything I might have missed that is preventing further reviews?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
res0nance: IIRC it wont call the constructor, it might cause deserialization issues if the field no longer exists

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -172,14 +174,18 @@
...
+                       String instanceCapStr, List<? extends SlaveTemplate> templates, String roleArn, String roleSessionName) {
         super(id);
         this.useInstanceProfileForCredentials = useInstanceProfileForCredentials;
         this.roleArn = roleArn;
         this.roleSessionName = roleSessionName;
         this.credentialsId = credentialsId;
-        this.privateKey = new EC2PrivateKey(privateKey);
+        this.sshKeysCredentialsId = sshKeysCredentialsId;
+
+        if (this.sshKeysCredentialsId == null && privateKey != null){
```
res0nance, line 186: This should probably be in `readResolve()`

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -944,9 +1037,34 @@ public FormValidation doCheckUseInstanceProfileForCredentials(@QueryParameter bo
...
+
+            StandardListBoxModel result = new StandardListBoxModel();
+
+            return result
+                    .includeMatchingAs(Jenkins.getAuthentication(), Jenkins.get(), BasicSSHUserPrivateKey.class, Collections.<DomainRequirement>emptyList(), CredentialsMatchers.always())
+                    .includeMatchingAs(ACL.SYSTEM, Jenkins.get(), BasicSSHUserPrivateKey.class, Collections.<DomainRequirement>emptyList(), CredentialsMatchers.always())
+                    .includeCurrentValue(sshKeysCredentialsId);
+        }
+
+        public FormValidation doCheckSshKeysCredentialsId(@QueryParameter String value) throws IOException, ServletException {
```
res0nance, line 1051: I'm not 100% certain but it might need to `@RequirePOST`

src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
```diff
@@ -82,6 +86,11 @@
 
     @Rule public JenkinsRule r = new JenkinsRule();
 
+    @Before
+    public void prepareCredentialStore(){
```
res0nance, line 90: Empty?

src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
```diff
@@ -171,7 +167,11 @@ private WinConnection connectToWinRM(EC2Computer computer, EC2AbstractSlave node
                             Thread.sleep(sleepBetweenAttempts);
                             continue;
                         }
-                        String password = node.getCloud().getPrivateKey().decryptWindowsPassword(passwordData);
+                        EC2PrivateKey ec2PrivateKey = node.getCloud().resolvePrivateKey();
+                        String password = "";
+                        if (ec2PrivateKey != null){
```
res0nance, line 172: If I have to decrypt the password but there is no private key is there any point to continuing?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -926,6 +1002,23 @@ public static URL checkEndPoint(String url) throws FormValidation {
         }
     }
 
+    private static BasicSSHUserPrivateKey getSshCredential(String id){
```
res0nance, line 1005:
```suggestion
    @CheckForNull
    private static BasicSSHUserPrivateKey getSshCredential(String id){
```

src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
djesionek: I was not sure what other way to go, but continuing with the empty password will most definitely fail which should happen anyways in the case of a missing private key. Any suggestion for better error handling here?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
djesionek: No other doCheckX method has this annotation in place. How to find out?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
djesionek: According to the comment at the readResolve() method call I interpret it as method caring about setting the template parents while this part checks if a migration to credentials is necessary here which in my opinion would not be part of this topic. Am I missing something here?

src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
djesionek: Forget to clean up, will be deleted.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
djesionek: Added it

src/main/java/hudson/plugins/ec2/EC2Cloud.java
djesionek: My observations in debugging showed that the constructor is called here and the migration seemed to also work but should maybe be checked to be removed in future. I rewrote it to similar to the handling of the other deprecated and transient attributes now.

djesionek:
The build fails because it uses version 2.164.1 of Jenkins for some reason now. This changed in the last days. Before versions above the required version 2.190.3 according to my updated pom.xml were used.
@res0nance will your PR #465 fix this issue or do I need to change something else here?

djesionek:
@res0nance @jvz @julienduchesne @thoulen what am I missing that prevents further reviews to happen? Some of my comments are still open for your feedback.
I would be happy to fix everything necessary to get this through before bigger changes in the master branch lead to further merge-conflicts.

jvz:
I've already approved this.

src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
res0nance: We should probably log it and throw an exception. I believe it will terminate the instance in this case

src/main/java/hudson/plugins/ec2/EC2Cloud.java
res0nance: @jvz @MRamonLeon do either of you know if this is required?

src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
djesionek: A few lines above the missing passwordData is handled by calling `continue` in the whole connection loop with the expectation that the passwordData will be added between the loops. 
Do you think it would lead to other issues handling a null privateKey exactly like that, expecting that the logs are read an the user configured the right credential?

```suggestion
                        if (ec2PrivateKey == null){
                            logger.println("Waiting for privateKey to be available. Consider checking the credentials in the cloud configuration. Sleeping 10s. ");
                            Thread.sleep(sleepBetweenAttempts);
                            continue;
                        }
                        String password = ec2PrivateKey.decryptWindowsPassword(passwordData);
```

Seems for me like a legit way to handle this as nothing gets terminated but a correct configuration gets enforced.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
markyjackson-taulia: There was a ping on twitter regarding this Pr and jira issue this is addressing so I thought I would jump in to help. A @RequirePOST is required when doing `doCheckX`
@djesionek

MRamonLeon: I would say no, as long as the method is not modifying any state. Maybe @daniel-beck can confirm.

src/main/java/hudson/plugins/ec2/win/EC2WindowsLauncher.java
djesionek: I now made this change as it makes sense to me like that. Thanks for pointing that out

src/main/java/hudson/plugins/ec2/EC2Cloud.java
daniel-beck: https://www.jenkins.io/doc/developer/security/form-validation/ explains it.

At a glance, depends on exactly what `getSshCredential` does. If there are potentially network lookups for credentials involved, it's safer to `@RequirePOST` (or `@POST`).

djesionek: It only calls the `CredentialsProvider`, so `@RequirePOST` is not necessary in that case?

daniel-beck: Unsure how expensive network lookups can become with external credentials providers.

I mean, when in doubt add it; it's a reasonable default.

djesionek: Makes sense, didn't thought about external credentials providers at all. So I will add it. Thanks :+1:

pom.xml
djesionek: @res0nance is this topic holding back further reviews and approvals?

froblesmartin:
What is missing on this PR to be merged? @djesionek @markyjackson-taulia

djesionek:
AFAIK I adressed all the requested changes from the reviewers that reviewed this PR. I'm just waiting for either feedback or a merge.
Maybe @res0nance can tell what is missing.

froblesmartin:
> AFAIK I adressed all the requested changes from the reviewers that reviewed this PR. I'm just waiting for either feedback or a merge.
> Maybe @res0nance can tell what is missing.

Well, at least what you can see below, there is a conflict on the EC2Cloud.java file

djesionek:
Thanks for the info, that one seems to be new. I won't be able to resolve it until Sunday probably.

froblesmartin:
@djesionek FYI in https://issues.jenkins-ci.org/browse/SECURITY-1883 @daniel-beck asked @jetersen to have a look into this.

djesionek:
@froblesmartin thanks for the info somehow my jenkins community account got lost that's probably why I missed that.

jetersen:
not sure what I can assist with.

djesionek:
And I'm not sure what is missing for getting this merged other than maybe further reviews when necessary. Maybe @daniel-beck knows more?

jvz:
This plugin needs active maintainers. I've re-added the `adopt-this-plugin` label as this hasn't received new maintainers since I thought I last added that.

froblesmartin:
> This plugin needs active maintainers. I've re-added the `adopt-this-plugin` label as this hasn't received new maintainers since I thought I last added that.

@jvz I have been thinking since a while about maintaining a plugin to start contributing. This could be a perfect moment 😋 .

Following https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/, I am going to send an email to the DL requesting access 👍

res0nance (approved):

djesionek:
Thanks, hope it gets released soon 😄

batmat:
@res0nance @MRamonLeon @fcojfernandez @froblesmartin could you please consider doing a release for this? Thanks!

MRamonLeon:
Released as https://github.com/jenkinsci/ec2-plugin/releases/tag/ec2-1.52

varyvol:
How was migration tested here? I had one EC2 cloud configured with 1.50.3 and when updated to 1.52, the cloud is there but no credentials are in place for the private key. Also an error is displayed because the "doCheckSshKeysCredentialsId" requires POST and the validation performed when the credential is selected is not using that.

MRamonLeon:
^ @res0nance

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -165,14 +162,20 @@
...
         super(id);
         this.useInstanceProfileForCredentials = useInstanceProfileForCredentials;
         this.roleArn = roleArn;
         this.roleSessionName = roleSessionName;
         this.credentialsId = credentialsId;
-        this.privateKey = new EC2PrivateKey(privateKey);
+        this.sshKeysCredentialsId = sshKeysCredentialsId;
+
+        if (this.sshKeysCredentialsId == null && ( this.privateKey != null || privateKey != null)){
+            migratePrivateSshKeyToCredential(this.privateKey != null ? this.privateKey.getPrivateKey() : privateKey);
```
daniel-beck, line 175: A migration call in the constructor makes no sense. This should be in `readResolve`. This was probably never tested.

djesionek:
I made a PR with a fix for that and retested the migration scenario there manually. Maybe it would be good to create an automated test for that but at this point I have no clue how this could be achieved 😄 
Sorry that I missed this!

erik-hakansson-wcar:
This broke our setup as the old key wasn't migrated.

varyvol:
ec2-1.53 fixes it.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -946,6 +1019,24 @@ public static URL checkEndPoint(String url) throws FormValidation {
         }
     }
 
+    @CheckForNull
+    private static BasicSSHUserPrivateKey getSshCredential(String id){
+
+        BasicSSHUserPrivateKey credential = CredentialsMatchers.firstOrNull(
+                CredentialsProvider.lookupCredentials(
+                        BasicSSHUserPrivateKey.class, // (1)
```
jglick, line 1027: [As noted in Jira](https://issues.jenkins.io/browse/JENKINS-56927?focusedId=398071&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-398071) this is incorrect: should be using the interface type, `SSHUserPrivateKey`.

jglick: I see #602 changing other things. I think you meant #520.

res0nance: Oh yes you're right!

PR #447, opened 2020-04-01
Conditionally set AWS network configuration directly on EC2 request if associate public ip is true

jzyeezy:
I upgraded the EC2 plugin to the latest available version (`ec2-1.49.1`) and found that our cloud was spinning up workers with `null` public DNS hostnames:

An agent trying to connect would show these logs:

> Connecting to null on port 22, with timeout 10000.

Additionally, I observed in the EC2 UI that this new instance's public DNS ip address was blank (`i-003...` corresponds to `ec2-1.49.1`; `i-00f...` corresponds to `ec2-1.45`).

Screen Shot 2020-04-01 at 5 13 51 PM

The underlying AWS VPC, subnet, and security groups were left unchanged between the plugin versions. Our EC2 cloud template is configured (and also unchanged) with the "Public DNS" connection strategy and "Associate Public IP" set to `false`.

Under `ec2-1.49.1`, the necessary AWS network configuration needed to provision a new EC2 instance is only set on the request's network interface specification, which seems to be inadequate for an instance to be created with a public DNS name.

Reviving the `getAssociatePublicIP()` conditions to set the target subnet and security group ids on the `RunInstancesRequest` culminates in the correct AWS EC2 request to provision a new worker with our current configuration.

I smoke-tested by uploading a patched plugin with these changes. New instances are created and can connect to master once again!

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -733,7 +733,11 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
         InstanceNetworkInterfaceSpecification net = new InstanceNetworkInterfaceSpecification();
         if (StringUtils.isNotBlank(subnetId)) {
-            net.setSubnetId(subnetId);
+            if (getAssociatePublicIp()) {
+                net.setSubnetId(subnetId);
+            } else {
+                riRequest.setSubnetId(subnetId);
+            }
 
             diFilters.add(new Filter("subnet-id").withValues(subnetId));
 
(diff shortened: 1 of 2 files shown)


jzyeezy:
@jvz I tried to look into that before opening, but it doesn't look like there were ever tests removed as part of removing the conditions that I added in this PR.  See ` git diff ec2-1.45..ec2-1.49.1 src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java`. I'm not sure what I would be the most helpful kind of test to add here?

jvz:
Any tests that would fail if your PR were changed incorrectly in the future. Regression tests mainly. Otherwise, features pop in and out of existence over time when their expected behavior is unknown. :)

jzyeezy:
Ok, I'll have to look into adding tests on another day. I do want to point [this](https://github.com/jenkinsci/ec2-plugin/pull/387) PR out though.

jvz:
If you find there's no reasonable way to add tests, let me know. I don't want to block the PR for no reason.

jzyeezy:
@jvz I have added a couple tests in the latest commit. Please let me know if there's a better direction than the one I took!

jzyeezy:
@jvz when will this get merged?

jvz:
I'm not the primary maintainer here, so I'm not really sure. @varyvol?

res0nance:
I think this is ok to merge as is, I left it to see if others wanted to take a look at it

res0nance:
Looks like this caused an issue https://issues.jenkins-ci.org/browse/JENKINS-62090
@jzyeezy

PR #449, opened 2020-04-17
Use Spot Pagination API

bwagner5:
The spot API is unbounded if MaxResults is not specified which can cause long waits and possible connection timeouts if an account has a lot of spot instances. This commit specifies the MaxResults parameter to the spot-describe-instances-request API to trigger the pagination response and wraps the code in a pagination do-while loop. 

I tested this on a clean jenkins install with 105 spot instances. The response properly paged through the additional 5 instances in the second page of results.


res0nance:
Could you add a test to this?

bwagner5:
I've tested my change locally and everything works.  Can you show me an example of an existing test so I can model mine similarly?

bwagner5:
added a unit test.

PR #453, opened 2020-04-28
FIX: Bad filter passed to DescribeSpotInstanceRequests API

bwagner5:
https://github.com/jenkinsci/ec2-plugin/pull/449 contained a bug where a DescribeInstances filter "instance-state-name" was propagated to the DescribeSpotInstanceRequests API call and is unsupported within that call.  I was able to reproduce the bug on the 1.50 release and tested that it is working properly with this patch.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -413,7 +413,7 @@ public abstract class EC2Cloud extends Cloud {
             }
         } while(result.getNextToken() != null);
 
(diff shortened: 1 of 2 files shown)


bwagner5:
@res0nance  I'm open to suggestion, but I don't think there's a great way to test this. The problem stemmed from the fact that the code was too coupled with the DescribeInstances call. I've broken up the functions to make them more isolate which I think should prevent too many args from being shared between the two calls.

res0nance:
Tested manually and counting works thanks for the patch

PR #460, opened 2020-05-08
[JENKINS-62195] Replace off by no to guarantee compatibility with old ssh clients

MRamonLeon:
See [JENKINS-62195](https://issues.jenkins-ci.org/browse/JENKINS-62195)

Versions of ssh client previous to OpenSSH_7.6 didn't have the new options used by the plugin:
* accept-new
* off

The option **_off_** seems to be synonym of the **_no_** option even on recent versions of the plugin. See help of OpenSSH_8.0p1 version:

```
     StrictHostKeyChecking
...
 If this flag is set to “no” or “off”, ssh will automatically add new host keys to the user known hosts files and allow connections to hosts with changed hostkeys to proceed, subject to some restrictions.
...
```

So we change to use this option instead of _off_.

It was expected (on 7.6 version) that the **_off_** option differs from the **_no_** one in the future, but this future doesn't yet come, so we will use the **_no_** option. Attending the documentation of the 7.6 version:

```
The second setting "off", is a synonym for the current behaviour of StrictHostKeyChecking=no: accept new host keys, and continue connection for hosts with incorrect hostkeys. A future release will change the meaning of  StrictHostKeyChecking=no to the behaviour of "accept-new"
```

That's why we didn't use the **_no_** option since the beginning.

Desired reviewers: @res0nance @fcojfernandez @jvz @varyvol @alecharp @rsandell @oleg-nenashev @MarkEWaite @daniel-beck

### DIFF
--- a/src/main/java/hudson/plugins/ec2/HostKeyVerificationStrategyEnum.java
+++ b/src/main/java/hudson/plugins/ec2/HostKeyVerificationStrategyEnum.java
@@ -34,7 +34,7 @@ public enum HostKeyVerificationStrategyEnum {
     CHECK_NEW_HARD("check-new-hard", "yes", new CheckNewHardStrategy()),
     CHECK_NEW_SOFT("check-new-soft", "accept-new", new CheckNewSoftStrategy()),
     ACCEPT_NEW("accept-new", "accept-new", new AcceptNewStrategy()),
-    OFF("off", "off", new NonVerifyingKeyVerificationStrategy());
+    OFF("off", "no", new NonVerifyingKeyVerificationStrategy());
     
     private final String displayText;
     private final SshHostKeyVerificationStrategy strategy;


MarkEWaite:
OpenSSH 8.1 on OpenBSD 6.6 lists the following in its `man 5 ssh_config`:

```
     StrictHostKeyChecking
             If this flag is set to yes, ssh(1) will never automatically add
             host keys to the ~/.ssh/known_hosts file, and refuses to connect
             to hosts whose host key has changed.  This provides maximum
             protection against man-in-the-middle (MITM) attacks, though it
             can be annoying when the /etc/ssh/ssh_known_hosts file is poorly
             maintained or when connections to new hosts are frequently made.
             This option forces the user to manually add all new hosts.

             If this flag is set to "accept-new" then ssh will automatically
             add new host keys to the user known hosts files, but will not
             permit connections to hosts with changed host keys.  If this flag
             is set to "no" or "off", ssh will automatically add new host keys
             to the user known hosts files and allow connections to hosts with
             changed hostkeys to proceed, subject to some restrictions.  If
             this flag is set to ask (the default), new host keys will be
             added to the user known host files only after the user has
             confirmed that is what they really want to do, and ssh will
             refuse to connect to hosts whose host key has changed.  The host
             keys of known hosts will be verified automatically in all cases.

```

timja:
Seems like there's already a PR for this: https://github.com/jenkinsci/ec2-plugin/pull/455 and it updates the docs as well

res0nance:
If its ok with folks, I'll merge #455 as it updates docs and close this. Any objections?

MRamonLeon:
Fine to me. Please add https://github.com/jenkinsci/ec2-plugin/pull/461 as well, I had some troubles while testing the problems with the log conflict. Thank you @res0nance !!

PR #462, opened 2020-05-10
[JENKINS-62231] Warn about downgrade issue

MRamonLeon:
See [JENKINS-62231](https://issues.jenkins-ci.org/browse/JENKINS-62231)

I've created this PR to warn users about the problem downgrading from any of the security releases. The issue triggers you loose your cloud configuration. 

The workaround is to remove the ec2.xml file before restarting Jenkins. This file is created if you click on the button _Dismiss these messages_ of the new AdministrativeMonitor which warns you when you have templates with a weak template. If you don't do that, there isn't any issue.

I was pondering to put this kind of warning in the `index.jelly` file, to have it very visible when you see the update on the pluginManager. Because here, is very late IMO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,6 +615,15 @@ This issue can be solved in different ways:
       
     ![](docs/images/init-scripts.png)
 
+## Downgrade from 1.50.2, 1.49.2, 1.46.3 to a previous version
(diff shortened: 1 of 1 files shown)


res0nance:
I've added release notes for the security releases with this warning included. since this file is for older releases people might not think to look in here. We can included it in both WDYT?

PR #454, opened 2020-05-05
Fix CI Errors

res0nance:
WIP to fix the build on ci.jenkins.io

PR #464, opened 2020-05-11
[JENKINS-62231] Move start logic to postInitialize to avoid OldDataMo…

MRamonLeon:
See [JENKINS-62231](https://issues.jenkins-ci.org/browse/JENKINS-62231) for full story

To avoid the plugin to fail when loading its configuration if any modification has been done we move the logic from `start` to `postInitialize` method. It may help for future versions upgrade/downgrade ability.

@res0nance @fcojfernandez @alecharp @varyvol @rsandell ARggghhhh I cannot even request reviewers

### DIFF
--- a/src/main/java/hudson/plugins/ec2/PluginImpl.java
+++ b/src/main/java/hudson/plugins/ec2/PluginImpl.java
@@ -33,6 +33,8 @@ import jenkins.model.Jenkins;
 import java.io.IOException;
 import java.util.logging.Logger;
 
+import java.io.IOException;
+
 /**
  * Added to handle backwards compatibility of xstream class name mapping.
  */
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/ec2/PluginImpl.java
```diff
@@ -85,7 +76,15 @@ public String getDisplayName() {
...
     @Override
-    public void postInitialize() {
+    public void postInitialize() throws IOException {
+        // backward compatibility with the legacy class name
+        Jenkins.XSTREAM.alias("hudson.plugins.ec2.EC2Cloud", AmazonEC2Cloud.class);
+        Jenkins.XSTREAM.alias("hudson.plugins.ec2.EC2Slave", EC2OndemandSlave.class);
+        // backward compatibility with the legacy instance type
+        Jenkins.XSTREAM.registerConverter(new InstanceTypeConverter());
+
+        load();
```
MRamonLeon, line 86: maybe just `load()` is needed

MRamonLeon:
The build is failing because the [recommendedConfiguration uses jenkins.version=2.164.1](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPlugin.groovy#L258) and [we updated jenkins.version to 2.176.4](https://github.com/jenkinsci/ec2-plugin/pull/454/files#diff-600376dffeb79835ede4a0b285078036R86). We should use a different configuration or revert the bump of the jenkins.version

MRamonLeon:
I believe my changes on Jenkinsfile wasn't taken into account

res0nance:
There seems to be an issue on windows as well, I'll work on it later, right now I'm trying to get the builds going again quickly

MRamonLeon:
@res0nance green again, great!! Do you think you would be able to cut a release with this one and more interesting, with #455, to avoid pain on people upgrading with old ssh clients?

MRamonLeon:
@res0nance would you be able to cut a release with #455 and other improvements? To avoid pain on people upgrading with old ssh clients.

PR #470, opened 2020-05-22
get root device fix

obstiger:
Ran into a situation where rootMapping wasn't the first item in the list causing some errors.

This PR attempts to fix:
- Getting the root device mapping through the actual root device name.


res0nance:
In what situations does this happen and can this be tested?

PR #471, opened 2020-05-22
reduce number of getPasswordData calls

obstiger:
If the AMI was not setup to generate passwords, AWS does not generate passwords. In such cases, ec2 plugin keeps making EC2 api attempting to get the password. This exhausts the allowed EC2 api calls from aws side and AWS starts throttling the api calls, especially if multiple nodes are spun up at the same time.

This PR attempts to address this issue by reducing the frequency of GetPasswordData calls using exponential back off with full jitter.

Here is an example of the log messages printed out:
image


res0nance:
This seems to come with the problem that it would increase the time it takes for the node to connect by a large amount

PR #492, opened 2020-09-02
Add Support for Host Tenancy Option

ykhandelwal913:
- Added tenancy as dropdown to Slave Template.
- Added support for tenancy for OnDemand Instances only, as this doesn't pertain to Spot Instances.
- Updated SlaveTemplateTest.testTenancy to ensure testing is set as expected.
- Migration of older value of useDedicatedTenancy is taken care.

More details: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-instance.html

Tenancy: host -> Your instance runs on a Dedicated Host, which is an isolated server with configurations that you can control.


ykhandelwal913:
@res0nance I don't have access to add a reviewers. Can you please review the PR or add the reviewers?

ykhandelwal913:
Can someone review this PR?

ykhandelwal913:
@fcojfernandez Can you please review this PR?

fcojfernandez:
I won't have any kind of availability until October, sorry.

Requesting review from the new maintainer @froblesmartin

froblesmartin:
> I won't have any kind of availability until October, sorry.
> 
> Requesting review from the new maintainer @froblesmartin

Hi! Thanks @fcojfernandez for including me here. 

I have been pretty busy since I joined, will try to get into the project this week and hopefully will be able to review the PR with some knowledge 😋

ykhandelwal913:
Thanks @fcojfernandez @froblesmartin

froblesmartin:
@ykhandelwal913 I have been checking your PR.

You have basically added a new property for "host" tenancy, but this should not be a new property, as you are basically replacing the Placement object previously stored in "riRequest" here: https://github.com/jenkinsci/ec2-plugin/blob/176ea49e3028e7500916e57d7b8da09b2e928adc/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L941

I think the tenancy should become an object with 3 possible values, "default", "dedicated", and "host" as it is declared like this in the AWS documentation. This would be a drop-down list as it is for example the "ConnectionStrategy" or the "AMITypeData", with the default value as "default".

froblesmartin (dismissed):
Do not create a second variable for host tenancy, instead use one variable that gives the 3 possible values.

ykhandelwal913:
Thanks @froblesmartin for looking into it.  I thought of doing it then this below condition makes me to create separate entry.


Tenancy Value | Description
-- | --
default | An instance launched into the VPC runs on shared hardware by default, unless you explicitly specify a different tenancy during instance launch.
dedicated | An instance launched into the VPC is a Dedicated Instance by default, unless you explicitly specify a tenancy of host during instance launch. You cannot specify a tenancy of default during instance launch.

Let me know if i am thinking wrongly.

froblesmartin:
> Thanks @froblesmartin for looking into it. I enabled this new field because this doesn't need to be set on VPC. If you have a dedicated hosts attached to your aws account and you want to launch instance out of if, we will need to provide the host tenancy.
> Default and Dedicated Tenancy, needs to be set on VPC.
> 
> > To create Dedicated Instances, you can do the following:
> > Create the VPC with the instance tenancy set to dedicated (all instances launched into this VPC are Dedicated Instances).
> > Create the VPC with the instance tenancy set to default, and specify a tenancy of dedicated for any instances when you launch them.
> 
> Tenancy Value	Description
> default	An instance launched into the VPC runs on shared hardware by default, unless you explicitly specify a different tenancy during instance launch.
> dedicated	An instance launched into the VPC is a Dedicated Instance by default, unless you explicitly specify a tenancy of host during instance launch. You cannot specify a tenancy of default during instance launch.
> Let me know if i am thinking wrongly.

Not really.

The last thing you are mentioning are **VPC default configuration**. **This plugin does not configure a VPC, it just creates Jenkins Agents (still called Slaves)**, and when you create an Instance in EC2 it allows you to specify which tenancy you want. Depending on the default VPC configuration, it would allow the 3 possibilities (default, dedicated and host) if the default VPC tenancy is "default", and only 2 possibilities (dedicated and host) if the default VPC tenancy is "dedicated".

In our case, as we do not configure the VPC, we only care about the EC2 instance tenancy. The perfect way would be to [check the VPC configuration](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/ec2/model/Vpc.html#getInstanceTenancy--) meanwhile you are configuring this plugin in Jenkins, and then display the 3 possible values or only 2 of them, but that would require more work, so I feel okay leaving it up to the person configuring it. If they have a VPC with tenancy "dedicated" they should know that they have to use only "dedicated" or "host" when configuring the Jenkins Agent (Slave) in this plugin.

ykhandelwal913:
exactly thats why i have added the new field. If we add a drop down ,having 3 values in this, can fail the existing setup if VPC is not configured to use dedicated or default. 
This host tenancy is set to dedicated host if user account has some dedicated host and they want to bring the instance out of it , can choose host tenancy. 
If VPC is configured for dedicated , they can choose the dedicated tenancy. 

This way it will impact the existing setup.
image

froblesmartin:
> exactly thats why i have added the new field. If we add a drop down ,having 3 values in this, can fail the existing setup if VPC is not configured to use dedicated or default.
> This host tenancy is set to dedicated host if user account has some dedicated host and they want to bring the instance out of it , can choose host tenancy.
> If VPC is configured for dedicated , they can choose the dedicated tenancy.
> 
> Hence leaving it to the user to choose without setting any default value. That way it will impact the existing setup.
> image

I am sorry but I am not sure if I understand what you mean @ykhandelwal913 . 

The only possibility for it to fail would be:
- VPC is configured with "dedicated" tenancy.
- User does not change the Agent tenancy leaving it by default in "default".

Even with your way of doing it, if the user leaves by default unchecked both checkboxes (Use dedicated tenancy and Use host tenancy) it would fail, as we are not checking the tenancy on the VPC.

Also it is not acceptable to have by default the "host" tenancy, as it has to be "default".

ykhandelwal913:
What I meant it, for default and dedicated , vpc has to configured to support both. For host tenancy, user account should have dedicated host. So if I set tenancy  as default , it can fail the existing user setup of vpc is not configured to support. If vpc is not configured to support anything and user account do not have dedicated host, user can unchecked both boxes and things will work fine.

ykhandelwal913:
@froblesmartin let me know if i am thinking otherwise.

froblesmartin:
> What I meant it, for default and dedicated , vpc has to configured to support both. For host tenancy, user account should have dedicated host. So if I set tenancy as default , it can fail the existing user setup of vpc is not configured to support. If vpc is not configured to support anything and user account do not have dedicated host, user can unchecked both boxes and things will work fine.

Yes, but then what will happen for all the people that have their VPC configured with the default value (which is the default, so probably most of the people)?? Once we release this feature it will start failing for everybody.

This PR has to have the logic to keep the same values that are already configured to do not break anything on a update, and the default values, should be the DEFAULT ones, no dedicated or host.

That is my opinion. Let's see if any other maintainer can come here and say something 😅

ykhandelwal913:
Thanks @froblesmartin . Not sure if i have confused you by sharing the screenshot where i have shown useHostTenancy as checked tick mark.

So here is the what i meant to say:

- For the existing one it will not impact as **by default this checkbox is unchecked** and this is a new field added.
- For the new Cloud and AMI , user can select the options based on his requirements.

froblesmartin:
> Thanks @froblesmartin . Not sure if i have confused you by sharing the screenshot where i have shown useHostTenancy as checked tick mark.
> 
> So here is the what i meant to say:
> 
> * For the existing one it will not impact as **by default this checkbox is unchecked** and this is a new field added.
> * For the new Cloud and AMI , user can select the options based on his requirements.

What I actually did not understand correctly is this:
> exactly thats why i have added the new field. If we add a drop down ,having 3 values in this, can fail the existing setup if VPC is not configured to use dedicated or default.

Let me try one more time. So, what you are saying is that, if someone has already configured the EC2 plugin to use `dedicated` tenancy, and they have their VPC configured with `dedicated` as well, and we replace this with a drop down list with `default` as the default value, it will fail? Is that what you mean?

Also, just to clarify:

- **VPC configured with `default` tenancy:** allows to choose between `default`, `dedicated` and `host` tenancy when starting new EC2 instances.
- **VPC configured with `dedicated` tenancy:** allows to choose only between `dedicated` and `host` tenancy.

ykhandelwal913:
> Let me try one more time. So, what you are saying is that, if someone has already configured the EC2 plugin to use `dedicated` tenancy, and they have their VPC configured with `dedicated` as well, and we replace this with a drop down list with `default` as the default value, it will fail? Is that what you mean?

No. I meant if VPC is not configured for default tenancy and we set the tenancy default value as default, it will fail as i think for them tenancy should be dedicated.
if there is nothing configured, we can set the tenancy Type value as default.

I just wanted to have the minimal impact to already existing configuration , hence created a new checkbox item so user can select it if their account has dedicated host configured.

froblesmartin:
> > Let me try one more time. So, what you are saying is that, if someone has already configured the EC2 plugin to use `dedicated` tenancy, and they have their VPC configured with `dedicated` as well, and we replace this with a drop down list with `default` as the default value, it will fail? Is that what you mean?
> 
> No. I meant if VPC is not configured for default tenancy and we set the tenancy default value as default, it will fail as i think for them tenancy should be dedicated.
> if there is nothing configured, we can set the tenancy Type value as default.
> 
> I just wanted to have the minimal impact to already existing configuration , hence created a new checkbox item so user can select it if their account has dedicated host configured.

Got you now. Then what needs to be done here is a log to migrate the old values to the new ones. In case they have the `useDedicatedTenancy` as `true`, in that case set the drop down list with the `dedicated` option, in any other case, set it to `default`.

In this way we do not interfere at all with existing configuration, and we do not create 2 checkboxes that overlap the same configuration as only one should be possible to be checked 😉 .

ykhandelwal913:
Sure @froblesmartin . 
Let me try that. 
Adding new field item as a checkbox was easy ;)

ykhandelwal913:
@froblesmartin I have pushed the requested changes where tenancy attribute is coming as dropdown now.
image

ykhandelwal913:
@froblesmartin can you please review this PR?

ykhandelwal913:
@froblesmartin can you please review this PR?

froblesmartin:
> @froblesmartin can you please review this PR?

Hi @ykhandelwal913 , I am sorry, I am currently overloaded, I have an exam on the 13th and I expect to be able to review this PR decently the following week.

ykhandelwal913:
@froblesmartin Hope your exam went well. Can you please review this PR?

ykhandelwal913:
@froblesmartin can you please review this PR?

taneishamitchell:
can someone please review this PR, we need this

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -23,6 +23,7 @@
  */
 package hudson.plugins.ec2;
 
+import com.amazonaws.services.ec2.model.*;
```
res0nance, line 26: No star imports

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -92,7 +82,6 @@
     public final String jvmopts; // e.g. -Xmx1g
     public final boolean stopOnTerminate;
     public final String idleTerminationMinutes;
-    public final boolean useDedicatedTenancy;
```
res0nance, line 95: You cannot simply remove a field, this can cause deserialization issues make that field transient if you do not need it, its readResolve() method should do the appropriate migration

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -140,7 +129,9 @@
...
 
-    public EC2AbstractSlave(String name, String instanceId, String templateDescription, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
+    public final Tenancy tenancy;
+
+    public EC2AbstractSlave(String name, String instanceId, String templateDescription, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses, Tenancy tenancy)
```
res0nance, line 134: Do not modify existing constructors only add a new one and deprecate the old one

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -36,37 +36,39 @@
...
+        this(instanceId, templateDescription, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, false, launchTimeout, amiType, tenancy);
     }
 
     @Deprecated
-    public EC2OndemandSlave(String instanceId, String templateDescription, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType)
+    public EC2OndemandSlave(String instanceId, String templateDescription, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean usePrivateDnsName, int launchTimeout, AMITypeData amiType, Tenancy tenancy)
```
res0nance, line 49: Same here about constructors

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -22,51 +22,7 @@
...
-import com.amazonaws.services.ec2.model.SpotInstanceRequest;
-import com.amazonaws.services.ec2.model.SpotMarketOptions;
-import com.amazonaws.services.ec2.model.SpotPlacement;
-import com.amazonaws.services.ec2.model.SpotPrice;
-import com.amazonaws.services.ec2.model.StartInstancesRequest;
-import com.amazonaws.services.ec2.model.StartInstancesResult;
-import com.amazonaws.services.ec2.model.Subnet;
-import com.amazonaws.services.ec2.model.Tag;
-import com.amazonaws.services.ec2.model.TagSpecification;
+import com.amazonaws.services.ec2.model.*;
```
res0nance, line 25: Same here about star imports

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -354,10 +310,10 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
                          String tmpDir, String userData, String numExecutors, String remoteAdmin, AMITypeData amiType, String jvmopts,
                          boolean stopOnTerminate, String subnetId, List<EC2Tag> tags, String idleTerminationMinutes, int minimumNumberOfInstances,
                          int minimumNumberOfSpareInstances, String instanceCapStr, String iamInstanceProfile, boolean deleteRootOnTermination,
-                         boolean useEphemeralDevices, boolean useDedicatedTenancy, String launchTimeoutStr, boolean associatePublicIp,
+                         boolean useEphemeralDevices, String launchTimeoutStr, boolean associatePublicIp,
                          String customDeviceMapping, boolean connectBySSHProcess, boolean monitoring,
                          boolean t2Unlimited, ConnectionStrategy connectionStrategy, int maxTotalUses,
-                         List<? extends NodeProperty<?>> nodeProperties, HostKeyVerificationStrategyEnum hostKeyVerificationStrategy) {
+                         List<? extends NodeProperty<?>> nodeProperties, HostKeyVerificationStrategyEnum hostKeyVerificationStrategy, Tenancy tenancy) {
```
res0nance, line 316: Again about constructors here

res0nance (changes requested):
I have not tested any of these changes but this breaks compatibility with existing scripts

taneishamitchell:
@ykhandelwal913 it would be good to include a way to pass the [host ID](https://docs.aws.amazon.com/cli/latest/reference/ec2/run-instances.html) for the target dedicated host?

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
ykhandelwal913: Thanks @res0nance , Added the review comments.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
ykhandelwal913: Thanks @res0nance , Added the review comments.

ykhandelwal913:
@froblesmartin  @res0nance  I have added the requested changes. Please review.

ykhandelwal913:
@froblesmartin @res0nance can you review?

res0nance (changes requested):
readResolve() should still need to be edited to handle migration to the new variable

ykhandelwal913:
@res0nance readResolve() method in slaveTemplate class?

ykhandelwal913:
@froblesmartin @res0nance can you review?

ykhandelwal913:
@froblesmartin @res0nance can you review?

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -51,12 +51,19 @@ public EC2OndemandSlave(String instanceId, String templateDescription, String re
...
     }
 
-    @DataBoundConstructor
+    @Deprecated
     public EC2OndemandSlave(String name, String instanceId, String templateDescription, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
             throws FormException, IOException {
 
+        this(name, instanceId, templateDescription, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, nodeProperties, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, launchTimeout, amiType, connectionStrategy, maxTotalUses, Tenancy.Default);
```
res0nance, line 58: Breaks backwards compat it should use the old usededicatedtenancy variable instead of always passing tenancy default

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -160,15 +165,23 @@ public EC2AbstractSlave(String name, String instanceId, String templateDescripti
...
+        this.tenancy = tenancy != null ? tenancy : Tenancy.Default;
         readResolve();
         fetchLiveInstanceData(true);
     }
 
+    @Deprecated
+    public EC2AbstractSlave(String name, String instanceId, String templateDescription, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, boolean useDedicatedTenancy, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses)
+            throws FormException, IOException {
+
+        this(name, instanceId, templateDescription, remoteFS, numExecutors, mode, labelString, launcher, retentionStrategy, initScript, tmpDir, nodeProperties, remoteAdmin, jvmopts, stopOnTerminate, idleTerminationMinutes, tags, cloudName, launchTimeout, amiType, connectionStrategy, maxTotalUses, Tenancy.Default);
```
res0nance, line 182: Breaks backwards compat it should use the old usededicatedtenancy variable instead of always passing tenancy default

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -420,11 +424,33 @@ public SlaveTemplate(String ami, String zone, SpotConfiguration spotConfig, Stri
...
+                         InstanceType type, boolean ebsOptimized, String labelString, Node.Mode mode, String description, String initScript,
+                         String tmpDir, String userData, String numExecutors, String remoteAdmin, AMITypeData amiType, String jvmopts,
+                         boolean stopOnTerminate, String subnetId, List<EC2Tag> tags, String idleTerminationMinutes, int minimumNumberOfInstances,
+                         int minimumNumberOfSpareInstances, String instanceCapStr, String iamInstanceProfile, boolean deleteRootOnTermination,
+                         boolean useEphemeralDevices, boolean useDedicatedTenancy, String launchTimeoutStr, boolean associatePublicIp,
+                         String customDeviceMapping, boolean connectBySSHProcess, boolean monitoring,
+                         boolean t2Unlimited, ConnectionStrategy connectionStrategy, int maxTotalUses,
+                         List<? extends NodeProperty<?>> nodeProperties, HostKeyVerificationStrategyEnum hostKeyVerificationStrategy) {
+        this(ami, zone, spotConfig, securityGroups, remoteFS,
```
res0nance, line 443: Breaks backwards compat it should use the old usededicatedtenancy variable instead of always passing tenancy default

res0nance (changes requested):

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
ykhandelwal913: added the backward compatibility.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
ykhandelwal913: added the backward compatibility

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
ykhandelwal913: added the backward compatibility

ykhandelwal913:
sorry @res0nance for asking multiple review. This is my first PR on this project. Your review comments are really helping me to learn.

res0nance (approved):
AFAICT this is ok, but unfortunately I can't test this change

ykhandelwal913:
@res0nance do i need to take any action to get that merge?

ykhandelwal913:
@res0nance Any action needed from my side?

res0nance:
I think the change is ok but i lack the infrastructure to test this properly. Have you tested these changes?

ykhandelwal913:
Yes, i have validated it from my end.

PR #493, opened 2020-09-03
Sts consideration

suthsc:
[JENKINS-63537](https://issues.jenkins-ci.org/browse/JENKINS-63537) - EC2SlaveMonitor terminates agent with a running job when STS session expires
* Adding a isCheckable(Node) method to determine if connectivity should be checked to the node. A node with a running job will not get checked. If the computer associated with the node can not be established or if the node is idle, then normal processing is performed.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2SlaveMonitor.java
+++ b/src/main/java/hudson/plugins/ec2/EC2SlaveMonitor.java
@@ -14,6 +14,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import static java.util.logging.Level.INFO;
+
 /**
  * @author Bruno Meneguello
  */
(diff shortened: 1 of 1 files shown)


suthsc:
@froblesmartin and @MRamonLeon would either of you be able to provide some guidance on how to proceed with this PR. I don't have any traction at the moment.

froblesmartin:
Looks okay for me, would it be possible to add tests for that? 🤔

MRamonLeon (commented):
I don't understand well the solution. Reading the ticket I would say the problem is that on `removeDeadNodes`, when checking `isAlive`, as the token expired, it triggers an exception and the catch code remove the node from the Jenkins instance, per the line:
```
...
removeNode(ec2Slave);
```

The log (on the Jira ticket) seems to second that:

```
17:06:22 Agent EC2 (CLOUD) - NODE (...) was deleted; cancelling node body
```

So if we catch the specific exception for **_token expired_** and just ignored this case, because it's going to be solved by the `EC2ConnectionUpdater` after 1 minute, the next time, the monitor will return the node `isAlive`, so all good.

suthsc:
> So if we catch the specific exception for **_token expired_** and just ignored this case, because it's going to be solved by the `EC2ConnectionUpdater` after 1 minute, the next time, the monitor will return the node `isAlive`, so all good.

@MRamonLeon You understand the situation, and I think your idea would work also. I can look at the exception to see if it can be singled out. Thanks for the suggestion.

suthsc:
@froblesmartin I looked at some of the existing tests, and I have no idea how to proceede. Would you be able to point me to some resources on testing plugins so I can get my head around it? Thanks!

suthsc:
@MRamonLeon The other idea I had was to refresh the STS token as part of the EC2SlaveMonitor run, but I couldn't find a path to accomplish that either.

suthsc:
Closing in favor of #560

PR #498, opened 2020-09-16
Bugfix: Removed @RequirePost and changed ssh key migration behavior

djesionek:
I changed the behavior from my feature added in #445 as suggested by @daniel-beck by moving the migration code to EC2Cloud.readResolve(). I also tested again the migration from 1.51 to 1.52. I could confirm that it wasn't working before and that it works now for me with that change.
Not sure why the @RequirePost lead to issues as it was requested as an addition but maybe someone can clarify if there is a better way than just removing it for now.

Thanks to @varyvol for the bugnotice and @daniel-beck for the hint how to fix it quickly.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -171,12 +171,6 @@ public abstract class EC2Cloud extends Cloud {
         this.credentialsId = credentialsId;
         this.sshKeysCredentialsId = sshKeysCredentialsId;
 
-        if (this.sshKeysCredentialsId == null && ( this.privateKey != null || privateKey != null)){
-            migratePrivateSshKeyToCredential(this.privateKey != null ? this.privateKey.getPrivateKey() : privateKey);
-        }
-        this.privateKey = null; // This enforces it not to be persisted and that CasC will never output privateKey on export
-
-
         if (templates == null) {
             this.templates = Collections.emptyList();
         } else {
@@ -249,6 +243,11 @@ public abstract class EC2Cloud extends Cloud {
 
         for (SlaveTemplate t : templates)
             t.parent = this;
+
+        if (this.sshKeysCredentialsId == null && this.privateKey != null ){
+            migratePrivateSshKeyToCredential(this.privateKey.getPrivateKey());
+        }
+        this.privateKey = null; // This enforces it not to be persisted and that CasC will never output privateKey on export
 
         if (this.accessId != null && this.secretKey != null && credentialsId == null) {
             String secretKeyEncryptedValue = this.secretKey.getEncryptedValue();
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1066,7 +1065,6 @@ public ListBoxModel doFillSshKeysCredentialsIdItems(@QueryParameter String sshKe
                     .includeCurrentValue(sshKeysCredentialsId);
         }
 
-        @RequirePOST
```
MRamonLeon, line 1069: It looks like fine to remove this annotation here, the method is just returning whether the credential-id showed is an actual rsa private key. But better to confirm with some security person: @daniel-beck, @Wadeck, ...

daniel-beck: There's still a permission check.

The docs for this are at https://www.jenkins.io/doc/developer/security/form-validation/ -- please read and if anything specific is unclear, ask.

Wadeck: getSshCredential retrieves credentials, but as the RequirePOST protects against CSRF, there is no issue. The method is already checking for ADMINISTER.

Side note, such methods should not have troubles to use POST for the check. You can change the https://github.com/jenkinsci/ec2-plugin/blob/8702cc4a89bd7c096bba07211e222cde6d931f5d/src/main/resources/hudson/plugins/ec2/AmazonEC2Cloud/config-entries.jelly#L39 (and similar for Eucalyptus) by adding `checkMethod="post"`

Wadeck:
@markyjackson-taulia Please read the comment above about the RequirePOST ;)
> A @RequirePOST is required when doing doCheckX
(from https://github.com/jenkinsci/ec2-plugin/pull/445#pullrequestreview-416564695)

----

> Not sure why the @RequirePost lead to issues as it was requested as an addition

Be sure to test your code before commiting, there is no "manual test" performed everytime on PR and this could lead to broken release sometimes ;)

djesionek:
Now I tested the migration manually after all changes maybe we should release that as a hotfix or something before more updates fail.

varyvol:
@MRamonLeon @res0nance @daniel-beck Tests are passing and even if no test was created to verify the issue, @djesionek has tested it manually. Should we maybe release this and work in the test later?

res0nance (approved):

res0nance:
I think releasing as a bugfix release is reasonable. waiting to see if there are any other comments

batmat:
I think that if really we've had people manually verifying the fix, we _may_ want to merge.

However, given the relative ease of use of `@LocalData`, I would recommend we also have a test accompanying this fix so it's caught next time.
* Docs: https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/recipes/LocalData.html
* Working example: https://github.com/batmat/localdata-jenkinsrule/

varyvol:
See #500 for an automated test.

MRamonLeon (approved):

PR #499, opened 2020-09-16
inclusive language tags: jenkins_slave_type renamed to jenkins_node_type

nitrocode:
I know there are many other master/slave references. This one will only correct the issue with jenkins ec2 worker tags.

This is also consistent with Jenkins using the word `node` for their workers.

Ref: https://tools.ietf.org/id/draft-knodel-terminology-02.html

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Tag.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Tag.java
@@ -42,7 +42,7 @@ public class EC2Tag extends AbstractDescribableImpl<EC2Tag> {
     /**
      * Tag name for the specific jenkins slave type tag, used to identify the EC2 instances provisioned by this plugin.
      */
-    public static final String TAG_NAME_JENKINS_SLAVE_TYPE = "jenkins_slave_type";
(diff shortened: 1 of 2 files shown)


nitrocode:
Hi @fcojfernandez @res0nance @MRamonLeon 

What are some of the thoughts on this ?

res0nance:
This is going to break the node counting. Not sure how to best handle this

nitrocode:
Where is node counting codified?

res0nance:
https://github.com/jenkinsci/ec2-plugin/blob/dfa60b734c7645cd43060070b9de23268636e9f1/src/main/java/hudson/plugins/ec2/EC2Cloud.java#L606

nitrocode:
Thanks. It sounds like it might be best to have logic to count both the `jenkins_slave_type` and `jenkins_node_type` tag names but tag the latter going forward for newly created instances. For the `getGenericFilters`, we could count both tag names. I'll put in some more work.

dannykansas:
What additional changes are needed, if any, for this to be ready for the next release?

res0nance:
> What additional changes are needed, if any, for this to be ready for the next release?

Something to make it backwards compatible, as is this will break backwards compatibility

PR #500, opened 2020-09-18
Test to validate the EC2Cloud private key migration 

varyvol:
Adds to #498 an automated test validating the migration for an EC2 cloud is correctly done.

@MRamonLeon @djesionek

### DIFF
--- a/src/test/java/hudson/plugins/ec2/EC2CloudMigrationTest.java
+++ b/src/test/java/hudson/plugins/ec2/EC2CloudMigrationTest.java
@@ -18,7 +18,7 @@ public class EC2CloudMigrationTest {
     public JenkinsRule r = new JenkinsRule();
 
     @Test
-    @LocalData
+    @LocalData // config.xml file contains an ec2-cloud configuration from a version previous to ec2-1.52
     public void testPrivateKeyMigrationToSshCredentials() {
         assertEquals(1, r.jenkins.clouds.size());
         EC2Cloud cloud = (EC2Cloud) Jenkins.get().getCloud("ec2-myEc2Cloud");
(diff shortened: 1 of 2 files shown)


src/test/java/hudson/plugins/ec2/EC2CloudMigrationTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static org.junit.Assert.*;
+
+public class EC2CloudMigrationTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test
+    @LocalData
```
MRamonLeon, line 21: Could you add some comment explaining that the local data is from a previous version of EC2 to show it clear the migration works? Or maybe in the config.xml you can add a comment noticing the `privateKey` is the actual key and not a credential and it works.

MRamonLeon (commented):
I guess this PR superseeds https://github.com/jenkinsci/ec2-plugin/pull/498 right?

src/test/java/hudson/plugins/ec2/EC2CloudMigrationTest.java
varyvol: Added.

varyvol:
I don't pretend this to supersede the original PR, only to demonstrate how to use LocalData to perform the related testing.

MRamonLeon (approved):

res0nance (approved):

CptPlastic:
This breaks the plugin. We ended up rolling back.

varyvol:
@CptPlastic this only adds a test, so I guess you'll be referring to a previous PR. Which version did you update to?

timja:
this is raised as https://issues.jenkins-ci.org/browse/JENKINS-63707

PR #515, opened 2020-10-14
Add retry logic for connection timeout exception

JamesForsee:
On a private jenkins server we were experiencing a problem where the EC2SlaveMonitor would query aws for an agent and get a response code with the error code 'RequestExpired'. This would result in the slave monitor deleting the agents from jenkins but the instances were still running within aws. Worse the orphaned agents would still count towards the instance cap and eventually prevent us from spinning up additional agents when required.
Adding the timeout condition to the retry logic resolved this problem for us.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/CloudHelper.java
+++ b/src/main/java/hudson/plugins/ec2/CloudHelper.java
@@ -30,7 +30,7 @@ final class CloudHelper {
             try {
                 return getInstance(instanceId, cloud);
             } catch (AmazonServiceException e) {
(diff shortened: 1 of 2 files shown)


res0nance:
Could a test be added for this?

JamesForsee:
I added some testing for the CloudHelper class as requested.

JamesForsee:
@res0nance Any chance this could be merged now that tests have been added

JamesForsee:
@roblesmartin @res0nance Could I get some eyes on this when you get a chance? I've updated the pull request as requested

res0nance (approved):

res0nance (approved):

PR #516, opened 2020-10-16
Add re-submission of tasks during spot interruption disconnects

blanked:
This PR adds a new feature - re-submission of tasks for agents that are disconnected due to spot interruption event in AWS.
Whenever an agent is disconnected, there are checks to determine if it is an unexpected disconnect and if the disconnection is a spot interruption event. If the answer is yes to both, the tasks that were running on the agent will be re-submitted to the queue.

### Motivation

Builds may fail due to spot instances being terminated. This PR can help to reduce the number of build failures for spot interruption events.

### Notes
This may or may not prevent build failures. There doesn't seem to be any documentation on how tasks can be resubmitted. This PR is inspired by another Jenkins plugin that has the suggested behaviour implemented - https://github.com/jenkinsci/ec2-fleet-plugin/blob/master/src/main/java/com/amazon/jenkins/ec2fleet/EC2FleetAutoResubmitComputerLauncher.java

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2ComputerLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/EC2ComputerLauncher.java
@@ -23,15 +23,29 @@
  */
 package hudson.plugins.ec2;
 
+import com.amazonaws.services.ec2.model.SpotInstanceRequest;
+import hudson.model.Action;
+import hudson.model.Actionable;
+import hudson.model.Executor;
+import hudson.model.Queue;
+import hudson.model.Result;
+import hudson.model.Slave;
 import hudson.model.TaskListener;
 import hudson.slaves.ComputerLauncher;
+import hudson.slaves.OfflineCause;
 import hudson.slaves.SlaveComputer;
+import hudson.model.queue.SubTask;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import com.amazonaws.AmazonClientException;
+import jenkins.model.CauseOfInterruption;
+
+import javax.annotation.Nonnull;
 
 /**
  * {@link ComputerLauncher} for EC2 that wraps the real user-specified {@link ComputerLauncher}.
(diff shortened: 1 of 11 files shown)


blanked:
Can someone help to review this PR to see if its ok? It's actually identical to #485 but I opened a new PR so that it's eligible for hacktoberfest 😅

res0nance (approved):
Feature looks very interesting, AFAICT this seems to do what it says but it is a hard to test feature.

blanked:
yeah i'll have a think on how to mock a spot interruption event and see if its possible using the aws sdk. if anyone has any idea on how to do so, that'll be super helpful!

src/main/java/hudson/plugins/ec2/EC2ComputerLauncher.java
```diff
@@ -75,4 +89,95 @@ public void launch(SlaveComputer slaveComputer, TaskListener listener) {
...
+    @Override
+    public void afterDisconnect(SlaveComputer computer, TaskListener listener) {
+        if (computer == null) return;  // potential edge case where computer is null
+
+        Slave node = computer.getNode();
+        if (node instanceof EC2SpotSlave) {
+
+            // checking if its an unexpected disconnection
+            final boolean isUnexpectedDisconnection = computer.isOffline() && computer.getOfflineCause()
+                    instanceof OfflineCause.ChannelTermination;
```
imuqtadir, line 110: Some of the customers complained that ``OfflineCause.ChannelTermination`` was not always triggered for spot interruption. You may be able to dig this further here: https://github.com/jenkinsci/ec2-fleet-plugin/issues/121

dgiffordaudio:
This seems to have been approved in October 2020.  Is this going to be merged soon?  This would be really helpful for us

opajonk:
Yes, this would be really awesome to add - any plans?

minhnnhat-urbanise:
Hello, we're also looking forward to this feature.

schottsfired:
> AFAICT this seems to do what it says but it is a hard to test feature.

> i'll have a think on how to mock a spot interruption event and see if its possible using the aws sdk. if anyone has any idea on how to do so, that'll be super helpful!

It should be possible to test it now with this new AWS feature:
[AWS Fault Injection Simulator now injects Spot Instance Interruptions](https://aws.amazon.com/about-aws/whats-new/2021/10/aws-fault-injection-simulator-injects-spot-instance-interruptions/)

DhruvJ225:
We are looking forward to use this feature.

PR #520, opened 2020-10-22
Use SSHUserPrivateKey instead of BasicSSHUserPrivateKey

chriskilding:
Use the general `SSHUserPrivateKey` type when looking up SSH credentials instead of the `BasicSSHUserPrivateKey` concrete subtype. This will allow the plugin to use any compatible SSH key, not just `BasicSSHUserPrivateKey` instances.

Fixes [JENKINS-63986](https://issues.jenkins-ci.org/browse/JENKINS-63986)

### DIFF
--- a/src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
+++ b/src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
@@ -24,6 +24,7 @@
 package hudson.plugins.ec2;
 
 import com.amazonaws.services.ec2.AmazonEC2;
+import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
 import com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.CredentialsScope;
@@ -32,6 +33,7 @@ import com.cloudbees.plugins.credentials.SystemCredentialsProvider;
 import com.cloudbees.plugins.credentials.domains.Domain;
 import com.gargoylesoftware.htmlunit.html.HtmlForm;
 import com.gargoylesoftware.htmlunit.html.HtmlTextInput;
+import hudson.plugins.ec2.util.TestSSHUserPrivateKey;
 import hudson.util.ListBoxModel;
 import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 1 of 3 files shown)


chriskilding:
Can't proceed on this until we have a way to test the change. What are the relevant tests to modify?

chriskilding:
@res0nance I believe you're one of the maintainers - would you be able to advise where the tests are for using SSH keys from the credentials API, and thus which tests I should modify to ensure this fix works?

res0nance:
Hey Chris, just saw this, give me a bit to double check this bit.

res0nance:
Hey Chris i added a test in #523 which should help you out

src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
```diff
@@ -116,6 +118,32 @@ public void testSshCredentials() throws IOException {
         assertThat(actual.resolvePrivateKey(), notNullValue());
     }
 
+    /**
+     * Ensure that EC2 plugin can use any implementation of SSHUserPrivateKey (not just the default implementation, BasicSSHUserPrivateKey).
+     *
+     * See <a href="https://issues.jenkins-ci.org/browse/JENKINS-63986">JENKINS-63986</a>.
+     */
+    @Test
```
res0nance, line 126:
```suggestion
    @Issue("JENKINS-63986")
    @Test
```
Would probably need an import as well

chriskilding: done

chriskilding:
I've pinged the reporter to ask if he'd like to test a build of the fix in his local Jenkins

chriskilding:
Mark has confirmed that it works.

He also said it might be worth putting out a notice of a breaking change for JCasC, since we have to change the config or our clouds will be unable to start.

johnl2323:
Chris,

We ran into this issue using the AWS Secrets Manager Credential Provider
plugin as the source of our credentials.
We can test this if you can provide a link to the plugin version to install.

johnl

chriskilding:
The plugin build with the fix can be found at:

https://ci.jenkins.io/job/Plugins/job/ec2-plugin/job/PR-520/5/artifact/org/jenkins-ci/plugins/ec2/1.54-rc1218.bd9a8dd69948/ec2-1.54-rc1218.bd9a8dd69948.hpi

We've had a confirmation that it works so it should be good to merge, but if you need the fix in the meantime, here it is :)

johnl2323:
I can confirm that this PR fixes our issue and allows access to the AWS
secrets manager credentials via the secrets manager credential provider
plugin.

PR #528, opened 2020-11-05
[JENKINS-64147] Forwards compat with table-to-div

res0nance:
Still broken on < 2.264 but seems to work on the newest versions.

Not sure what the issue is, any idea @timja


timja:
what's broken? which page?

timja:
You're missing the `taglib` marker file

```
$ git diff --cached
diff --git a/src/main/resources/lib/ec2/taglib b/src/main/resources/lib/ec2/taglib
new file mode 100644
index 0000000..e69de29
 timja at Frankel in ~/projects/jenkinsci/ec2-plugin on table-to-div*
$ git status
On branch table-to-div
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	new file:   src/main/resources/lib/ec2/taglib
```

fixes it

timja (approved):

PR #531, opened 2020-11-09
JENKINS-63982 Support EC2/S3 Endpoints for AWS-CN Partition

els-grazziotinf:
AWS operates different Partitions such as US-GovCloud and AWS-CN.
This patch improves support for Jenkins instances operating
ec2-plugin in the AWS-CN partition as it has a different DNS Suffix
than those in US-based Amazon Web Services.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,5 @@
 ## Changelog
 
-### Next version
-- Support AWS China partition under amazonaws.com.cn for EC2 and S3 endpoints.
-
 ### Version 1.46 (September 25, 2019)
 - Configure release drafter and Dependabot
 - JENKINS-25106 - Support for minimum number of instances
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### Next version
+- Support AWS China partition under amazonaws.com.cn for EC2 and S3 endpoints.
+
 ### Version 1.46 (September 25, 2019)
 - Configure release drafter and Dependabot
 - JENKINS-25106 - Support for minimum number of instances
--- a/src/test/java/hudson/plugins/ec2/AmazonChinaEC2CloudTest.java
+++ b/src/test/java/hudson/plugins/ec2/AmazonChinaEC2CloudTest.java
@@ -1,0 +1,94 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2004-, Kohsuke Kawaguchi, Sun Microsystems, Inc., and a number of other of contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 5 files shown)


MRamonLeon (commented):
How can I test it? I tried to set an alternate ec2 endpoint using `https://ec2.cn-north-1.amazonaws.com.cn` but the regions didn't get populated.

Do I have to be in China or china vpn or maybe some restriction I have?

I would like to take another approach to avoid the hack for comparing the region with `cn-`. For example by changing the connect method to pass to it the region instead of the endpoint, so the SDK is the one which calculate the final endpoint, not us. The only problem is how to pass the proxy configured in Jenkins.

els-grazziotinf:
> How can I test it? I tried to set an alternate ec2 endpoint using `https://ec2.cn-north-1.amazonaws.com.cn` but the regions didn't get populated.

AWS-CN is a different partition, you need to be operating in it (e.g. EC2 Instance locally, or AWS-CN IAM Credentials). If you have AWS "global" credentials and try to access the AWS-CN region it won't do. If you check the logs, you probably received the following response from AWS API:

{code}
An error occurred (InvalidClientTokenId) when calling the GetCallerIdentity operation: The security token included in the request is invalid
{code}

 
> Do I have to be in China or china vpn or maybe some restriction I have?

You would need an AWS Account in AWS-CN. Due Chinese regulations you need identity verification, more information https://www.amazonaws.cn/en/about-aws/china/faqs/#beijing-account-cn

> I would like to take another approach to avoid the hack for comparing the region with `cn-`. For example by changing the connect method to pass to it the region instead of the endpoint, so the SDK is the one which calculate the final endpoint, not us. The only problem is how to pass the proxy configured in Jenkins.

I wish the SDK had a better way of determining the partition, unfortunately I could not find a better way.
I'm not that versed in AWS Java SDK to refactor in the way you suggested, sorry.

src/test/java/hudson/plugins/ec2/AmazonChinaEC2CloudTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.io.IOException;
+import java.util.Collections;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Felipe Grazziotin
+ */
+public class AmazonChinaEC2CloudTest {
```
imuqtadir, line 48: Can we keep the class name generic and not specific to China. Mayb something like `AmazonEC2PartitionCloudTest`. We may have other partitions in future and keeping it generic may help with extensibility. I'm okay with test names, it's just the class naming.

els-grazziotinf: I've renamed as AmazonPartitionEC2CloudTest as suggested, also rebased on top of latest master.

MRamonLeon (approved):
LGTM although I believe the S3 endpoint is never used, so we could remove the method. I approve anyway if you don't want to review this.

imuqtadir (approved):
lgtm!

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog
 
+### Next version
+- Support AWS China partition under amazonaws.com.cn for EC2 and S3 endpoints.
```
res0nance, line 4: We no longer use this, we use the releases page for changelogs so you should revert this

res0nance (approved):

els-grazziotinf:
> LGTM although I believe the S3 endpoint is never used, so we could remove the method. I approve anyway if you don't want to review this.

I was going to remove it but saw Eucalyptus.java also has references to it, decided to play safe.

CHANGELOG.md
els-grazziotinf: My apologies, I've reverted in new commit.

MRamonLeon:
Thank you for your contribution @els-grazziotinf :clap:

PR #533, opened 2020-11-16
[JENKINS-59180] Allow AMI filtering instead of AMI ID

dbnicholson:
Allow finding the image to launch by the various `describeImages` criteria instead of just a specific image ID. Currently you're required to reconfigure the plugin whenever the image is updated, which is inconvenient.

Fixes: https://issues.jenkins.io/browse/JENKINS-59180

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Filter.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Filter.java
@@ -1,0 +1,119 @@
+/*
+ * The MIT License
+ *
+ * Copyright © 2020 Endless OS Foundation LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package hudson.plugins.ec2;
+
+import hudson.model.Descriptor;
+import hudson.model.AbstractDescribableImpl;
+import hudson.Extension;
+import hudson.Util;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+import com.amazonaws.services.ec2.model.Filter;
+
+public class EC2Filter extends AbstractDescribableImpl<EC2Filter> {
+    @Nonnull
+    private final String name;
+
+    /* FIXME: Ideally this would be List<String>, but Jenkins currently
+     * doesn't offer a usable way to represent those in forms. Instead
+     * the values are interpreted as a comma separated list.
+     *
+     * https://issues.jenkins.io/browse/JENKINS-27901
+     */
+    @Nonnull
+    private final String values;
+
+    @DataBoundConstructor
+    public EC2Filter(@Nonnull String name, @Nonnull String values) {
+        this.name = Objects.requireNonNull(name);
+        this.values = Objects.requireNonNull(values);
+    }
+
+    @Override
+    public String toString() {
+        return "EC2Filter{name=\"" + name + "\", values=\"" + values + "\"}";
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null)
+            return false;
+        if (this.getClass() != o.getClass())
+            return false;
+
+        EC2Filter other = (EC2Filter) o;
+        return name.equals(other.name) && getValuesList().equals(other.getValuesList());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, values);
+    }
+
+    @Nonnull
+    public String getName() {
+        return name;
+    }
+
+    @Nonnull
+    public String getValues() {
+        return values;
+    }
+
+    @Nonnull
+    private List<String> getValuesList() {
+        return Stream.of(Util.tokenize(values))
+            .collect(Collectors.toList());
+    }
+
+    /* Helper method to convert EC2Filter to Filter */
+    @Nonnull
+    public Filter toFilter() {
+        return new Filter(name, getValuesList());
+    }
+
+    /* Helper method to convert list of EC2Filter to list of Filter */
+    @Nonnull
+    public static List<Filter> toFilterList(@CheckForNull List<EC2Filter> filters) {
+        return Util.fixNull(filters).stream()
+            .map(EC2Filter::toFilter)
+            .collect(Collectors.toList());
+    }
+
+    @Extension
+    public static class DescriptorImpl extends Descriptor<EC2Filter> {
+        @Override
+        public String getDisplayName() {
+            return "";
+        }
+    }
+}
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -812,12 +812,13 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
      * @return always non-null. This needs to be then added to {@link Hudson#addNode(Node)}.
      */
     public List<EC2AbstractSlave> provision(int number, EnumSet<ProvisionOptions> provisionOptions) throws AmazonClientException, IOException {
+        final Image image = getImage();
         if (this.spotConfig != null) {
             if (provisionOptions.contains(ProvisionOptions.ALLOW_CREATE) || provisionOptions.contains(ProvisionOptions.FORCE_CREATE))
-                return provisionSpot(number, provisionOptions);
+                return provisionSpot(image, number, provisionOptions);
             return null;
         }
-        return provisionOndemand(number, provisionOptions);
+        return provisionOndemand(image, number, provisionOptions);
     }
 
     /**
@@ -857,12 +858,18 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
         LOGGER.info(this + ". " + message);
     }
 
-    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(int number, AmazonEC2 ec2) throws IOException {
-        return makeRunInstancesRequestAndFilters(number, ec2, true);
+    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(Image image, int number, AmazonEC2 ec2) throws IOException {
+        return makeRunInstancesRequestAndFilters(image, number, ec2, true);
     }
 
-    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(int number, AmazonEC2 ec2, boolean rotateSubnet) throws IOException {
-        RunInstancesRequest riRequest = new RunInstancesRequest(ami, 1, number).withInstanceType(type);
+    @Deprecated
+    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(int number, AmazonEC2 ec2) throws IOException {
+        return makeRunInstancesRequestAndFilters(getImage(), number, ec2);
+    }
+
+    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(Image image, int number, AmazonEC2 ec2, boolean rotateSubnet) throws IOException {
+        String imageId = image.getImageId();
+        RunInstancesRequest riRequest = new RunInstancesRequest(imageId, 1, number).withInstanceType(type);
         riRequest.setEbsOptimized(ebsOptimized);
         riRequest.setMonitoring(monitoring);
 
@@ -872,7 +879,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
             riRequest.setCreditSpecification(creditRequest);
         }
 
-        setupBlockDeviceMappings(riRequest.getBlockDeviceMappings());
+        setupBlockDeviceMappings(image, riRequest.getBlockDeviceMappings());
 
         if(stopOnTerminate){
             riRequest.setInstanceInitiatedShutdownBehavior(ShutdownBehavior.Stop);
@@ -883,7 +890,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
         }
 
         List<Filter> diFilters = new ArrayList<>();
-        diFilters.add(new Filter("image-id").withValues(ami));
+        diFilters.add(new Filter("image-id").withValues(imageId));
         diFilters.add(new Filter("instance-type").withValues(type.toString()));
 
         KeyPair keyPair = getKeyPair(ec2);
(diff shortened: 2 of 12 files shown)


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1178,31 +1226,55 @@ private void setupRootDevice(List<BlockDeviceMapping> deviceMappings) {
...
 
-            if (ami.equals(image.getImageId())) {
+    @NonNull
+    private Image getImage() throws AmazonClientException {
+        DescribeImagesRequest request = makeDescribeImagesRequest();
 
-                return image;
-            }
-        }
+        LOGGER.info("Getting image for AMI " + request);
```
froblesmartin, line 1269: Sorry if I am wrong but, Is this sentence right? 
AFAI understand, the request could have no AMI but only filters, right? So it would be something like `"Getting images with request: " + request`?

froblesmartin (changes requested):

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
dbnicholson: Yeah, that's better.

froblesmartin (approved):
LGTM but I am still pretty new. Would like a review from @res0nance @fcojfernandez or @MRamonLeon

src/main/java/hudson/plugins/ec2/EC2Filter.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null)
+            return false;
+        if (this.getClass() != o.getClass())
+            return false;
+
+        EC2Filter other = (EC2Filter) o;
+        return name.equals(other.name) && values.equals(other.values);
```
fcojfernandez, line 74: should be `values = "value1,value2,value3"` equals to `values="value2,value1,value3"`? And what about `values="value1, value2, value3"? According to `toFilter` method, these `equals` and `hashCode` are going to work and you might get duplicated Filter objects.

From UI I see you're allowing a list of EC2Filter

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -787,11 +803,41 @@ public void setHostKeyVerificationStrategy(HostKeyVerificationStrategyEnum hostK
...
+    @DataBoundSetter
+    public void setAmiFilters(List<EC2Filter> amiFilters) {
+        this.amiFilters = amiFilters;
+    }
     
     @Override
     public String toString() {
         return "SlaveTemplate{" +
-                "ami='" + ami + '\'' +
+                "description='" + description + '\'' +
```
fcojfernandez, line 840: why this change? I don't see the relationship with the desired fix and I'd appreciate if you could elaborate. Thanks!

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -857,12 +904,13 @@ private void logProvisionInfo(String message) {
         LOGGER.info(this + ". " + message);
     }
 
-    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(int number, AmazonEC2 ec2) throws IOException {
-        return makeRunInstancesRequestAndFilters(number, ec2, true);
+    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(Image image, int number, AmazonEC2 ec2) throws IOException {
+        return makeRunInstancesRequestAndFilters(image, number, ec2, true);
```
fcojfernandez, line 908: I don't think this concrete change should be an issue since I don't think this package is used out of this plugin, but it's a good practice in the community to deprecate the method and create the new method with the signature changes. The deprecated method can be removed in some versions.
This good practice is applied to all non-private and non-restricted methods just in case there is any proprietary plugin from any user that is extending this code. -> FYI @froblesmartin

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -857,12 +904,13 @@ private void logProvisionInfo(String message) {
...
     }
 
-    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(int number, AmazonEC2 ec2, boolean rotateSubnet) throws IOException {
-        RunInstancesRequest riRequest = new RunInstancesRequest(ami, 1, number).withInstanceType(type);
+    HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(Image image, int number, AmazonEC2 ec2, boolean rotateSubnet) throws IOException {
```
fcojfernandez, line 911: ditto

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1178,31 +1226,55 @@ private void setupRootDevice(List<BlockDeviceMapping> deviceMappings) {
...
+        return Stream.of(Util.tokenize(Util.fixNull(attr)))
+            .collect(Collectors.toList());
     }
 
-    private Image getImage() {
-        DescribeImagesRequest request = new DescribeImagesRequest().withImageIds(ami);
-        for (final Image image : getParent().connect().describeImages(request).getImages()) {
+    @NonNull
+    private DescribeImagesRequest makeDescribeImagesRequest() {
+        List<String> imageIds = Util.fixEmpty(ami) == null ?
```
fcojfernandez, line 1242: Sorry if my question is a nonsense, but any reason to use `Util.fixEmpty` instead of `fixEmptyAndTrim`? if ami is "   ", that would result on a List with the unique value "   "
```suggestion
        List<String> imageIds = fixEmptyAndTrim(ami) == null ?
```

fcojfernandez (changes requested):
See my comment about changing a signature of a non-private method

Besides, if this PR intends to fix https://issues.jenkins.io/browse/JENKINS-59180, shouldn't be the ticket assigned to you and in progress?

src/main/java/hudson/plugins/ec2/EC2Filter.java
dbnicholson: Good call, I didn't think of that. It would be better if equality tested the split tokens instead of the raw string.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
dbnicholson: Because the image ID can be dynamic are this PR. In order to list the image ID, you'd need to do a round trip to AWS every time this is called. Which is actually often because `logProvisionInfo` uses it.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
dbnicholson: Good call. I hadn't thought of any of the methods besides the constructor being used outside of the package.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
dbnicholson: I think because the plugin was already taking whatever was in `ami` I had decided that the only thing I wanted to detect was the empty string that would result from the form being filled in. But I agree that trimming would be more robust.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
dbnicholson: 👍

dbnicholson:
> Besides, if this PR intends to fix https://issues.jenkins.io/browse/JENKINS-59180, shouldn't be the ticket assigned to you and in progress?

I didn't think I could do anything more than make comments on there as an outside contributor. I've changed that now.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
dbnicholson: To expand on that a bit more, you could just leave it as is but in the case where you were using filters it would just say `ami=""`, which wouldn't be helpful if you were looking in the logs. I actually think this is an improvement generally since you could have multiple templates sharing the same image ID. This way you can differentiate them if you use a unique description for each template.

fcojfernandez (approved):
Thank you very much for your contribution @dbnicholson
> I didn't think I could do anything more than make comments on there as an outside contributor. I've changed that now.

Please, feel free and assign any open ticket you decide to work on. That way you will let others
know you are on it.

And let me another piece of advice: try to avoid the force-push command as much as you can. It makes harder review the new changes since the force-push hides them. There are even reviewers that stop reviewing the PRs after a force-push :( 
 
@froblesmartin my concerns are addressed, so I'm approving the PR. I'll let you merge it just in case you want to have a second look at the latest changes.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1178,31 +1236,55 @@ private void setupRootDevice(List<BlockDeviceMapping> deviceMappings) {
...
-            }
-        }
+        LOGGER.info("Getting image for request " + request);
+        List<Image> images = getParent().connect().describeImages(request).getImages();
+        if (images.isEmpty()) {
+            throw new AmazonClientException("Unable to find image for request " + request);
+         }
 
-        throw new AmazonClientException("Unable to find AMI " + ami);
+        // Sort in reverse by creation date to get latest image
```
res0nance, line 1285: We should document this behaviour the the help text

res0nance (approved):
Untested with an aws account but overall looks good

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
dbnicholson: It does say that in the form help for each field, but maybe you mean somewhere else?

dbnicholson:
> And let me another piece of advice: try to avoid the force-push command as much as you can. It makes harder review the new changes since the force-push hides them. There are even reviewers that stop reviewing the PRs after a force-push :(

It's unfortunately the only way to properly maintain commit history on GitHub. That or someone applies the fixups before merging. Do you have something that does that? Or maybe you do squashed merges? I'm fine following whatever flow you normally use. There is a link on the text that says force push that shows the differences from the previous state.

froblesmartin:
@dbnicholson is there anything else you want to change before merging, or may I merge already? 😉

dbnicholson:
Nothing more from me. Go ahead and merge :smiley_cat:

aviadcye:
Hi,
Sorry, but no matter what I do, I can't seem to get this new configuration to work. I always get the error "The request must contain the parameter ImageId". However if I set the ImageId, all other filters are ignored... Am I missing something?

Steps to reproduce.

1. Retrofitting an existing AMI configuration that uses a specific ImageId.
2. Removed the "AMI ID" set, so the filed is blank.
3. Set "AMI Owners" & "AMI Users" to self (as per example test in code).
4. Set "AMI Filters", Added just "Name" (also tried all lower case) with the value of the "Jenkins_docker_agent_base_-_7.12.20a". Also tired with "Jenkins_docker_agent_base*".
5. Click on "Check AMI".

Thank you, Aviad.

dbnicholson:
Does your AWS account have an AMI named `Jenkins_docker_agent_base_-_7.12.20a`? Using an owner of `self` says your account is the owner of the AMI.

dbnicholson:
Also, if you look in the logs, you should see the results of the `DescribeImages` call.

aviadcye:
> Does your AWS account have an AMI named `Jenkins_docker_agent_base_-_7.12.20a`? Using an owner of `self` says your account is the owner of the AMI.

Yes i have this AMI in my account.



> Also, if you look in the logs, you should see the results of the DescribeImages call.

So when I use "Name" i get the error "The filter 'Name' is invalid" (the tag key is "Name", so I don't get why... but OK...
So i tried using "name", and I just didn't get a log entry.

dbnicholson:
`name` and a tag named `Name` are different things. If you want to filter on a `Name` tag, you have to use a filter of `tag:Name`. This is in the AWS documentation linked to in the help text.

aviadcye:
I tried that as well... didn't  work... :/
Btw, I think an example in the plugin context documentation before linking
to the aws docs would help...

dbnicholson:
Can you post the logs with the `DescribeImages` request and response? Have you tried looking up this combination with some like awscli?

aviadcye:
As I said, when I use the "check ami" button, there are no logs unless it
has a filter error, and I'm not even sure ab that.
I'll try to create the definition, and simply attempt to create the
agent... maybe that will give logs.
About awscli, ill try that as soon as possible and report back.

dbnicholson:
The Check AMI button only works if you supply an image ID due to limitations in Jenkins. I thought that was clear that the button was not relevant when using filters.

aviadcye:
Well this is interesting...
Even though the "Check AMI" button, still doesn't work as expected, trying to spawn an agent, worked on the first go...
That is, with simply leaving the imageid blank and adding the filter of "tag:Name = Jenkins_docker_agent_base", I was able to create an agent (logs below).

So this is just a lacking in the check AMI button functionality. 
IMHO, it should both be able to do the same as spawning an agent can, and provide logs... because I gather now that the log I got about the filter "Name" no being valid, was after I had saved the configuration.

At any rate, this is totally usable as it is right now.
Thank you for an awesome feature, I'll start using it immediately :)

`2020-12-31 15:01:36.746+0000 [id=140760]	INFO	hudson.plugins.ec2.SlaveTemplate#getImage: Getting image for request {ExecutableUsers: [],Filters: [{Name: tag:Name,Values: [Jenkins_docker_agent_base*]}],ImageIds: [],Owners: []}
2020-12-31 15:01:37.557+0000 [id=140760]	INFO	hudson.plugins.ec2.SlaveTemplate#logProvisionInfo: SlaveTemplate{description='test', labels=''}. Considering launching
2020-12-31 15:01:37.557+0000 [id=140760]	INFO	hudson.plugins.ec2.SlaveTemplate#logProvisionInfo: SlaveTemplate{description='test', labels=''}. Setting Instance Initiated Shutdown Behavior : ShutdownBehavior.Terminate
2020-12-31 15:01:37.728+0000 [id=140760]	INFO	hudson.plugins.ec2.SlaveTemplate#logProvisionInfo: SlaveTemplate{description='test', labels=''}. Looking for existing instances with describe-instance: {Filters: [{Name: image-id,Values: [ami-yepyepeyep]}, {Name: instance-type,Values: [t1.micro]}, {Name: key-name,Values: [jenkins-slaves]}, {Name: tag:jenkins_server_url,Values: [https://jenkins.dev.example.com/]}, {Name: tag:jenkins_slave_type,Values: [demand_test]}],InstanceIds: [],}
2020-12-31 15:01:39.154+0000 [id=140760]	INFO	hudson.plugins.ec2.SlaveTemplate#logProvisionInfo: SlaveTemplate{description='test', labels=''}. Return instance: {AmiLaunchIndex: 0,ImageId: ami-yepyepeyep,InstanceId: i-idididid,InstanceType: t1.micro,KeyName: jenkins-slaves,LaunchTime: Thu Dec 31 15:01:38 GMT 2020,Monitoring: {State: disabled},Placement: {AvailabilityZone: eu-west-1a,GroupName: ,Tenancy: default,},PrivateDnsName: ip-172-31-19-43.eu-west-1.compute.internal,PrivateIpAddress: 172.31.19.43,ProductCodes: [],PublicDnsName: ,State: {Code: 0,Name: pending},StateTransitionReason: ,SubnetId: subnet-09271c40,VpcId: vpc-031c1064,Architecture: x86_64,BlockDeviceMappings: [],ClientToken: 59eb1abe-d117-4183-bfff-b674b60fb26f,EbsOptimized: false,EnaSupport: true,Hypervisor: xen,ElasticGpuAssociations: [],ElasticInferenceAcceleratorAssociations: [],NetworkInterfaces: [{Attachment: {AttachTime: Thu Dec 31 15:01:38 GMT 2020,AttachmentId: eni-attach-045b1c07fe3064f89,DeleteOnTermination: true,DeviceIndex: 0,Status: attaching},Description: ,Groups: [{GroupName: default,GroupId: sg-c86505b3}],Ipv6Addresses: [],MacAddress: 06:54:38:0d:a6:0f,NetworkInterfaceId: eni-0e558a4d70a8bd16f,OwnerId: actactact,PrivateDnsName: ip-172-31-19-43.eu-west-1.compute.internal,PrivateIpAddress: 172.31.19.43,PrivateIpAddresses: [{Primary: true,PrivateDnsName: ip-172-31-19-43.eu-west-1.compute.internal,PrivateIpAddress: 172.31.19.43}],SourceDestCheck: true,Status: in-use,SubnetId: subnet-09271c40,VpcId: vpc-031c1064,InterfaceType: interface}],RootDeviceName: /dev/sda1,RootDeviceType: ebs,SecurityGroups: [{GroupName: default,GroupId: sg-c86505b3}],SourceDestCheck: true,StateReason: {Code: pending,Message: pending},Tags: [{Key: jenkins_slave_type,Value: demand_test}, {Key: jenkins_server_url,Value: https://jenkins.dev.example.com/}],VirtualizationType: hvm,CpuOptions: {CoreCount: 1,ThreadsPerCore: 1},CapacityReservationSpecification: {CapacityReservationPreference: open,},Licenses: [],MetadataOptions: {State: pending,HttpTokens: optional,HttpPutResponseHopLimit: 1,HttpEndpoint: enabled}}
2020-12-31 15:01:39.249+0000 [id=140760]	INFO	h.p.ec2.EC2RetentionStrategy#start: Start requested for EC2 (AWS-connector--eu-west-1(Ireland)((DEV)) - test (i-idididid)
2020-12-31 15:01:39.249+0000 [id=151193]	INFO	hudson.plugins.ec2.EC2Cloud#log: Launching instance: i-idididid`

PR #439, opened 2020-03-09
JENKINS-60482 Alternate Endpoint fix

troymohl:
This fixes a bug in the EC2 Alternate URL endpoint configuration reported in https://issues.jenkins-ci.org/browse/JENKINS-60482. Additionally, small bug fix to fix cases if an empty endpoint got passed to the AmazonEC2 client initialization.

~This PR has https://github.com/jenkinsci/ec2-plugin/pull/438 as a parent, as my configuration required that fix to be in place in order to make this fix.~
This has been fixed in master, so that PR was no longer needed.

PR #537, opened 2020-11-23
[JENKINS-60482] Persist the alternate ec2 endpoint

MRamonLeon:
See https://issues.jenkins-ci.org/browse/JENKINS-60482

Simplified version of https://github.com/jenkinsci/ec2-plugin/pull/439

Reasoning:

The field `Alternate EC2 endpoint` has only one purpose, to fill up the `Region` field. If you're in an environment where the default endpoint is not reachable, you can specify here the full url to use instead. Like: `https://ec2.us-west-1.amazonaws.com`. Once the field is populated, the connection and all other operations are performed using this region, so they should go fine.

It's true that every time you enter the configuration page, the `Alternate EC2 endpoint` is empty, so the `Region` field never gets populated until you set the endpoint properly. So it causes a lot of frustration and misunderstanding, IMO.

This PR tries to resolve the problem by persisting this field, so every time you reach the configuration page, the field is established, so the region field is populated correctly. This PR is simpler than https://github.com/jenkinsci/ec2-plugin/pull/439 where we use the `Alternate EC2 endpoint` as the full new AWS endpoint (IIUC), which I think it's not necessary.

I've recorded a video to show the behavior: https://photos.app.goo.gl/aADg5roUcKbMqLKj8

CC: @jkfernan @GaToRAiD @troymohl

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
@@ -64,6 +64,8 @@ public class AmazonEC2Cloud extends EC2Cloud {
      */
     private String region;
 
+    private String altEC2Endpoint;
+
     public static final String CLOUD_ID_PREFIX = "ec2-";
 
     private boolean noDelayProvisioning;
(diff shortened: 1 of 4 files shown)


fcojfernandez (approved):

res0nance (approved):
Looks ok, perhaps you should add some form of test for its persistence, perhaps just extend one of the Casc test?

MRamonLeon:
I've added a new test @res0nance , fine to you?

res0nance (approved):

res0nance:
> I've added a new test @res0nance , fine to you?

Looks good

alecharp (approved):

PR #538, opened 2020-11-24
[JENKINS-61161] Avoid infamous UI connection strategy warning

MRamonLeon:
See https://issues.jenkins.io/browse/JENKINS-61161

Avoid the ugly and erroneous _Could not find selected connection strategy_ every time one changes the connection strategy.

See:
Screenshot from 2020-11-24 14-16-51

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -2040,7 +2040,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
         public FormValidation doCheckConnectionStrategy(@QueryParameter String connectionStrategy) {
             return Stream.of(ConnectionStrategy.values())
-                    .filter(v -> v.equalsName(connectionStrategy))
+                    .filter(v -> v.name().equals(connectionStrategy))
                     .findFirst()
                     .map(s -> FormValidation.ok())
                     .orElse(FormValidation.error("Could not find selected connection strategy"));


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -2040,7 +2040,7 @@ public ListBoxModel doFillConnectionStrategyItems(@QueryParameter String connect
 
         public FormValidation doCheckConnectionStrategy(@QueryParameter String connectionStrategy) {
             return Stream.of(ConnectionStrategy.values())
-                    .filter(v -> v.equalsName(connectionStrategy))
+                    .filter(v -> v.name().equals(connectionStrategy))
```
MRamonLeon, line 2043: Previously: Public_IP == PUBLIC_IP ??
Now:        PUBLIC_IP == PUBLIC_IP ??

alecharp (approved):

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
fcojfernandez: is it not the same? why is it fixing the error?

https://github.com/jenkinsci/ec2-plugin/blob/2f348aa519338aac64dd19954b0b41a62a8ea6fc/src/main/java/hudson/plugins/ec2/ConnectionStrategy.java#L15-L17

res0nance: I too am interested in Francisco's point

MRamonLeon: `equalsName` is comparing the text added in the constructor (`Public IP` lowercase), but the `connectionStrategy` variable contains [the declared name of the enum](https://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#name()) (`PUBLIC IP` uppercase), so they were never equals.

fcojfernandez: Oh! I see now! Really good point Ramon

fcojfernandez (approved):

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
res0nance: Interesting, then shouldn't `equals()` also produce the desired behaviour?

res0nance (approved):

amedee:
Regression? I have this error again in EC2 plugin version 1.56

jensenbox:
I can corroborate - this is still an issue. We use CasC and it will not work under any condition. The knock on effect is that after a configuration update and restart of Jenkins, no EC2 instance can be created. Perhaps the logic of this validation function is flawed? Frankly I do not mind seeing the actual enum value show up in as the form value `PUBLIC_IP` - it gives a clear understanding of the underlying value. So perhaps just yank the `name` property - it would sure make translations that much better too (do not compare the human value to the internal value) - just spitballing here as this is a real pain.

radnov:
This is still an issue in 1.59 🤕  

@amedee, @jensenbox Have you guys found any intermediate solution to this?

Doing "Apply" on the Clouds configuration from the UI "fixes" the Connection Strategy problem. Having to `POST $JENKINS_URL/configureClouds/configure` on each JCasC reload is just nasty.

res0nance:
> This is still an issue in 1.59 🤕
> 
> @amedee, @jensenbox Have you guys found any intermediate solution to this?
> 
> Doing "Apply" on the Clouds configuration from the UI "fixes" the Connection Strategy problem. Having to `POST $JENKINS_URL/configureClouds/configure` on each JCasC reload is just nasty.

I can recreate the issue when adding a new template which is fixed in #626

radnov:
@jensenbox See [this comment](https://github.com/jenkinsci/ec2-plugin/pull/626#issuecomment-852951180) in case you're still facing the JCasC issue. 👍

amedee:
As of 8 hours and 39 minutes ago, I no longer work at the company where 
we had this issue.

I no longer have access to the Jenkins there so I can't check the 
configuration.

I am unable to help you, sorry.

albertodebortoli:
This is more broken now with 1.60 than before. I literally tried every possible combination of fields but the connection strategy always falls back to "Public DNS".

res0nance:
> This is more broken now with 1.60 than before. I literally tried every possible combination of fields but the connection strategy always falls back to "Public DNS".

Fixed in #627 has to do with a weird issue where the name of the variable was causing issues. I tested it locally and it seems to be fine now even when saving and loading. Will be making a new release 1.60.1 for this

PR #560, opened 2021-01-22
Detect an expired request exception

suthsc:
[JENKINS-63537](https://issues.jenkins-ci.org/browse/JENKINS-63537) - EC2SlaveMonitor terminates agent with a running job when STS session expires

* Add conditional logic to skip removal of nodes when RequestExpiredPredicate.test(AmazonClientException) returns true

This pull request is an alternative implementation to #493 as proposed @MRamonLeon. I am working on including tests for the new functionality.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2SlaveMonitor.java
+++ b/src/main/java/hudson/plugins/ec2/EC2SlaveMonitor.java
@@ -14,6 +14,9 @@ import hudson.plugins.ec2.util.MinimumInstanceChecker;
 import jenkins.model.Jenkins;
 
 import com.amazonaws.AmazonClientException;
+import com.amazonaws.services.ec2.model.AmazonEC2Exception;
+
+import static hudson.plugins.ec2.EC2Cloud.EC2_REQUEST_EXPIRED_ERROR_CODE;
 
 /**
  * @author Bruno Meneguello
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/ec2/EC2SlaveMonitor.java
```diff
@@ -51,8 +51,12 @@ private void removeDeadNodes() {
                         ec2Slave.terminate();
                     }
                 } catch (AmazonClientException e) {
-                    LOGGER.info("EC2 instance is dead and failed to terminate: " + ec2Slave.getInstanceId());
-                    removeNode(ec2Slave);
+                    if (new RequestExpiredPredicate().test(e)) {
```
res0nance, line 54: Maybe this is better as a private static instance

suthsc: @res0nance Thanks for taking a look. I've updated the source branch with the code we are using internally. I'm not crazy about a static instance either, but it's better than creating the instance every check. I do see I'm doing the same thing in the CloudHelper class. I'll move that to a static field as well for now.

suthsc:
Might be related to #515

res0nance (approved):

PR #570, opened 2021-02-19
Option to launch encrypted EBS root volume from unencrypted AMI

anitabee:
**What does PR do:**
This PR implements option to encrypt root EBS volume from unencrypted AMI

**Why:**
Our security policy requires EBS volumes to be encrypted, and since all public AMIs need to be unencrypted, in order for us to use them we would require this implementation.

Since 2019 it is possible to encrypt EBS backed EC2 instances from unencrypted AMIs; more about  it [here](https://aws.amazon.com/about-aws/whats-new/2019/05/launch-encrypted-ebs-backed-ec2-instances-from-unencrypted-amis-in-a-single-step/).

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -23,20 +23,16 @@ import com.amazonaws.AmazonServiceException;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.services.ec2.AmazonEC2;
 import com.amazonaws.services.ec2.model.AmazonEC2Exception;
-import com.amazonaws.services.ec2.model.AvailabilityZone;
 import com.amazonaws.services.ec2.model.BlockDeviceMapping;
 import com.amazonaws.services.ec2.model.CancelSpotInstanceRequestsRequest;
 import com.amazonaws.services.ec2.model.CreateTagsRequest;
 import com.amazonaws.services.ec2.model.CreditSpecificationRequest;
-import com.amazonaws.services.ec2.model.DescribeAvailabilityZonesResult;
 import com.amazonaws.services.ec2.model.DescribeImagesRequest;
 import com.amazonaws.services.ec2.model.DescribeInstancesRequest;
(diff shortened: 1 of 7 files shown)


anitabee:
@res0nance sorry for tagging you directly, let me know if there is any other preferred way to start the conversation over this?
In any case, do you think this change would be of interest to the rest of the community, and not sure if it's clear enough? Also, if you have any other ideas, I would be happy to hear them out.
Many thanks in advance

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1195,14 +1226,14 @@ private void setupRootDevice(Image image, List<BlockDeviceMapping> deviceMapping
...
-            BlockDeviceMapping newMapping = rootMapping.clone();
+            // pass deleteRootOnTermination to shadow of the AMI mapping
             newMapping.getEbs().setDeleteOnTermination(Boolean.TRUE);
-            //Per the documentation, "If you are creating a volume from a snapshot, you can't specify an encryption value. This is because only blank volumes can be encrypted on creation. "
-            //The root volume will always have a snapshot, so this value needs to be set to null to work correctly
-            newMapping.getEbs().setEncrypted(null);
-            deviceMappings.add(0, newMapping);
         }
+
+        newMapping.getEbs().setEncrypted(encryptEbsRootVolume);
```
res0nance, line 1233: This looks like it alters behaviour, if I did not specify it should do what the default was previously. Now if i save the configuration we get either true or false. If i specify false on an encrypted ami does it become unencrypted? This is likely to cause issues with certain users configurations

res0nance (changes requested):

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
anitabee: Thanks for review! 
I've updated it so it is setting default to null to make it backward compatible

res0nance (approved):

anitabee:
@res0nance many thanks for the merge! Are there any plans for next release?

PR #574, opened 2021-02-26
Don't fetch instance data on node creation

mtughan:
When a new node is provisioned from EC2, the instance data should be
lazily fetched in order to avoid generating network traffic in the
NodeProvisioner thread. All the getters for the attributes updated by
fetchLiveInstanceData call it first to update if need be, so there's no
need to be eager here.


res0nance (approved):
Looks reasonable

PR #582, opened 2021-03-17
Support Mac Metal ec2 instance

a-pikachu:
Support Mac Metal ec2 instance

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@ THE SOFTWARE.
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.222.4</jenkins.version>
         <java.level>8</java.level>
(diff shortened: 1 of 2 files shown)


froblesmartin:
@a-pikachu This does not really add support for it, it only specifies a default number of executors (which, by the way, seems to be too low in comparison with the rest). To add support for it you will probably need to update the aws-java-sdk.version in pom.xml.

a-pikachu:
by the time I create the PR, the master branch has bump the aws-java-sdk.version from 1.11.854 to 1.11.973. Mac1Metal support is added since 1.11.909

a-pikachu:
seems like I need to add dedicated host implementation. Closing this PR for now

PR #585, opened 2021-03-22
Adding Mac AMI Type support to launch mac1.metal instances by ec2-plugin

ykhandelwal913:
Adding following feature as part of the PR

- Adding support for macType in AmiType.
- Written separate classes for macSupport. In case in future, something specific needed for macOS. Currently it is like UnixLauncher but having java installation and url changes. Using aws corretto java.
- Written Testcases to validate new Type.
This PR has a dependency on https://github.com/jenkinsci/ec2-plugin/pull/492 as mac1.metal instances need HostTenancy.


ykhandelwal913:
@res0nance can you label it as enhancement.

res0nance:
I do not understand this change, the new launcher simply looks like a copy of UnixLauncher so why can't we use that instead?

ykhandelwal913:
@res0nance installation of java would be different in case of OSX.

ykhandelwal913:
I could use System.getProperty("os.name") and added check in unix launcher to solve it  but kept a new class and other files separate for mac, in case we want to add more functionality related to mac. let me know your thoughts, i can add system property changes and have another PR.

ykhandelwal913:
@res0nance In future if there is a drift having different path(linux and mac) is helpful

res0nance:
I think the change is ok but i lack the infrastructure to test this properly. Have you tested these changes?

ykhandelwal913:
yes i have validated this changes.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1738,6 +1741,10 @@ public boolean isUnixSlave() {
         return amiType.isUnix();
     }
 
+    public boolean isMacSlave() {
```
res0nance, line 1744: We should avoid the use of the world slave in new apis could this be amended to be 
```suggestion
    public boolean isMacAgent() {
```

taneishamitchell:
@res0nance who's responsible for version releases?
I am so looking forward to these tenancy changes to be available

res0nance:
> @res0nance who's responsible for version releases?
> I am so looking forward to these tenancy changes to be available

Technically any maintainer can release it

res0nance (approved):

taneishamitchell:
> Technically any maintainer can release it

Is there any hopes of that happening soon after the tests are passing on the `master` branch? A girl is desperate 😞

a-pikachu:
Is there any documentation on how to config this? Especially the dedicated host part. Do we need to first reserved a host or the plugin will do it for us? Thanks

ykhandelwal913:
This plugin won't reserve the host. You will have to reserve the host. This plugin can help you to launch the mac1.metal instance on the dedicated host.

a-pikachu:
in https://docs.aws.amazon.com/powershell/latest/reference/items/New-EC2Instance.html

I see this
New-EC2Instance -ImageId ami-1a2b3c4d -InstanceType m4.large -KeyName my-key-pair -SecurityGroupId sg-1a2b3c4d  -AvailabilityZone us-west-1a -Tenancy host -HostID h-1a2b3c4d5e6f1a2b3
Launches an instance into the specified dedicated host environment.

But with option in the plugin only have Tenancy choice, where to pass the host ID?

image

a-pikachu:
@ykhandelwal913 I am new to AWS so can you point me to where to turn on the instance auto-placement? Is it in the AWS panel? Thanks

ykhandelwal913:
@a-pikachu sorry for the delay in reply. Either when you are reserving the instance you can pass the parameter like `aws ec2 allocate-hosts --instance-family Mac1 --availability-zone us-east-1a --auto-placement on --quantity x` or change/create it from aws panel->EC2-> DedicatedHosts

a-pikachu:
@ykhandelwal913 thank you, I have it resolved now by enabling auto-placement.

PR #590, opened 2021-04-09
[JENKINS-64520][JENKINS-62340] Pass ssh host key to ssh process when using the strategy CHECK_NEW_HARD

nigelarmstrong-bc:
Pass ssh host key to ssh process when using the strategy CHECK_NEW_HARD. Without this when connecting via ssh process, the ssh process might not have the ssh host key. Take the ssh host key that jenkins already verified and pass that to the ssh process. 

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
@@ -30,6 +30,7 @@ import hudson.model.Descriptor;
 import hudson.model.TaskListener;
 import hudson.plugins.ec2.*;
 import hudson.plugins.ec2.ssh.verifiers.HostKey;
+import hudson.plugins.ec2.ssh.verifiers.HostKeyHelper;
 import hudson.plugins.ec2.ssh.verifiers.Messages;
 import hudson.remoting.Channel;
 import hudson.remoting.Channel.Listener;
@@ -44,6 +45,7 @@ import java.io.PrintStream;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
 import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -322,6 +333,36 @@ private File createIdentityKeyFile(EC2Computer computer) throws IOException {
...
+        HostKey ec2HostKey = HostKeyHelper.getInstance().getHostKey(computer);
+        if (ec2HostKey == null){
+            return null;
+        }
+        File tempFile = File.createTempFile("ec2_", "_known_hosts");
+        String knownHost = "";
+        knownHost = String.format("%s %s %s", ec2HostAddress, ec2HostKey.getAlgorithm(), Base64.getEncoder().encodeToString(ec2HostKey.getKey()));
+
+        try {
+            FileOutputStream fileOutputStream = new FileOutputStream(tempFile);
```
res0nance, line 346: This bit will probably just read better in a try with resources way
try (OutputStreamWriter writer = new OutputStreamWriter(new FileOutputStream(tempFile), StandardCharsets.UTF_8))

which will take care of closing it on an exception

res0nance (approved):
Looks ok untested

src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
nigelarmstrong-bc: I've given this a shot. My java isn't great, so would appreciate a double check of my changes.

nigelarmstrong-bc:
I tested this on our installation with `CHECK_NEW_HARD` and `CHECK_NEW_SOFT`, but I think it would be good to get more testing on other configurations.

res0nance (approved):

PR #602, opened 2021-04-23
Fill credentials from current context instead of just Jenkins singleton

Dohbedoh:
Proposing this as this is useful in CloudBees products where Cloud implementation can be configured within Folders within folders, then pushed connected Jenkins instance. Similar feature were added to for example the kubernetes plugin: https://github.com/jenkinsci/kubernetes-plugin/commit/00ca6a819d951ebf481b21be6750493594a9b8ce

I have added a test to validate that this does not impact the current functionalities.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -67,6 +67,7 @@ import hudson.util.StreamTaskListener;
 import jenkins.model.Jenkins;
 import jenkins.model.JenkinsLocationConfiguration;
 import org.apache.commons.lang.StringUtils;
+import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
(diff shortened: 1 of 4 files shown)


res0nance (approved):

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1153,12 +1154,12 @@ public FormValidation doCheckSshKeysCredentialsId(@QueryParameter String value)
          * @throws IOException
          * @throws ServletException
          */
-        protected FormValidation doTestConnection(URL ec2endpoint, boolean useInstanceProfileForCredentials, String credentialsId, String sshKeysCredentialsId, String roleArn, String roleSessionName, String region)
+        protected FormValidation doTestConnection(@AncestorInPath ItemGroup context, URL ec2endpoint, boolean useInstanceProfileForCredentials, String credentialsId, String sshKeysCredentialsId, String roleArn, String roleSessionName, String region)
```
MRamonLeon, line 1157: Keeping the former method, without the context, to keep the compatibility would be good. Some subclasses may rely on the old signature.

MRamonLeon: Since this is a method expected to be called from jellies, it shouldn't break anything.

PR #605, opened 2021-04-28
[JENKINS-64571] Fall back to on demand provisioning when spot quota exceeded and option set

TheImplementer:
Description of the issue available at: https://issues.jenkins.io/browse/JENKINS-64571

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -1367,8 +1367,18 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
                 spotRequest.setBlockDurationMinutes(getSpotBlockReservationDuration() * 60);
             }
 
-            // Make the request for a new Spot instance
-            RequestSpotInstancesResult reqResult = ec2.requestSpotInstances(spotRequest);
+            RequestSpotInstancesResult reqResult;
+            try {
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1367,8 +1367,18 @@ private void setupCustomDeviceMapping(List<BlockDeviceMapping> deviceMappings) {
                 spotRequest.setBlockDurationMinutes(getSpotBlockReservationDuration() * 60);
             }
 
-            // Make the request for a new Spot instance
-            RequestSpotInstancesResult reqResult = ec2.requestSpotInstances(spotRequest);
+            RequestSpotInstancesResult reqResult;
+            try {
+                // Make the request for a new Spot instance
+                reqResult = ec2.requestSpotInstances(spotRequest);
```
res0nance, line 1373: There seems to be some logic below for something similar could the logic be merged instead of duplicated?

TheImplementer: The logic is not really duplicated. In this case the client would just throw an exception, while in SlaveTemplate.java:1405 the logic is based on the status of spots requests that were created successfully. (hence the need to cancel them before falling back to on demand)

res0nance (approved):
Thanks for clearing things up

PR #612, opened 2021-05-06
[JENKINS-64738] Fixed reference to get the amiType class name

TheImplementer:
Turns out the typo fixed in [608](https://github.com/jenkinsci/ec2-plugin/pull/608) wasn't the only thing wrong. The `instance` var is only reassigned inside the `dropdownListBlock` element, so `instance.amiType` should be used in the `staplerClass` attribute instead.

The fix has been tested manually and now works as expected.


TheImplementer:
Anyone who can review this? @res0nance maybe? It would be good to have it in the next release.

res0nance (approved):

PR #617, opened 2021-05-22
Raise an exception when AMI search is blank.

bazzargh:
Changes the behaviour when the AMI ID and search are blank from logging a warning to raising an exception. Motivated by experiencing what happens when it just warns:

We use the configuration-as-code plugin; when interpolating secrets, if you have a typo in the secret name, it inserts the empty string:

https://github.com/jenkinsci/configuration-as-code-plugin/blob/e7501727c4704aaa1e95bdc246a1454dc5b20440/plugin/src/main/java/io/jenkins/plugins/casc/SecretSourceResolver.java#L124

Unfortunately we had a typo in the name of an AMI used by a rarely-run job, which used a slightly different AMI from all the other jobs on our server. So, when we deployed, we missed the warning message:

    Configuration import: Found unresolved variable '/plugins/amazonEC2/special-ami'. Will default to empty string

... then a day later, the job that used this special ami ran. Because the code warned rather than throwing an exeption, it attempted to search for an AMI with no search parameters, then picked one of the results to launch. As luck would have it, there was also a typo in our IAM rule that prevented launching public AMIs; eventually, after two days of retries, the job manage to launch one of the random AMIs: a bitcoin miner :facepalm:. The node was shut down seconds later when jenkins failed to connect to it, and our alarms noticed the miner, leading us to investigate where it had originated.

Hence this PR...I really don't think anyone wants to use Jenkins as a random AMI launcher :)

- [ x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [ x ] Ensure that the pull request title represents the desired changelog entry
- [ x ] Please describe what you did
- [ x ] Link to relevant issues in GitHub or Jira
- [ x ] Link to relevant pull requests, esp. upstream and downstream changes
- [ x ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

(there is no issue raised to link to)

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -1234,7 +1234,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
     }
 
     @NonNull
-    private DescribeImagesRequest makeDescribeImagesRequest() {
+    private DescribeImagesRequest makeDescribeImagesRequest() throws AmazonClientException {
         List<String> imageIds = Util.fixEmptyAndTrim(ami) == null ?
             Collections.emptyList() :
             Collections.singletonList(ami);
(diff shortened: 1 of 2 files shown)


res0nance (approved):
Looks reasonable, ~does the UI warn of this case?~
I checked and it doesn't. Probably should add a UI warning here

PR #621, opened 2021-05-24
[JENKINS-64569] Add support for EBS Throughput parameter

Panfilwk:
Closes https://issues.jenkins.io/browse/JENKINS-64569 by adding the `throughput` param to the end of the parser for the colon separated string. Based on other existing tests which have fewer separators than options, this should be entirely backwards compatible with existing block device config strings.

I've written a simple test, but I'm having issues getting the `util` tests to run both locally and in CI. It seems like maybe they aren't getting picked up by the Maven project?

### DIFF
--- a/src/main/java/hudson/plugins/ec2/util/DeviceMappingParser.java
+++ b/src/main/java/hudson/plugins/ec2/util/DeviceMappingParser.java
@@ -60,7 +60,7 @@ public class DeviceMappingParser {
         return deviceMappings;
     }
 
-    // [<snapshot-id>]:[<size>]:[<delete-on-termination>]:[<type>]:[<iops>]:[encrypted]
+    // [<snapshot-id>]:[<size>]:[<delete-on-termination>]:[<type>]:[<iops>]:[encrypted]:[throughput]
     private static EbsBlockDevice parseEbs(String blockDevice) {
 
         String[] parts = blockDevice.split(":");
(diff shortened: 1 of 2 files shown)


res0nance:
Hey I noticed the tests were not getting run in CI and enabled them but accidentally broke this PR
I Enabled them in #625 
Could you rebase it?

Panfilwk:
Thanks @res0nance! Just rebased and all the device mapping tests (including the new one) passed locally, so we'll see what happens in CI :)

One thing I was lightly concerned about is that the `throughput` param in EC2 is only supported for `gp3` volume types. It seems like the prior art here with provisioned iops is to let the EC2 SDK handle the error when a volume type doesn't support a parameter, but I just wanted to make sure that's acceptable here also.

res0nance (approved):

Panfilwk:
Is there any action left on my end for this to get merged? This feature is mostly a nice-to-have for us, so I'm not in an enormous rush, but also don't want to be blocking :)

res0nance:
> Is there any action left on my end for this to get merged? This feature is mostly a nice-to-have for us, so I'm not in an enormous rush, but also don't want to be blocking :)

No, I was waiting for CI to finish

Panfilwk:
Awesome, thanks again for your help with this!

julian-alarcon:
There is a missing change in the documentation https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-customDeviceMapping.html#L32

Also, an example with encrypted disabled could be useful. Currently I'm not sure if it should be `nonencrypted` , `non-encrypted`, `noencrypted`, etc.

julian-alarcon:
Mmm, it seems that I just need to set the encrypted section empty to disable encryption according to this code.
https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/java/hudson/plugins/ec2/util/DeviceMappingParser.java#L85

So, `/dev/xvda=:900::gp3:::3000` should set a size of 900, gp3, and 3000 of the throughput value.

PR #622, opened 2021-05-27
Adds new delimeters for subnet configuration: white spaces, comma, semicolon can be used

TalAntR:
It looks reasonable to use more symbols to divide VPC subnets in Jenkins agent configuration. It can be used when you want to configure EC2 plugin through CASC (configuration-as-a-code) and perform evaluation some values from AWS SSM Parameter Store. In this case list of VPC subnets is provided as a comma separated values in a string.

I've added a regular expression that includes new symbols to split string of VPC subnet ids into array of its values. 

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -135,6 +135,8 @@ import java.util.stream.Stream;
 public class SlaveTemplate implements Describable<SlaveTemplate> {
     private static final Logger LOGGER = Logger.getLogger(SlaveTemplate.class.getName());
 
+    private static final String EC2_RESOURCE_ID_DELIMETERS = "[\\s,;]+";
+
     public String ami;
 
     public final String description;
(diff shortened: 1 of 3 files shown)


res0nance:
Shouldn't security groups also receive the same treatment?

TalAntR:
> Shouldn't security groups also receive the same treatment?

@res0nance 

Yes, it's reasonable as well. I thought about this but security groups can be defined in the configuration as names (not id) and all of suggested symbols [are permitted](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_SecurityGroups.html#VPCSecurityGroups) to use. So I'm not sure that it's right choice.

res0nance:
Ahh ok, could the corresponding help also be updated? https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-subnetId.html

res0nance:
Also could you provide tests?

TalAntR:
> Also could you provide tests?

@res0nance Updated. Is it ok now?

res0nance (approved):

PR #632, opened 2021-06-14
[JENKINS-23792] Do not terminate stopped EC2 nodes that were just booted

unicolet:
when a node just booted it is not connected, and simultaneously its idle start time could still be set to the previous time the node was booted. If the instance is checked by the plugin while in this phase, the plugin will erroneously decide to terminate it again. The change in this PR adds a safety check to ensure that idle time is calculated from the greatest of idle time as returned by jenkins or the node uptime, as returned by EC2.

Fixes: https://issues.jenkins.io/browse/JENKINS-23792

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Computer.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Computer.java
@@ -106,7 +106,7 @@ public class EC2Computer extends SlaveComputer {
         GetConsoleOutputRequest request = new GetConsoleOutputRequest(getInstanceId());
         return ec2.getConsoleOutput(request).getDecodedOutput();
     }
-    
+
     /**
      * Obtains the instance state description in EC2.
      *
@@ -152,6 +152,15 @@ public class EC2Computer extends SlaveComputer {
      */
     public String getUptimeString() throws AmazonClientException, InterruptedException {
         return Util.getTimeSpanString(getUptime());
+    }
+
+    /**
+     * Return the time this instance was launched in ms since the epoch.
+     *
+     * @return Time this instance was launched, in ms since the epoch.
+     */
+    public long launchedAtMs() throws InterruptedException {
+        return this.describeInstance().getLaunchTime().getTime();
     }
 
     /**
--- a/src/main/java/hudson/plugins/ec2/EC2Computer.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Computer.java
@@ -159,7 +159,7 @@ public class EC2Computer extends SlaveComputer {
      *
      * @return Time this instance was launched, in ms since the epoch.
      */
-    public long launchedAtMs() throws InterruptedException {
+    public long getLaunchTime() throws InterruptedException {
         return this.describeInstance().getLaunchTime().getTime();
     }
 
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -146,7 +146,7 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
             try {
                 state = computer.getState(); //Get State before Uptime because getState will refresh the cached EC2 info
                 uptime = computer.getUptime();
-                launchedAtMs = computer.launchedAtMs();
+                launchedAtMs = computer.getLaunchTime();
             } catch (AmazonClientException | InterruptedException e) {
                 // We'll just retry next time we test for idleness.
                 LOGGER.fine("Exception while checking host uptime for " + computer.getName()
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -140,11 +140,13 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
 
         if (computer.isIdle() && !DISABLED) {
             final long uptime;
+            final long launchedAtMs;
             InstanceState state;
 
             try {
                 state = computer.getState(); //Get State before Uptime because getState will refresh the cached EC2 info
                 uptime = computer.getUptime();
+                launchedAtMs = computer.launchedAtMs();
             } catch (AmazonClientException | InterruptedException e) {
                 // We'll just retry next time we test for idleness.
                 LOGGER.fine("Exception while checking host uptime for " + computer.getName()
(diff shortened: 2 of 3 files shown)


src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
```diff
@@ -86,6 +86,7 @@ private EC2Computer computerWithIdleTime(final int minutes, final int seconds) t
      * the computer returns the value established.
      */
     private EC2Computer computerWithIdleTime(final int minutes, final int seconds, final Boolean isOffline, final Boolean isConnecting) throws Exception {
+        idleTimeoutCalled.set(false);
```
unicolet, line 89: required, since the state is otherwise never reset across tests. Admittedly, this could be fixed better, but I wanted to keep the PR scope as small as possible to make it easy to review :-)

unicolet:
@res0nance @MRamonLeon could you please take a look? This issue is affecting our jenkins setup a lot :-( I'd be more than happy to timely clarify any question you'll sen our way. Thanks!

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -190,7 +190,7 @@ private long internalCheck(EC2Computer computer) {
                 }
             }
 
-            final long idleMilliseconds = this.clock.millis() - computer.getIdleStartMilliseconds();
+            final long idleMilliseconds = this.clock.millis() - Math.max(computer.getIdleStartMilliseconds(), uptime);
```
res0nance, line 193: So this looks wrong because,
https://github.com/jenkinsci/ec2-plugin/blob/8867fe65acaa999c52dc4689192a8cd5da136234/src/main/java/hudson/plugins/ec2/EC2Computer.java#L146-L148

uptime is `Currtime - instance_launch_time`
idlestartmiliseconds is `instance_idle_start_time`

so wouldnt this be `Currtime - (currtime - instance_launch_time) = instance_launch_time`?

Would the right solution here not be

`this.clock.millis() - Math.max(computer.getIdleStartMilliseconds(), computer.describeInstance().getLaunchTime().getTime());`

instead so we get the time the instance launched?

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
```diff
@@ -221,6 +222,20 @@ public void testTerminateOfflineComputerIfNotConnecting() throws Exception {
         rs.check(computer2);
         assertThat("The computer is terminated, it should not accept more tasks", idleTimeoutCalled.get(), equalTo(true));
         assertThat(logging.getMessages(), hasItem(containsString("offline but not connecting, will check if it should be terminated because of the idle time configured")));
+
+        // A computer that just booted should not be terminated. Let's wait
+        // until up to termination minutes before terminating it
+        final Instant inOneMinute = Instant.now().plus(Duration.ofSeconds(60));
```
res0nance, line 228: Could this be moved into a separate test?

res0nance (commented):
From my understanding this happens only if the stop option is used?

unicolet:
> From my understanding this happens only if the stop option is used?

@res0nance yes, that's my understanding too

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
unicolet: ah right, well spotted! hold on

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
res0nance: This test is probably not right as well given that the codefix was not correct

unicolet:
@res0nance please take a look again. Thanks for the prompt reply :heart:

src/main/java/hudson/plugins/ec2/EC2Computer.java
```diff
@@ -154,6 +154,15 @@ public String getUptimeString() throws AmazonClientException, InterruptedExcepti
         return Util.getTimeSpanString(getUptime());
     }
 
+    /**
+     * Return the time this instance was launched in ms since the epoch.
+     *
+     * @return Time this instance was launched, in ms since the epoch.
+     */
+    public long launchedAtMs() throws InterruptedException {
```
res0nance, line 162: getLaunchTime() ?

Then getUptime can use getLaunchTime() instead

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
```diff
@@ -238,6 +243,28 @@ public void testTerminateOfflineComputerIfNotConnecting() throws Exception {
...
 
+
+    /**
+     * Do not terminate na instance if a computer just launched, and the
+     * retention strategy timeout has not expired yet. The idle timeout
+     * is correctly accounted for nodes that have been stopped and started
+     * again (i.e termination policy is stop, rather than terminate).
+     */
+    @Test
+    public void testDoNotTerminateInstancesJustBooted() throws Exception {
```
res0nance, line 240: This test is insufficient, it passes with or without this patch

res0nance (changes requested):

src/main/java/hudson/plugins/ec2/EC2Computer.java
unicolet: fixed

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
unicolet: yeah, let me see what I can do about it

unicolet: we should be good now, thanks!

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
```diff
@@ -215,18 +220,80 @@ public void testTerminateOfflineComputerIfNotConnecting() throws Exception {
...
+        final int COMPUTER_UPTIME_MINUTES=5;
+        final int RETENTION_MINUTES=5;
+        final int CHECK_TIME_MINUTES=COMPUTER_UPTIME_MINUTES+RETENTION_MINUTES+1;
+        final Instant checkTime = Instant.now().plus(Duration.ofMinutes(CHECK_TIME_MINUTES));
+        new EC2RetentionStrategy(
+            String.format("%d", RETENTION_MINUTES),
+            Clock.fixed(checkTime.plusSeconds(1), zoneId),
+            checkTime.toEpochMilli()
+        ).check(computerWithUpTime(-COMPUTER_UPTIME_MINUTES, 0, null, false));
+        assertThat("The computer is terminated, but should not be", idleTimeoutCalled.get(), equalTo(true));
```
res0nance, line 289: Computer should be terminated in this case yes?

res0nance (approved):

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
unicolet: thanks, fixed :-)

res0nance (approved):

unicolet:
@res0nance thanks for the review :heart: How does the process for merging and releasing this plugin looks like? in other words, when will the plugin be avail for upgrade? :sweat_smile:

PR #650, opened 2021-09-14
Add Security Group and Subnet Filters to Counting

pyieh:
When counting the number of existing instances to determine if there's capacity to create new instances for a specific slave template, there is a bug where the EC2 plugin would count EC2 instances regardless of different subnet and/or with a different security group than the ones configured for the slave template. These instances are incorrectly counted because having a existing instances with a different subnet and/or security group can have different connectivity and network permissions than desired by the slave template's configs. Adding the security group and subnet IDs filters gives us a more accurate number of existing instances that actually match all the desired configurations of the corresponding slave template.

Note that this doesn't solve incorrect counting of ALL instances (there's a separate count for all instances vs instances that match a given slave template).

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -646,6 +646,18 @@ public abstract class EC2Cloud extends Cloud {
                     }
                 }
             }
+            // Add Security group filters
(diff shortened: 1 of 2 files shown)


mwebber:
This might also fix https://issues.jenkins.io/browse/JENKINS-75739

PR #658, opened 2021-09-28
Logged info whenever falling back to on-demand instance

arunmastermind:
Logged info whenever falling back to on-demand instance as this is very useful to track the logs

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -889,6 +889,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
                 return provisionSpot(image, number, provisionOptions);
             return Collections.emptyList();
         }
+        LOGGER.info("No Configuration Spot, falling back to on-demand instance.");
         return provisionOndemand(image, number, provisionOptions);
     }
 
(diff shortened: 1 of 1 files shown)


arunmastermind:
Logged info whenever falling back to on-demand instance as i found this missing and there is no way to find out the reason for on demand instances.
we must be very focused on providing the logs as the end user will always judge the product as per these features.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1352,6 +1353,7 @@ private void setupCustomDeviceMapping(List<BlockDeviceMapping> deviceMappings) {
     private List<EC2AbstractSlave> provisionSpot(Image image, int number, EnumSet<ProvisionOptions> provisionOptions)
             throws IOException {
         if (!spotConfig.useBidPrice) {
+            LOGGER.info("No Bidding for Spot, falling back to on-demand instance.");
```
res0nance, line 1356: This actually creates a spot instance

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -889,6 +889,7 @@ public Tenancy getTenancyAttribute() {
                 return provisionSpot(image, number, provisionOptions);
             return Collections.emptyList();
         }
+        LOGGER.info("No Configuration Spot, falling back to on-demand instance.");
```
res0nance, line 892: I see no point to this, this is clearly configured without spot

PR #676, opened 2021-10-21
[JENKINS-60292] Close `com.trilead.ssh2.Connection` used during setup in `connectBySSHProcess` mode

jglick:
May fix [JENKINS-60292](https://issues.jenkins.io/browse/JENKINS-60292). Untested!

#222 looks like the nicer fix but that is a far bigger change.


Dohbedoh:
I have an environment setup to test this, where I can reproduce the leak easily. Will provide feedback shortly.

jglick:
@res0nance?

PR #716, opened 2022-03-04
Add optional bootDelay parameter for SSH slaves

djesionek:
This PR contains changes from #254 by @varju updated to be mergeable with the current master version of this plugin repository.

The PR changes add an optional "Boot Delay" option for unix and mac data configurations, similar as it is already present for WinRM based windows connections.

My use case was to use SSH connection with windows agents and WinRM is simply not performant enough. With the current state of unix ssh connection I ran into this situation which could be resolved by this change:
https://forums.aws.amazon.com/thread.jspa?messageID=380785&#380785 
https://stackoverflow.com/questions/12369090/ec2-windows-instances-initializing-twice

I'd be still missing more elaborate tests which demonstrate the delay is actually working to comply with the PR template. But since I wasn't able to find any for windows I left it at adding this change to respective CasC tests for now.


chrisipa:
Really important code change. Thanks for your work @djesionek.

varju:
I'm not personally working much in Jenkins these days, but I know my company continues to maintain a private fork of this plugin with the bootDelay change. I'm really excited to see somebody else looking into this, and hope the feature can finally get merged into the main release.

src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java
```diff
@@ -151,6 +151,10 @@ protected void launchScript(EC2Computer computer, TaskListener listener) throws
         try {
             boolean isBootstrapped = bootstrap(computer, listener, template);
             if (isBootstrapped) {
+                logInfo(computer, listener, "SSH service responded. Waiting for service to stabilize");
```
res0nance, line 154: Can we add the number of seconds its sleeping here in the log? And only print these messages if bootdelay is > 0 else a 0 number sees these log lines spammed

djesionek: Makes sense, however in that case this should be done in all three places

djesionek:
Looks like the failure is due to a false positive in SpotBugs. Any suggestions?

res0nance:
> Looks like the failure is due to a false positive in SpotBugs. Any suggestions?

computer.getNode() can return null so its not a false positive just do a null check and it should work

djesionek:
computer.getNode() == false is in an early fail condition prior to this usage but the variable node really should be reused. SpotBugs does recognize this properly now.

res0nance (approved):

PR #748, opened 2022-06-29
[Documentation][Security] Enforce IMDSv2 by default & update groovy section to use up-to-date constructor

jdelnano:
# Background
When recently setting up the EC2 plugin, I noticed the Groovy script detailed in the README used a deprecated `SlaveTemplate` constructor. The [current link](https://github.com/jenkinsci/ec2-plugin/blob/ec2-1.38/src/main/java/hudson/plugins/ec2/SlaveTemplate.java) in the comment above the `SlaveTemplate` object instantiation points to an older version of the EC2 Plugin and doesn't include the ability to specify various additional settings such as `metadataEndpointEnabled`, `metadataTokensRequired `, `metadataHopsLimit `, `hostKeyVerificationStrategy`, etc.

# What does this PR do?

- This PR updates the Groovy script section of the README, bringing it up-to-date. The section now reflects the setting of additional properties for `SlaveTemplate` objects and updates the associated comment link to the `SlaveTemplate`'s javadoc webpage.

- Additionally, I've updated the `metadataTokensRequired` field to be **true** by default, enforcing the usage of IMDSv2 over IMDSv1 out-of-the-box when configuring an AMI. Explained [here](https://github.com/SummitRoute/imdsv2_wall_of_shame)  on [Scott Piper's](https://www.linkedin.com/in/scott-piper-security/) (an industry-leading AWS security expert) "wall of shame" repository, IMDSv2 should be used exclusively (over IMDSv1) as an AWS security best practice. [IMDSv2 was released in](https://github.com/SummitRoute/imdsv2_wall_of_shame#what-is-imdsv2) response to the 2019 Capital One security breach, so I believe it is imperative that:
  - the field to enforce IMDSv2 usage be turned on by default
  - the README be updated to allow users to configure the plugin via groovy

Both of the above bullets are accomplished in this particular PR.

I've ensured that these changes run without issues, producing the updated Jenkins agent configuration in a configured cloud. Please see the short video below for a visual confirmation of the settings defined in the README.

https://user-images.githubusercontent.com/18095335/176363013-310e6ce1-42da-4bbb-8755-9aee2fe8fe43.mov

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
[...]

PR #776, opened 2022-09-08
Ensure that IMDSv2 Configuration Option is set to `true` by default

jdelnano:
# Background
This PR properly builds upon part of the aim of https://github.com/jenkinsci/ec2-plugin/pull/748. I had had an issue (my issue, unrelated to the plugin) locally building and installing the plugin to test that the default value of `metadataTokensRequired` was set to true by default. After the release of `2.x.x` of the plugin was released, I upgraded the EC2 plugin on my own Jenkins controller and saw that value of still false by default.

_This screencast shows the current behavior, using ec2 plugin 2.0.2_
https://user-images.githubusercontent.com/18095335/189275546-3d3895d2-ffdd-47af-b773-3377aeec49bf.mov

- As a result, I'm creating this follow up PR, which I have been able to successfully build, install on a Jenkins controller, and verify that `metadataTokensRequired` is checked by default (seen screencast below).
- I've also updated the tests accordingly and added a test to capture the use case of someone wanting to opt out of IMDSv2, using IMDSv1.

_This screencast shows what will be the new behavior, using a locally generated ec2 plugin with the PR's changes_
https://user-images.githubusercontent.com/18095335/189274511-7803f5cc-fa99-4b56-ab3d-31174ed41593.mov

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -77,7 +77,7 @@ import com.amazonaws.services.ec2.model.TerminateInstancesRequest;
 @SuppressWarnings("serial")
 public abstract class EC2AbstractSlave extends Slave {
     public static final Boolean DEFAULT_METADATA_ENDPOINT_ENABLED = Boolean.TRUE;
-    public static final Boolean DEFAULT_METADATA_TOKENS_REQUIRED = Boolean.FALSE;
+    public static final Boolean DEFAULT_METADATA_TOKENS_REQUIRED = Boolean.TRUE;
     public static final Integer DEFAULT_METADATA_HOPS_LIMIT = 1;
     public static final String DEFAULT_JAVA_PATH = "java";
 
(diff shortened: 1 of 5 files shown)


jdelnano:
@res0nance sorry for the direct tag, but flagging this since my previous PR missed some functionality to actually make IMDSv2 the default behavior. My apologies for not having fully verified the plugin locally the first time around!

res0nance (approved):

0x91:
I don't disagree with the intention of this PR but this is absolutely a breaking change. Any Jobs running on EC2 agents that do not support IMDSv2 and require AWS credentials will fail (as we just found out).

@res0nance can we make this clear in the release notes? It's not a bugfix either, it's a security enhancement.

jdelnano:
> I don't disagree with the intention of this PR but this is absolutely a breaking change. Any Jobs running on EC2 agents that do not support IMDSv2 and require AWS credentials will fail (as we just found out).
> 
> @res0nance can we make this clear in the release notes? It's not a bugfix either, it's a security enhancement.

So this was a follow up change to https://github.com/jenkinsci/ec2-plugin/pull/748 — which did prompt the release of major version 2.x.x — to actually have IMDSv2 take effect (by default). If folks had read the release notes for major version 2, they would’ve been prepared for the breaking behavior (although they didn’t take effect until this most recent release 2.0.3), which was correctly identified as a Security Enhancement. I think classifying this follow up change as a “bug fix” makes more sense than labeling it as, say, a feature since it does properly fix functionality that was promised in a previous pull-request.

PR #766, opened 2022-08-17
[JENKINS-69304] Allow specifying a java path

lemeurherve:
Cf https://issues.jenkins.io/browse/JENKINS-69304 and https://github.com/jenkins-infra/helpdesk/issues/3090, this PR add a "javaPath" option.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #777, opened 2022-09-19
JENKINS-69581: Install java-11-openjdk

grogersxyz:
If Jenkins recieves a non-zero error code when runing `java -fullversion` on the remote agent then it will install java 11 using yum.

Previously this installed java 8 which is no longer supported on Jenkins agents where the controller is running version 2.361.x or higher.

This issue is discussed further in [JENKINS-69581](https://issues.jenkins.io/browse/JENKINS-69581)

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
@@ -236,7 +236,7 @@ public class EC2UnixLauncher extends EC2ComputerLauncher {
 
             // TODO: parse the version number. maven-enforcer-plugin might help
             final String javaPath = node.javaPath;
-            executeRemote(computer, conn, javaPath + " -fullversion", "sudo yum install -y java-1.8.0-openjdk.x86_64", logger, listener);
+            executeRemote(computer, conn, javaPath + " -fullversion", "sudo amazon-linux-extras install java-openjdk11 -y; sudo yum install -y fontconfig java-11-openjdk", logger, listener);
             executeRemote(computer, conn, "which scp", "sudo yum install -y openssh-clients", logger, listener);
 
             // Always copy so we get the most recent remoting.jar


basil (commented):
Thanks for the PR! How was this PR tested? [This page](https://tecadmin.net/install-java-on-amazon-linux/) states that the command to install Java 11 on Amazon Linux is `sudo amazon-linux-extras install java-openjdk11`, which is the same command we use in the [Jenkins packaging test suite](https://github.com/jenkinsci/packaging/blob/e902bf931577c88d61bcf4287482bf3cefa384ac/molecule/default/install-rpm.yml). There, the (tested) algorithm is as follows:

- On Amazon Linux, check for `/usr/lib/jvm/jre-11-openjdk`; if it does not exist, run `amazon-linux-extras install java-openjdk11 -y`.
- On all RPM-based distributions (including Amazon Linux), run `yum install -y fontconfig java-11-openjdk` (on Amazon Linux, the `java11-openjdk` portion will be a no-op because of the previous command, while on other RPM-based distributions that will be the place where Java 11 is installed).

I am not the maintainer of this plugin, but I would expect this needs to work out of the box for the default EC2 Linux image at the very least, if not other common RPM-based distributions.

grogersxyz:
Of course yeap! I can work on that now and will run it through on a staging server I have. If we need unit tests for this I might need some help with that, I wouldn't say I'm especially familiar with Java :)

basil:
Although I am not the maintainer of this plugin, I think the primary concern is avoiding regressions for the 11,802 users of this plugin rather than achieving a perfect work of software engineering. That is why my comments were mostly centered around testing. We are happy to have your contribution regardless of your Java experience. Regarding testing, I think there are two main use cases we need to think about:

- A user who is already using an image with Java 11: either as the default Java (returned by `java -version`) or in a non-default path specified by `javaPath` as in https://github.com/jenkinsci/ec2-plugin/pull/766): in this use case, we should not install anything.
- A user who is using e.g. a default EC2 Amazon Linux image (or possibly some other RPM-based distribution) without Java 11 (or any Java at all) installed by default: in this use case, we want to install a working Java runtime as described in my previous post.

grogersxyz:
I gave it a go trying to do it in Java but didn't really know enough about input and output streams and buffers etc so I'll just leave what I did here sorry. I imagine someone will want to add this support in the future if they use the feature and upgrade to a version of jenkins that doesn't support Java8.

If I were able to just do it in bash I'd do something like `[ $(java -fullversion 2>&1 | cut -d'"' -f2 | sed 's/^1\.//' | cut -d'.' -f1) == "11" ] 2>/dev/null`

basil:
This PR was a promising start but lacks sufficient testing and cannot be merged as-is. If nobody else picks up this work I would suggest closing this PR.

nfplatzke:
When is this going to be fixed? This plugin is currently broken.

basil:
When someone submits an acceptable pull request that fixes it, I suppose.

grogersxyz:
I might have some time this weekend to take another look at this. If I’m not able to get it working and test it I’ll close this PR as suggested but keep [JENKINS-69581](https://issues.jenkins.io/browse/JENKINS-69581) open for further discussion.

grogersxyz:
Here's the config that I used to test this build on Jenkins version 2.361.4:
- [config.txt](https://github.com/jenkinsci/ec2-plugin/files/10049021/config.txt)

Here's the console outputs of my testing:
- [amazon-linux2.txt](https://github.com/jenkinsci/ec2-plugin/files/10049012/amazon-linux2.txt)
- [rhel-9.txt](https://github.com/jenkinsci/ec2-plugin/files/10049013/rhel-9.txt)

Given that this is my first contribution to this repository I might suggest someone else verifying that this is working for them. Happy to make any modifications as needed!

As I mentioned in my commit it would be cool to check the version of Java installed, but at this point I wanted to focus on fixing this regression for current users more than implementing a new feature.

grogersxyz:
In the meantime if anyone finds this you can resolve this issue temporarily by adding the following commands to your init scripts:

```bash
# For Amazon Linux 2
sudo amazon-linux-extras install java-openjdk11 -y
# For RedHat Enterprise Linux
sudo yum install -y fontconfig java-11-openjdk
# For Ubuntu
sudo apt-get install openjdk-11-jdk
```

cc @nfplatzke

grogersxyz:
Just force pushed a change to the commit. I had previously been joining the commands together with `||` but that would lead to the second command not running in the instance the first command was a success (for example in amazon-linux2). I've now replaced this with a `;` so they're treated as two seperate commands and will both run regardless of the others return code.

basil (approved):

grogersxyz:
@thoulen tagging you in this PR as a maintainer, sorry I don't have permission to add reviewers as suggested in the contributing docs. This PR also fixes https://issues.jenkins.io/browse/JENKINS-68946 and https://issues.jenkins.io/browse/JENKINS-69383

basil:
Perhaps @res0nance would be interested in merging this as he was the last active maintainer. If not, my apologies for the obtrusive notification.

res0nance (approved):

PR #825, opened 2023-03-30
Fixes `Jenkins.SYSTEM_READ` support

Vlatombe:
This fixes readonly configuration support by changing multiple checks to `Jenkins.ADMINISTER` to return an empty validation if permission is missing.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
@@ -29,7 +29,6 @@ import hudson.Util;
 import hudson.model.Failure;
 import hudson.model.ItemGroup;
 import hudson.plugins.ec2.util.AmazonEC2Factory;
-import hudson.slaves.Cloud;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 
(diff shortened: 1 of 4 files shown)


daniel-beck:
It doesn't really benefit anyone to show errors to low-privilege users for use cases that are unexpected, so a simpler `if (!hasPermission(ADMINISTER)) return FormValidation.ok()` replacing the current `checkPermission` call should be enough. This PR as is should resolve the specific issue encountered too, of course.

res0nance (approved):

mwebber:
Reported as [JENKINS-70987](https://issues.jenkins.io/browse/JENKINS-70987)

I am running Jenkins 2.387.1 (latest LTS). When I upgraded the `ec2-plugin` from 2.0.6 to 2.0.7, I found that all Agents terminated after 10 minutes, which is our **Idle termination time**, even if a job was running on the Agent. In all cases, Agents were terminated 10 minutes after starting.

When we reverted back to 2.0.6, the problem went away, so this looks like a bug introduced in 2.0.7.
There were 2 PRs in that release, this one #825, and #827.

mwebber:
In response to the problem I reported in the previous comment (Agents terminating prematurely), deleting and recreating the cloud definition fixes the problem.
See https://github.com/jenkinsci/ec2-plugin/pull/827#issuecomment-1535391616

PR #827, opened 2023-03-31
Remove cloudId and use the name entered by the user directly.

Vlatombe:
This makes the plugin compatible with cloud copy feature that is being introduced in Jenkins core, cf. https://github.com/jenkinsci/jenkins/pull/7658

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
@@ -29,7 +29,6 @@ import hudson.Util;
 import hudson.model.Failure;
 import hudson.model.ItemGroup;
 import hudson.plugins.ec2.util.AmazonEC2Factory;
-import hudson.slaves.Cloud;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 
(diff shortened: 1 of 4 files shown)


Vlatombe:
JENKINS-15081 seems to predate the ability to provide a custom name for the EC2Cloud, which was the problem in their case (the cloud name was only computed based on the region, so if you had 2 EC2Cloud instances targetting the same region, they would mix up because of using the same name)

res0nance (approved):

mwebber:
@Vlatombe , I am running Jenkins 2.387.1 (latest LTS). When I upgraded the `ec2-plugin` from 2.0.6 to 2.0.7, I found that all Agents terminated after 10 minutes, which is the **Idle termination time**, even if a job was running on the Agent. In all cases, Agents were terminated 10 minutes after starting.

When we reverted back to 2.0.6, the problem went away, so this looks like a bug introduced in 2.0.7.
There were 2 PRs in that release, this one #827, and #825. 

The Jenkins job log shows this:
`Idle timeout of EC2 (CI-PROD) - Publisher v20230323 (i-0a844c202e5d3ac3d) after 11 idle minutes, instance statusRUNNING
Apr 04, 2023 3:24:40 PM INFO hudson.plugins.ec2.EC2AbstractSlave idleTimeout
EC2 instance idle time expired: i-0a844c202e5d3ac3d
Apr 04, 2023 3:24:40 PM INFO hudson.plugins.ec2.EC2AbstractSlave stop
EC2 instance stop request sent for i-0a844c202e5d3ac3d
Apr 04, 2023 3:24:40 PM INFO hudson.plugins.ec2.EC2RetentionStrategy internalCheck`

res0nance:
> Reported as [JENKINS-70987](https://issues.jenkins.io/browse/JENKINS-70987)
> 
> @Vlatombe , I am running Jenkins 2.387.1 (latest LTS). When I upgraded the `ec2-plugin` from 2.0.6 to 2.0.7, I found that all Agents terminated after 10 minutes, which is our **Idle termination time**, even if a job was running on the Agent. In all cases, Agents were terminated 10 minutes after starting.
> 
> When we reverted back to 2.0.6, the problem went away, so this looks like a bug introduced in 2.0.7. There were 2 PRs in that release, this one #827, and #825.
> 
> The Jenkins job log shows this:
> 
> ```
> Apr 04, 2023 3:24:40 PM INFO hudson.plugins.ec2.EC2AbstractSlave idleTimeout
> EC2 instance idle time expired: i-0a844c202e5d3ac3d
> Apr 04, 2023 3:24:40 PM INFO hudson.plugins.ec2.EC2AbstractSlave stop
> EC2 instance stop request sent for i-0a844c202e5d3ac3d
> Apr 04, 2023 3:24:40 PM INFO hudson.plugins.ec2.EC2RetentionStrategy internalCheck
> ```

I'm guessing this is the PR that causes the issue. After an upgrade does removing and re-adding the cloud resolve the issue?

mwebber:
@res0nance 
> After an upgrade does removing and re-adding the cloud resolve the issue?

Yes, that did fix the problem. Specifically, I did the following:

- Delete all Agent instances
- Delete all Cloud definitions
- Update the `ec2-plugin` from 2.0.6 to 2.0.7 (and restart controller)
- Recreate the Cloud and Agent definitions

basil:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertNotNull(Assert.java:713)
	at org.junit.Assert.assertNotNull(Assert.java:723)
	at io.jenkins.plugins.casc.EC2CloudTest.configure_ec2_cloud(EC2CloudTest.java:35)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:607)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

PR #840, opened 2023-04-13
Update AWS SDK to support M7g* EC2 instances

silverjam:
Updates Amazon SDK to allow new Graviton3 build agensts to be created with M7gXlarge (and related) constants.  Example: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/ec2/model/InstanceType.html#M7gXlarge.

Jira: https://issues.jenkins.io/browse/JENKINS-71059

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,14 +116,14 @@ THE SOFTWARE.
             <artifactId>bouncycastle-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.plugins.aws-java-sdk</groupId>
+            <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-ec2</artifactId>
-            <version>1.12.406-374.v4cdf53953691</version>
+            <version>1.12.447</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


mawinter69 (changes requested):
You're changing the dependency from a jenkins plugin to a plain java library. That way you will package those jars directly in this plugin. There are other plugins that need those libraries as well and by wrapping them in a plugin you can ensure that you have only a single version of those libraries in Jenkins and not 5 different version in each plugin that needs it.
So you need to change the jenkins plugins that wrap the corresponding aws libraries and not the consumers like the ec2 plugin
That will most likely also fix the build issues

silverjam:
Alright, I think that means I'm waiting on https://github.com/jenkinsci/aws-java-sdk-plugin/pull/975

silverjam:
Ping @mawinter69 @Vlatombe -- I think this is ready for review.  Thank you!

Vlatombe:
Updating the aws java sdk dependency is not needed, unless the plugin needs to use a new API. Just upgrade the sdk plugin and you'll be able to use the new instance type.

silverjam:
Ok, will give it a shot!

PR #875, opened 2023-07-28
Test support for cloud management changes

car-roll:
The moving of cloud management to multiple pages (https://github.com/jenkinsci/jenkins/pull/7658) appear to have broken some tests.

Bug also discovered when changing cloud name. See https://issues.jenkins.io/browse/JENKINS-71737

Also upstream of #888 

cc @Vlatombe 

### Testing done

```[tasklist]
### Submitter checklist
```


src/test/java/hudson/plugins/ec2/ConfigurationAsCodeTest.java
```diff
@@ -34,13 +34,14 @@ public class ConfigurationAsCodeTest {
...
     public JenkinsConfiguredWithCodeRule j = new JenkinsConfiguredWithCodeRule();
 
-    @Test
-    @ConfiguredWithCode("EC2CloudEmpty.yml")
-    public void testEmptyConfig() throws Exception {
-        final AmazonEC2Cloud ec2Cloud = (AmazonEC2Cloud) Jenkins.get().getCloud("");
-        assertNotNull(ec2Cloud);
-        assertEquals(0, ec2Cloud.getTemplates().size());
-    }
+    // Jenkins 2.403: Cloud Nodes cannot be created with an empty name
```
car-roll, line 37: Unable to catch the illegal argument exception when the cloud node is created. Wasn't sure if there was a way to do that for this particular test

src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
```diff
@@ -92,7 +92,14 @@ public void testSshKeysCredentialsIdRemainsUnchangedAfterUpdatingOtherFields() t
         HtmlTextInput input = form.getInputByName("_.cloudName");
 
         input.setText("test-cloud-2");
-        r.submit(form);
```
car-roll, line 95: For this test, `r.submit(form)` returns 
```
FailingHttpStatusCode 404 Not Found for http://localhost:65397/jenkins/cloud/us-east-1/
```
I'm not sure if this is a regression, but it looks like because we change the name of the cloud node, the url is now dead. If this is acceptable, I did not want to add an `assertThrows` in case changes in the core would make that error go away.

pom.xml
```diff
@@ -78,7 +78,7 @@ THE SOFTWARE.
     <properties>
         <revision>2.0.8</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.361.4</jenkins.version>
+        <jenkins.version>2.403</jenkins.version>
```
car-roll, line 81: I chose 2.403 because that was when the changes were first introduced. Most recent LTS is 2.401.3.

jtnord: `2.414` will become the lts line and there is a bom for this line (which you also should update!)

https://github.com/jenkinsci/ec2-plugin/pull/875/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R222-R223

src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
raul-arabaolaza: Sorry, @car-roll I did not fully understood your comment. From what you say it looks like changing the cloud name now is broken? If not what is being done by the production code to deal with that 404 that now is happening? It looks to me that instead of a try catch we should prevent that error to happen in first case

src/test/java/hudson/plugins/ec2/ConfigurationAsCodeTest.java
raul-arabaolaza: Maybe you can check the instance logs and search for the exception there?. More important, is this change in behaviour expected or a side effect?

car-roll: But isn't the point of the test to just create an empty cloud name? The behavior is now blocked in the cloud implementation

src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
```diff
@@ -217,7 +220,8 @@ public void testUnixConfigRoundTrip() throws Exception {
         AmazonEC2Cloud ac = new AmazonEC2Cloud("us-east-1", false, "abc", "us-east-1", "ghi", "3", templates, null, null);
         r.jenkins.clouds.add(ac);
 
-        r.submit(r.createWebClient().goTo("configureClouds").getFormByName("config"));
+        r.submit(getConfigForm(ac));
```
raul-arabaolaza, line 223: Why submit twice?

car-roll: because i didn't carefully check my find and replace 🤦

car-roll: resolved in ca8172ba72f4cbaa7aadee92e5ad3f336e77034d

src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
car-roll: Ouch, yeah so I manually tested this and this is a regression. Changing the names in the cloud config page fails on 404 😬

car-roll:
So `testSshKeysCredentialsIdRemainsUnchangedAfterUpdatingOtherFields` still fails. Looks to be a regression. When you change the cloud node name, the user then receives a 404 saying the URL to the old cloud config page does not exist.
However, the cloud name does change successfully when you go back, it's just that submitting the name change tries to take you back to the old cloud name config page.

car-roll:
Turns out the issue is here: https://github.com/jenkinsci/jenkins/blob/31721319231f19bead0adcbb005b99c84a6f9171/core/src/main/java/hudson/slaves/Cloud.java#L334
The redirect is incorrect to return to the top of the clouds page. Will file a PR for fix.

raul-arabaolaza:
I beieve this handles properly the UI changes, while the 404 is managed by the core pr, I see no reason not to ship this fixes.

pom.xml
```diff
@@ -78,7 +78,7 @@ THE SOFTWARE.
     <properties>
         <revision>2.0.8</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.361.4</jenkins.version>
+        <jenkins.version>2.420-rc34091.108b_c22ede78</jenkins.version> <!-- TODO: https://github.com/jenkinsci/jenkins/pull/8310 -->
```
raul-arabaolaza, line 81: I guess this should be `2.414` to be aligned with the BOM version?

src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
```diff
@@ -32,6 +32,7 @@
 import com.cloudbees.plugins.credentials.CredentialsStore;
 import com.cloudbees.plugins.credentials.SystemCredentialsProvider;
 import com.cloudbees.plugins.credentials.domains.Domain;
+import org.htmlunit.FailingHttpStatusCodeException;
```
raul-arabaolaza, line 35:
```suggestion
```

Leftover from previous commits now removed I guess

src/test/java/hudson/plugins/ec2/ConfigurationAsCodeTest.java
```diff
@@ -37,7 +37,7 @@ public class ConfigurationAsCodeTest {
     @Test
     @ConfiguredWithCode("EC2CloudEmpty.yml")
     public void testEmptyConfig() throws Exception {
-        final AmazonEC2Cloud ec2Cloud = (AmazonEC2Cloud) Jenkins.get().getCloud("");
+        final AmazonEC2Cloud ec2Cloud = (AmazonEC2Cloud) Jenkins.get().getCloud("empty");
```
raul-arabaolaza, line 40: This is changing the test intent, is it intentional?

raul-arabaolaza (commented):
I believe we should try to move this forward

src/test/java/hudson/plugins/ec2/ConfigurationAsCodeTest.java
car-roll: Actually we had this discussion previously that was resolved :)  https://github.com/jenkinsci/ec2-plugin/pull/875#discussion_r1277677549

The purpose of the test was to create an ec2Cloud with no templates, so using the name "empty" is still okay. There is a check in core now that prevents the creation of a template with no name.

pom.xml
car-roll: sounds good, I will go back to 2.414, but there still will be the failing test. Should I add an '@Ignore` with a comment on the related jenkins ticket then?

src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
car-roll: removed in 60e1b46a4ae4e2363216adf5e604acfa294512e7

pom.xml
raul-arabaolaza: > Should I add an '@Ignore` with a comment on the related jenkins ticket then?

AFAIUI the intent of `testSshKeysCredentialsIdRemainsUnchangedAfterUpdatingOtherFields` is to check the credentials are not changing after editing it, so we could also slightly adjust the test to change other field instead of the cloud name. That way we do not have to ignore the test or block the release. I would add also a comment with the link to the core issue so we restore/add the original test when done.

car-roll: test updated in 3ae03ee064c62110cdefc1d2d209ab325fedc6de

car-roll:
@Vlatombe I believe this PR is ready for review now. Would you be able to take a look at it?

src/test/java/hudson/plugins/ec2/AmazonEC2CloudTest.java
car-roll: yes, with a dedicated rename page

car-roll:
Also created https://github.com/car-roll/ec2-plugin/pull/1 to handle a separate name change should https://github.com/jenkinsci/jenkins/pull/8310 get in before this PR

rsandell (approved):

Vlatombe (approved):

car-roll:
Sorry forgot to remove an unnecessary TODO (related to name change fixes). I changed the test so it didn't rely on name changes and just changed a different property.

Vlatombe (approved):

car-roll:
dealing with CI flakes

car-roll:
@res0nance Could you take a look at this PR? Tests fail here when the jenkins version is upgraded. It also revealed a regression in core (See https://github.com/jenkinsci/jenkins/pull/8310/)

Will continue to work on getting CI to pass. I believe it is infrastructure right now that is making it fail.

car-roll:
ci now passes

res0nance (approved):

res0nance:
Hi @car-roll changes look fine would you like this merged or do you still have work to do in this area?

car-roll:
Hi @res0nance this PR is done so we can merge it 🚀  many thanks!

PR #945, opened 2024-03-06
JENKINS-66373: idle timeout should not fight with min spare instances setting

sparrowt:
The problem: currently if you use the "Minimum number of spare instances" setting as well as "Idle termination time", then there is a constant battle going on between:
- `MinimumInstanceChecker.checkForMinimumInstances` provisioning new agents to try and maintain the minimum number of spare instances as configured
- however the idle timeout checks in this method are killing off the 'spare' instances once they reach the idle termination time

This means that the spare instances are repeatedly killed & recreated as described on https://issues.jenkins.io/browse/JENKINS-66373 which is wasteful and means that, some percentage of the time, there are never enough spare instances because they are being booted (again).

This PR makes a simple change to the idle termination logic so that it takes account of "Minimum number of _spare_ instances" in the same way that it already accounts for the main "Minimum number of instances" setting.

### Testing done

None yet - I have found `EC2RetentionStrategyTest.java` and it would definitely make sense to add a relevant test in there for this case.

Right now I've got very limited time and wanted to first get the fix up for discussion.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -131,9 +131,14 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
         SlaveTemplate slaveTemplate = computer.getSlaveTemplate();
         if (slaveTemplate != null) {
             long numberOfCurrentInstancesForTemplate = MinimumInstanceChecker.countCurrentNumberOfAgents(slaveTemplate);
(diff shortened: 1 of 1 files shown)


sparrowt:
@res0nance who would be good to have an initial review of this?

res0nance:
> @res0nance who would be good to have an initial review of this?

Could you add some unit tests for this change?

jperez3:
@sparrowt @res0nance I'm running into something similar, any thoughts on getting this updated/reviewed? 🙏🏽

sparrowt:
Hi folks I'm afraid I don't have time right now to add tests (much as I approve that this would be a good idea!)

If others are keen for this too I'm open to contributions in a PR against my branch

PR #952, opened 2024-03-19
Terminate the EC2 on receiving an exception

BushraManga:
The EC2 node should terminate when an exception is received or exit code is not 0. The current script just ignores & continues after only logging the message but no termination action is taken.


Zendesk-JeffreyCai (approved):

BushraManga:
@lemeurherve Can you please take a look at this PR or ask some one else to do so?

src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -210,9 +210,9 @@ protected void launchScript(EC2Computer computer, TaskListener listener) throws
                 int exitStatus = waitCompletion(sess);
                 if (exitStatus != 0) {
                     logWarning(computer, listener, "init script failed: exit code=" + exitStatus);
-                    return;
+                    sess.close();
```
res0nance, line 213: This will clearly not close the sess object if the script is successful, you probably should hoist this above the if.

PR #957, opened 2024-03-26
JENKINS-72643: Fixed listing regions when using ec2 ARN role with instance profile

craigcomstock:
In the case where a user specified an ARN Role in the advanced section of cloud config we need to use that information when pulling the list of regions.

In my case without this ARN Role being used with credentials I would get an empty drop-down list with no errors.

I added a log message with the exception to expose a bit of information in case of failures.

### Testing done

I did test with an ARN Role in the advanced section of cloud config.

I did not (yet) test without this but will do this next. I suppose I could check if roleArn is empty/null and use a different createCredentialsProvider() signature. 

I did not see an existing test involving ec2 instance profile or listing regions so did not add one.

```[tasklist]
### Submitter checklist
- [x] (n/a) Link to relevant pull requests, esp. upstream and downstream changes
- [x] (skipped, see above comment) Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
@@ -185,6 +185,9 @@ public class AmazonEC2Cloud extends EC2Cloud {
         public ListBoxModel doFillRegionItems(
                 @QueryParameter String altEC2Endpoint,
                 @QueryParameter boolean useInstanceProfileForCredentials,
+                @QueryParameter String roleArn,
+                @QueryParameter String roleSessionName,
+                @QueryParameter String region,
                 @QueryParameter String credentialsId)
 
                 throws IOException, ServletException {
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -185,14 +185,20 @@ public FormValidation doCheckAltEC2Endpoint(@QueryParameter String value) {
...
+                @QueryParameter String roleArn,
+                @QueryParameter String roleSessionName,
+                @QueryParameter String region,
                 @QueryParameter String credentialsId)
 
                 throws IOException, ServletException {
             ListBoxModel model = new ListBoxModel();
             if (Jenkins.get().hasPermission(Jenkins.ADMINISTER)) {
                 try {
                     AWSCredentialsProvider credentialsProvider = createCredentialsProvider(useInstanceProfileForCredentials,
```
craigcomstock, line 197: as far as I can tell from EC2Cloud.java my use here should be OK because there is a guard for empty roleArn and roleSessionName

https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/java/hudson/plugins/ec2/EC2Cloud.java#L956

```
        if (StringUtils.isNotEmpty(roleArn) && StringUtils.isNotEmpty(roleSessionName)) {
            return new STSAssumeRoleSessionCredentialsProvider.Builder(roleArn, roleSessionName)
...
```

src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -201,6 +207,7 @@ public ListBoxModel doFillRegionItems(
                         model.add(name, name);
                     }
                 } catch (SdkClientException ex) {
+                    LOGGER.log(Level.INFO, "AmazonEC2Cloud.doFillRegionItems() got exception: " + ex);
```
craigcomstock, line 210: I was thinking I would change this message to incorporate the comment about this may happen before credentials are specified so maybe "OK".

Can we write unit tests which check for logging? The logging would have been VERY helpful to me initially when I encountered this issue.

craigcomstock:
this is quite stale so will let it go.

PR #968, opened 2024-05-08
JENKINS-73096 Create config for check interval and use the same default

sfc-gh-mayang:
This PR allowed us to mitigate an active incident. The previous behavior is as follows:
- terminate node is called
- updateComputerList calls updateComputer on every node
- updateComputer calls SlaveComputer.check which runs describeInstances if time is after nextCheckAfter

This isn't an issue when we are terminating less than 600 nodes. But since describeInstances take around 100ms. When UpdateComputerList is called, to update every node, the total duration will exceed the current hard coded value of 1 minutes. So every computer termination takes 1 min to finish, during this time Queue lock is taken so no new work can be accepted.

TESTING PLAN
Installed on our controllers, launched 800 nodes via minInstances, then set minInstances to 0. Instead of the controller locking up until it reaches below 700 idle instances which takes more than an hour. It cleans all nodes up within 5 min

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -68,6 +68,8 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
     private transient ReentrantLock checkLock;
     private static final int STARTUP_TIME_DEFAULT_VALUE = 30;
 
+    private static final Integer CHECK_INTERVAL_MINUTES = Integer.getInteger("jenkins.ec2.checkIntervalMinutes", 1);
+
     @DataBoundConstructor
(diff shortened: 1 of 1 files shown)


res0nance (approved):

mwebber:
@sfc-gh-mayang 

Is it possible to document the new setting `jenkins.ec2.checkIntervalMinutes`  somewhere, such as https://plugins.jenkins.io/ec2/ ?

You said
> TESTING PLAN
Installed on our controllers, launched 800 nodes via minInstances, then set minInstances to 0. Instead of the controller locking up until it reaches below 700 idle instances which takes more than an hour. It cleans all nodes up within 5 min

What value of `jenkins.ec2.checkIntervalMinutes` did you use when you did the test?

sfc-gh-mayang:
We used a value of 2 to test

PR #993, opened 2024-10-15
Dynamically create sshKeyPair if one is not supplied via credentials via the configuration

mikecirioli:
This PR proposes a means of automatically creating an `ssh keypair` to use when launching an instance if the user has not manually specified one via the use of credentials in the configuration.  This change is backwards compatible with the current behavior - if  an `ssh credential` is specified in the configuration page then no attempt will be made to use dynamic ssh keys management.  

If no ssh private key is configured then the plugin will automatically generate one and then assign it to the instance being launched.  The private key and the AWS `keyPair` name are persisted as part of the node definition so that it can be used for `ssh` connections and also so that the `keyPair` can deleted once the node is scheduled for termination.

In the case where the request is made to launch multiple instances from a single template **AND** dynamic ssh keys are being used then the request to create instances will internally be unrolled into **N** requests to create **N** instead of a single request for multiple instances.  This is done so that each instance gets its own unique SSH keypair. 

In the event the controller restarts while an `ec2` instance is running the controller will use the `sshPrivateKey` persisted with the agent  definition to re-attach to any running jobs.  

**Note 1:**  This change does not touch any legacy code where the static ssh key was originally stored as a `hudson.util.Secret` (or the associated migration code).  This is a new implementation with full backward compatibility.

**Note 2:**  If you decide to switch from dynamic key management to using a static SSH credential it is advisable that you terminate and restart any running agent instances, otherwise the controller will not be able to re-attach to them if they are disconnected.  This is because the presence of a configured SSH credential is used as the switch between the new and old behavior.  This could be overcome by always using the key associated with the instance (if one is defined) and otherwise attempt to use a static key if one is defined.

**Note 3:** Using dynamic ssh keys with windows ec2 instances is not supported
 
### Testing done

- [x] manual testing
- [x] new unit tests
- [x] update docs

### Submitter checklist


mikecirioli:
should I add a check and generate the key if needed when the cloud config is saved (instead of waiting for the first request to bring up an `ec2` instance?)

mikecirioli:
update:  WIP to change the approach:

- if the user has supplied their own private key, that will be used (current behavior of the plugin)
- if no user supplied private key then a new `keypair` will be created for the instance being spawned and this `keypair` will be deleted once the agent has been terminated  (ie. single-use `keypairs` for each agent)

mikecirioli:
The failing test sometimes passes locally and sometimes note, still investigating

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -112,6 +117,10 @@
     public final Tenancy tenancy;
     private String instanceType;
 
+    private KeyPair instanceSshKeyPair;
+    private Secret instanceSshPrivateKey;
+    private String instanceSshKeyPairName;
```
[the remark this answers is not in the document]
mikecirioli: it does not

mikecirioli:
> The failing test sometimes passes locally and sometimes note, still investigating

tests are fixed now

src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -314,7 +319,21 @@ private File createIdentityKeyFile(EC2Computer computer) throws IOException {
         EC2PrivateKey ec2PrivateKey = computer.getCloud().resolvePrivateKey();
         String privateKey = "";
         if (ec2PrivateKey != null){
+            LOGGER.log(Level.INFO, () -> "using global private key");
```
jglick, line 322: Not interesting enough to pollute the system log.

```suggestion
            LOGGER.fine(() -> "using global private key for " + computer.getName());
```

src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -364,38 +383,46 @@ private File createHostKeyFile(EC2Computer computer, String ec2HostAddress, Task
...
 
-    private boolean bootstrap(EC2Computer computer, TaskListener listener, SlaveTemplate template) throws IOException,
+    private boolean bootstrap(@NonNull EC2Computer computer, TaskListener listener, SlaveTemplate template) throws IOException,
             InterruptedException, AmazonClientException {
         logInfo(computer, listener, "bootstrap()");
         Connection bootstrapConn = null;
         try {
             int tries = bootstrapAuthTries;
             boolean isAuthenticated = false;
+            //TODO: mikec this will need to be adapted
```
jglick, line 393:
```suggestion
```

or what?

src/main/java/hudson/plugins/ec2/ssh/verifiers/CheckNewHardStrategy.java
```diff
@@ -65,7 +65,7 @@ public boolean verify(EC2Computer computer, HostKey hostKey, TaskListener listen
                     // Keep trying for at least 2 minutes
                     stop = computer.getUptime() > TimeUnit.SECONDS.toMillis(120);
                 } catch (Exception ignored) {
-
+                    stop = true;
```
jglick, line 68: Is this related?

src/main/resources/hudson/plugins/ec2/EC2Cloud/help-sshKeysCredentialsId.html
```diff
@@ -22,14 +22,21 @@
...
-		that starts with "-----BEGIN RSA PRIVATE KEY-----". You normally
-		generate it via using either the AWS Management Console, the "ec2-add-keypair" tool or a GUI tool
-		like ElastiFox.
+	<p>If no credential (ie. "none") is selected then dynamically managed SSH
+	key pairs will be used.  For each instance that is requested a new EC2 key pair
+	will be created and used only for the lifespan of each node.  Upon termination,
+	the keypair will then be removed from EC2.</p>
+
+
+	<p>Optionally, you may specify a Jenkins credential of the type"SSH Username with
```
jglick, line 31:
```suggestion
	<p>Optionally, you may specify a Jenkins credential of the type "SSH Username with
```

src/test/java/hudson/plugins/ec2/EC2RetentionStrategyTest.java
jglick, line 1: Why is this edited?

src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
```diff
@@ -70,6 +74,7 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static com.ibm.icu.impl.locale.KeyTypeData.ValueType.any;
```
jglick, line 77:
```suggestion
```

jglick (changes requested):
Bogus serializable field being introduced.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -485,20 +509,73 @@ void stop() {
...
+//            ec2.terminateInstances(request);
+//            LOGGER.info("EC2 instance terminate request sent for " + getInstanceId());
+//
+//            return true;
+//        } catch (AmazonClientException e) {
+//            LOGGER.log(Level.WARNING, "Failed to terminate EC2 instance: " + getInstanceId(), e);
+//            return false;
+//        }
+//    }
+
```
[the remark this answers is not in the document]
mikecirioli: I am in favor of removing it, i am not aware of any other implementation that might be relying on it.

src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -230,7 +231,8 @@ public FormValidation doTestConnection(
                 @QueryParameter String credentialsId,
                 @QueryParameter String sshKeysCredentialsId,
                 @QueryParameter String roleArn,
-                @QueryParameter String roleSessionName)
+                @QueryParameter String roleSessionName,
+                @QueryParameter Secret sshPrivateKeySecret)
```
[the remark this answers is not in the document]
mikecirioli: i think that might have been a "leftover" from an earlier implementation - i will check and remove if its not being used

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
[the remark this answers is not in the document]
jglick: As a rule, a PR should not introduce commented-out code without a good reason.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -485,20 +509,73 @@ void stop() {
...
+            }
+        }
+        return false;
+    }
+
+    protected void cleanupSshKeyPairIfNeeded() {
+        AmazonEC2 ec2 = getCloud().connect();
+
+        if (isUsingDynamicSshKeys()) {
+            //this instance is using dynamic ssh keys
```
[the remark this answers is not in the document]
mikecirioli: i think i was trying to be defensive against deleting a static ssh key by accident.  its not possible that this field could contain anything other than a dynamic key identifier with the current code though, so i will clean it up

src/main/java/hudson/plugins/ec2/ssh/verifiers/CheckNewHardStrategy.java
mikecirioli: no

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -485,17 +508,26 @@ void stop() {
...
-            ec2.terminateInstances(request);
-            LOGGER.info("EC2 instance terminate request sent for " + getInstanceId());
-            return true;
-        } catch (AmazonClientException e) {
-            LOGGER.log(Level.WARNING, "Failed to terminate EC2 instance: " + getInstanceId(), e);
-            return false;
+            ec2.deleteKeyPair(new DeleteKeyPairRequest().withKeyName(this.getInstanceSshKeyPairName()));
+        } else {
+            // should not be possible to reach this code
+            LOGGER.warning(() -> "Unable to determine keypair name to delete for this instance");
```
jglick, line 530: No longer true IIUC; rather, the normal case when deleting an agent using the original manually configured keypair?

Anyway you need to note the name of the agent or the log message is useless:

```suggestion
            LOGGER.fine(() -> "No dynamic keypair to delete for " + getNodeName());
```

src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -42,6 +42,7 @@
 import edu.umd.cs.findbugs.annotations.Nullable;
 import javax.servlet.ServletException;
 
+import hudson.util.Secret;
```
amuniz, line 45: Unused?

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -157,7 +165,7 @@ public abstract class EC2AbstractSlave extends Slave {
...
+    public EC2AbstractSlave(String name, String instanceId, String templateDescription, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy<EC2Computer> retentionStrategy, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String javaPath, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses, Tenancy tenancy, Boolean metadataEndpointEnabled, Boolean metadataTokensRequired, Integer metadataHopsLimit, Boolean metadataSupported, KeyPair keypair)
```
amuniz, line 168: Is this backward compatible?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1186,32 +1214,31 @@ protected FormValidation doTestConnection(@AncestorInPath ItemGroup context, URL
...
-                String privateKey = "";
-                if (sshCredential != null) {
-                    privateKey = sshCredential.getPrivateKey();
-                } else {
-                    return FormValidation.error("Failed to find credential \"" + sshKeysCredentialsId + "\" in store.");
-                }
-
                 AWSCredentialsProvider credentialsProvider = createCredentialsProvider(useInstanceProfileForCredentials, credentialsId, roleArn, roleSessionName, region);
                 AmazonEC2 ec2 = AmazonEC2Factory.getInstance().connect(credentialsProvider, ec2endpoint);
                 ec2.describeInstances();
```
amuniz, line 1219: Is this needed?

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
mikecirioli: yes, it is ok for `keypair` to be `null`

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: nope, i do not think so (that was in the original code) - i will remove it and test

amuniz: Maybe it's there just to check credentials are ok. Anyways, I thought it was added in this PR, if it's not, then let's leave it there.

amuniz (approved):
Looks good to me.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: seems to function fine without it, and i reviewed the api call docs and there seems to be no purpose for it here.

amuniz (approved):
Looks good. Assuming this has been tested interactively using a predefined key to verify the original behaviour is not broken.

mikecirioli:
> Looks good. Assuming this has been tested interactively using a predefined key to verify the original behaviour is not broken.

yes, i have tested this scenario locally (and all the original unit tests based on that approach also pass)

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
jglick: AFAICT this is still incompatible for Java callers. Probably no one cares about that, but perhaps it would be better to just revert this change (and the one to the deprecate ctor below) and just add a `setKeypair` method. Could be package scope.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -485,17 +508,25 @@ void stop() {
 
     }
 
-    boolean terminateInstance() {
-        try {
+    private boolean isUsingDynamicSshKeys() {
```
jglick, line 511: This is now dead code; delete it.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -172,7 +186,7 @@ public abstract class EC2Cloud extends Cloud {
 
     private transient volatile AmazonEC2 connection;
 
-    protected EC2Cloud(String name, boolean useInstanceProfileForCredentials, String credentialsId, String privateKey, String sshKeysCredentialsId,
+    protected EC2Cloud(String name, boolean useInstanceProfileForCredentials, String credentialsId, String sshPrivateKey, String sshKeysCredentialsId,
```
jglick, line 189: The fact that you could rename this parameter without making any other changes indicates that it was not actually used, so maybe delete the parameter assuming there are no external subclasses? Or at least rename it to something that indicates it is unused.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1186,32 +1214,30 @@ protected FormValidation doTestConnection(@AncestorInPath ItemGroup context, URL
...
-                String privateKey = "";
-                if (sshCredential != null) {
-                    privateKey = sshCredential.getPrivateKey();
-                } else {
-                    return FormValidation.error("Failed to find credential \"" + sshKeysCredentialsId + "\" in store.");
-                }
-
                 AWSCredentialsProvider credentialsProvider = createCredentialsProvider(useInstanceProfileForCredentials, credentialsId, roleArn, roleSessionName, region);
                 AmazonEC2 ec2 = AmazonEC2Factory.getInstance().connect(credentialsProvider, ec2endpoint);
-                ec2.describeInstances();
```
jglick, line 1199: Why is this being deleted? IIUC this was an important part of form validation.

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -67,14 +66,18 @@ public EC2OndemandSlave(String name, String instanceId, String templateDescripti
...
+        this(name, instanceId, templateDescription, remoteFS, numExecutors, labelString, mode, initScript, tmpDir, nodeProperties, remoteAdmin, DEFAULT_JAVA_PATH, jvmopts, stopOnTerminate, idleTerminationMinutes, publicDNS, privateDNS, tags, cloudName, launchTimeout, amiType, connectionStrategy, maxTotalUses, tenancy, metadataEndpointEnabled, metadataTokensRequired, metadataHopsLimit, DEFAULT_METADATA_SUPPORTED, null);
+    }
+
+    public EC2OndemandSlave(String name, String instanceId, String templateDescription, String remoteFS, int numExecutors, String labelString, Mode mode, String initScript, String tmpDir, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String javaPath, String jvmopts, boolean stopOnTerminate, String idleTerminationMinutes, String publicDNS, String privateDNS, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses, Tenancy tenancy, Boolean metadataEndpointEnabled, Boolean metadataTokensRequired, Integer metadataHopsLimit, Boolean metadataSupported) throws hudson.model.Descriptor.FormException, java.io.IOException {
```
jglick, line 72: Same, avoid changing existing ctors if you do not really need to.

src/main/java/hudson/plugins/ec2/EC2PrivateKey.java
```diff
@@ -143,6 +184,7 @@ public String decryptWindowsPassword(String encodedPassword) throws AmazonClient
         }
     }
 
+
```
jglick, line 187:
```suggestion
```

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -94,17 +99,15 @@ public void terminate() {
...
                                     TerminateInstancesRequest request = new TerminateInstancesRequest(Collections.singletonList(instanceId));
-                                    try {
-                                        ec2.terminateInstances(request);
-                                        LOGGER.info("Terminated EC2 instance (terminated): " + instanceId);
-                                    } catch (AmazonClientException e) {
-                                        // Spot request is no longer valid
-                                        LOGGER.log(Level.WARNING, "Failed to terminate the Spot instance: " + instanceId, e);
-                                    }
+                                    ec2.terminateInstances(request);
+                                    LOGGER.info("Terminated EC2 instance (terminated): " + instanceId);
```
jglick, line 107: So if `cleanupSshKeyPairIfNeeded` fails, then instance will not be deleted. I think you should revert the error-handling changes in this block and wrap `cleanupSshKeyPairIfNeeded` in its own `try`-`catch` just to be sure.

src/main/java/hudson/plugins/ec2/InstanceInfo.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.amazonaws.services.ec2.model.Instance;
+import com.amazonaws.services.ec2.model.KeyPair;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class InstanceInfo {
+    private Instance instance;
+    private KeyPair keypair;
```
jglick, line 11:
```suggestion
    private final Instance instance;
    private final KeyPair keypair;
```
(Candidate for `record` when on 2.479+ and thus Java 17.)

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -931,8 +932,9 @@ public enum ProvisionOptions { ALLOW_CREATE, FORCE_CREATE }
     public List<EC2AbstractSlave> provision(int number, EnumSet<ProvisionOptions> provisionOptions) throws AmazonClientException, IOException {
         final Image image = getImage();
         if (this.spotConfig != null) {
-            if (provisionOptions.contains(ProvisionOptions.ALLOW_CREATE) || provisionOptions.contains(ProvisionOptions.FORCE_CREATE))
+            if (provisionOptions.contains(ProvisionOptions.ALLOW_CREATE) || provisionOptions.contains(ProvisionOptions.FORCE_CREATE)) {
                 return provisionSpot(image, number, provisionOptions);
+            }
```
jglick, line 937: Unrelated?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1394,6 +1418,16 @@ private void setupCustomDeviceMapping(List<BlockDeviceMapping> deviceMappings) {
         }
     }
 
+    private boolean isUsingDynamicSshKeys() {
+        if (parent != null) {
+            String id = parent.getSshKeysCredentialsId();
+            if ((id == null) || (id.isEmpty())) {
```
jglick, line 1424: `Util.isEmpty` I guess, though best would be to make sure that the setter does not allow an empty string to be saved to begin with.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1644,18 +1686,22 @@ protected EC2SpotSlave newSpotSlave(SpotInstanceRequest sir) throws FormExceptio
 
     /**
      * Get a KeyPair from the configured information for the agent template
+     *
+     * returns null if no global private key has been configured
```
jglick, line 1690:
```suggestion
     * @return null if no global private key has been configured
```

src/main/java/hudson/plugins/ec2/ssh/verifiers/CheckNewHardStrategy.java
jglick: Then why are you doing it here?

src/main/resources/hudson/plugins/ec2/EC2Cloud/help-sshKeysCredentialsId.html
```diff
@@ -22,14 +22,21 @@
...
+	private key" containing the private key of the EC2 key pair. You can use the AWS Console
+	or any other tool of your choice to generate the key pair.</p>
+
+
+	<p>The username is not used, and the credential should not have a passphrase.</p>
+
+
+	<p>An RSA private key for EC2 is a text that starts with
+	"-----BEGIN RSA PRIVATE KEY-----". You normally generate it via using either
+	the AWS Management Console, the "ec2-add-keypair" tool or a GUI tool like ElastiFox.</p>
```
jglick, line 41: BTW the reformatting makes the diff harder to read: obscures the fact that you are not actually editing this paragraph.

jglick (changes requested):
https://github.com/jenkinsci/ec2-plugin/pull/993#discussion_r1820883617 & https://github.com/jenkinsci/ec2-plugin/pull/993#discussion_r1820890970 mainly

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: is that because it proves `ec2` policy is correctly defined?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1164,44 +1166,66 @@ private List<EC2AbstractSlave> provisionOndemand(Image image, int number, EnumSe
...
 
+    List<InstanceInfo> createNewInstances(AmazonEC2 ec2, RunInstancesRequest riRequest) {
+        if (isUsingDynamicSshKeys()) {
+            // no static ssh key defined, so submit riRequest.maxCount requests
+            // each with a unique keypair (instead of a single request for multiple identical instances)
+            List<InstanceInfo> instances = new ArrayList<>();
+            int maxRequested = riRequest.getMaxCount();
+            riRequest.setMaxCount(1);
+            for (int i=0;i < maxRequested; i++) {
+                String keyName = "jenkins-ec2-" + System.currentTimeMillis();
```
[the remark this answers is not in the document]
mikecirioli: i named it this way just to make it clear in the aws ux.  I suppose just using `jenkins-` is clear enough.  Thanks for the `uuid` suggestion as i was also a little worried about sub-millsecond method calls

src/main/java/hudson/plugins/ec2/EC2Cloud.java
jglick: Or just revert the change, which does not seem to have any purpose.

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -187,6 +195,9 @@ public EC2AbstractSlave(String name, String instanceId, String templateDescripti
         this.metadataTokensRequired = metadataTokensRequired;
         this.metadataHopsLimit = metadataHopsLimit;
         this.metadataSupported = metadataSupported;
+        this.instanceSshKeyPairName = null;
+        this.instanceSshPrivateKey = null;
+
```
jglick, line 200: just
```suggestion
```

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -256,6 +267,21 @@ protected Object readResolve() {
         return o;
     }
 
+    protected void setInstanceKeypair(KeyPair keypair) {
```
jglick, line 270: This does not appear to ever be called?

jglick (changes requested):

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
mikecirioli: i've reverted the error handling and wrapped the `cleanupSshKeyPairIfNeeded()` call with a `try/catch/log`

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
mikecirioli: yes, i missed a [commit](https://github.com/jenkinsci/ec2-plugin/pull/993/commits/4a3e23c7b5e271d0df45fe4a809631dbaeb2731d) :facepalm:

fcojfernandez (approved):
@mikecirioli Should 

> If you decide to switch from dynamic key management to using a static SSH credential it is advisable that you terminate and restart any running agent instances, otherwise the controller will not be able to re-attach to them if they are disconnected.

be added in the README before merging?

README.md
```diff
@@ -60,6 +60,13 @@ to generate the private key to interactively log in to EC2 instances.
 Alternatively you may select **None** for the SSH credentials and the plugin
 will dynamically manage the SSH key pair used by each instance.
 
+```
+If you decide to switch from dynamic key management to using a static SSH
+credential it is advisable that you terminate and restart any running agent
+instances, otherwise the controller will not be able to re-attach to them
+if they are disconnected.
+```
```
jglick, line 68: Why was this a code block?

```suggestion
If you decide to switch from dynamic key management to using a static SSH
credential it is advisable that you terminate and restart any running agent
instances, otherwise the controller will not be able to re-attach to them
if they are disconnected.
```

src/main/java/hudson/plugins/ec2/util/EC2AgentFactoryImpl.java
```diff
@@ -12,11 +12,15 @@ public class EC2AgentFactoryImpl implements EC2AgentFactory {
...
+        instance.setInstanceKeypair(config.keyPair);
+        return instance;
     }
 
     @Override
     public EC2SpotSlave createSpotAgent(EC2AgentConfig.Spot config) throws Descriptor.FormException, IOException {
-        return new EC2SpotSlave(config.name, config.spotInstanceRequestId, config.description, config.remoteFS, config.numExecutors, config.mode, config.initScript, config.tmpDir, config.labelString, config.nodeProperties, config.remoteAdmin, config.javaPath, config.jvmopts, config.idleTerminationMinutes, config.tags, config.cloudName, config.launchTimeout, config.amiType, config.connectionStrategy, config.maxTotalUses);
+        EC2SpotSlave instance =  new EC2SpotSlave(config.name, config.spotInstanceRequestId, config.description, config.remoteFS, config.numExecutors, config.mode, config.initScript, config.tmpDir, config.labelString, config.nodeProperties, config.remoteAdmin, config.javaPath, config.jvmopts, config.idleTerminationMinutes, config.tags, config.cloudName, config.launchTimeout, config.amiType, config.connectionStrategy, config.maxTotalUses, config.keyPair);
```
jglick, line 22:
```suggestion
        EC2SpotSlave instance =  new EC2SpotSlave(config.name, config.spotInstanceRequestId, config.description, config.remoteFS, config.numExecutors, config.mode, config.initScript, config.tmpDir, config.labelString, config.nodeProperties, config.remoteAdmin, config.javaPath, config.jvmopts, config.idleTerminationMinutes, config.tags, config.cloudName, config.launchTimeout, config.amiType, config.connectionStrategy, config.maxTotalUses);
```
? I think you forgot to revert the setter signature for this class.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -220,11 +220,15 @@ protected EC2Cloud(String id, boolean useInstanceProfileForCredentials, String c
     public KeyPair resolveKeyPair() throws IOException {
         KeyPair keyPair = null;
 
-        if (sshKeysCredentialsId != null) {
+        if (sshKeysCredentialsId != null && !sshKeysCredentialsId.isEmpty()) {
```
jglick, line 223: OK though better to have the corresponding setter call `fixEmpty` so the config does not store

```xml
<sshKeysCredentialsId></sshKeysCredentialsId>
```

to begin with.

amuniz (approved):
I don't see anything obviously wrong, but I don't feel like I have the required knowledge about this plugin to find real issue by code inspection. If you manually tested this and it works (both with static and dynamic SSH keys) then I guess it's fine.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -235,9 +239,11 @@ public EC2PrivateKey resolvePrivateKey() throws IOException {
         if (sshKeysCredentialsId != null) {
             SSHUserPrivateKey privateKeyCredential = getSshCredential(sshKeysCredentialsId, Jenkins.get());
             if (privateKeyCredential != null) {
+                LOGGER.fine("private key resolved from sshCredential");
                 return new EC2PrivateKey(privateKeyCredential.getPrivateKey());
             }
         }
+        LOGGER.info("unable to resolve private key");
```
jglick, line 246: This is going to appear in system logs with no context and make no sense. Was something supposed to “resolve [a] private key”? What? Why was it not resolved? Is something broken?

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
```diff
@@ -508,14 +508,14 @@ void stop() {
 
     }
 
-    protected void cleanupSshKeyPairIfNeeded() {
-        if (getInstanceSshKeyPairName() != null) {
+    protected void cleanupSshKeyPairIfNeeded() throws java.io.IOException {
+        if (getCloud().resolveKeyPair() == null) {
```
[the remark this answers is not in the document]
mikecirioli: Now we are always stashing a reference to the keypair regardless of using dynamic or static keys, so you can't rely on the prescense of this field to tell you how the controller is configured.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: i can take a look

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -807,7 +816,7 @@ public Collection<PlannedNode> provision(final Label label, int excessWorkload)
             }
         }
         LOGGER.log(Level.INFO, "We have now {0} computers, waiting for {1} more",
-            new Object[]{jenkinsInstance.getComputers().length, plannedNodes.size()});
+            new Object[]{jenkinsInstance.getComputers().length, plannedNodes.size() - jenkinsInstance.getComputers().length});
```
jglick, line 819: This does not sound right. `plannedNodes` is supposed to be the list of new nodes about to be added, so why would you substract a generally far larger number of agents, including ones that have nothing to do with this plugin at all?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -850,7 +859,7 @@ public void provision(SlaveTemplate t, int number) {
 
             LOGGER.log(Level.INFO, "{0}. Attempting provision finished", t);
             LOGGER.log(Level.INFO, "We have now {0} computers, waiting for {1} more",
-              new Object[]{Jenkins.get().getComputers().length, number});
+              new Object[]{Jenkins.get().getComputers().length, number - jenkinsInstance.getComputers().length});
```
jglick, line 862: ditto

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1232,6 +1242,10 @@ protected FormValidation doTestConnection(@AncestorInPath ItemGroup context, URL
                                         .error("The EC2 key pair private key isn't registered to this EC2 region (fingerprint is "
                                                 + pk.getFingerprint() + ")");
                         }
+                    } else {
+                        LOGGER.warning(() -> "SSH credential was invalid");
```
jglick, line 1246:
```suggestion
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1232,6 +1242,10 @@ protected FormValidation doTestConnection(@AncestorInPath ItemGroup context, URL
...
+                        return FormValidation.error("SSH Credential was invalid");
```
jglick, line 1247: “Invalid” how? I think you mean that no credentials of that id exist. Not something you would normally see since the dropdown should only be letting you pick valid ids.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1210,20 +1210,31 @@ List<InstanceInfo> createNewInstances(AmazonEC2 ec2, RunInstancesRequest riReque
...
             }
             return instances;
         } else {
             // using a static ssh key
-            return InstanceInfo.fromInstances(ec2.runInstances(riRequest).getReservation().getInstances());
+            List<InstanceInfo> instances = InstanceInfo.fromInstances(ec2.runInstances(riRequest).getReservation().getInstances());
+            instances.stream().forEach(instance -> {
+                try {
+                    LOGGER.fine(() -> "static ssh credential configured, retrieving keypair and setting it in instance");
+                    instance.setKeypair(getParent().resolveKeyPair());
```
jglick, line 1232: Seems off. The `InstanceInfo.keypair` should have been populated in its constructor by `fromInstances`.

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -43,12 +43,17 @@ public EC2SpotSlave(String name, String spotInstanceRequestId, String templateDe
...
             throws FormException, IOException {
+        this(name, spotInstanceRequestId, templateDescription, remoteFS, numExecutors, mode, initScript, tmpDir,labelString, nodeProperties, remoteAdmin, javaPath, jvmopts,idleTerminationMinutes, tags, cloudName, launchTimeout, amiType, connectionStrategy, maxTotalUses, null);
+    }
+
+    @DataBoundConstructor
+    public EC2SpotSlave(String name, String spotInstanceRequestId, String templateDescription, String remoteFS, int numExecutors, Mode mode, String initScript, String tmpDir, String labelString, List<? extends NodeProperty<?>> nodeProperties, String remoteAdmin, String javaPath, String jvmopts, String idleTerminationMinutes, List<EC2Tag> tags, String cloudName, int launchTimeout, AMITypeData amiType, ConnectionStrategy connectionStrategy, int maxTotalUses, KeyPair keyPair)
+            throws FormException, IOException {
```
jglick, line 56: I thought the point of introducing the keypair setter method was so that you could revert ctor signature changes?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: because i see log messages that make no sense, such as 1 have 1 computer, waiting for 1 more` when all that was requested was 1 computer.  I am fine to revert this change but i think the original code is also not correct.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1219,6 +1228,7 @@ protected FormValidation doTestConnection(@AncestorInPath ItemGroup context, URL
                 ec2.describeInstances();
 
                 if (!sshKeysCredentialsId.isEmpty()) {
+                    LOGGER.info("Static ssh credential is defined, validating....");
```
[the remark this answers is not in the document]
mikecirioli: missed one apparently :(

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
mikecirioli: it was, i missed this one

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: fixed

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: reverted

mikecirioli:
regarding manual testing, i have had 2 controllers running for the past couple of hours.  One is configured to use static ssh credentials and the other is using dynamic keys.  So far i have seen no unexpected behavior from either despite constantly running multiple concurrent builds over the duration.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
jglick: > I have 1 computer, waiting for 1 more

Perhaps because the one initial computer was `MasterComputer`? Are you running in a system that sets zero built-in executors?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1164,44 +1167,77 @@ private List<EC2AbstractSlave> provisionOndemand(Image image, int number, EnumSe
...
+                instances.add(new InstanceInfo(instance, keyPair));
+            }
+            return instances;
+        } else {
+            // using a static ssh key
+            List<InstanceInfo> instances = InstanceInfo.fromInstances(ec2.runInstances(riRequest).getReservation().getInstances(), getParent());
+            instances.stream().forEach(instance -> {
+                try {
+                    LOGGER.fine(() -> "static ssh credential configured, retrieving keypair and setting it in instance");
+                    instance.setKeypair(getParent().resolveKeyPair());
```
jglick, line 1232: Why was https://github.com/jenkinsci/ec2-plugin/pull/993#discussion_r1822857034 marked resolved?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
mikecirioli: Apparently i missed removing that after implementing the requested change :facepalm:

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
mikecirioli: Apparently i missed removing that after implementing the requested change :facepalm:  see 
[2bec189](https://github.com/jenkinsci/ec2-plugin/pull/993/commits/2bec189d50906915f549eb2012e65cf3e85b08b5)

mikecirioli:
force push becuase i had to amend the last commit as it contained a file i did not intend to include

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: it looks like the setter is allready calling `fixEmpty()` so i will remove this and test again

src/main/java/hudson/plugins/ec2/InstanceInfo.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public Instance getInstance() {
+        return instance;
+    }
+
+    public void setKeypair(KeyPair keypair) {
+        this.keypair = keypair;
+    }
+
```
jglick, line 25: So you still forgot to clean up AFAICT:

```suggestion
```

src/main/java/hudson/plugins/ec2/InstanceInfo.java
```diff
@@ -0,0 +1,39 @@
...
+import com.amazonaws.services.ec2.model.Instance;
+import com.amazonaws.services.ec2.model.KeyPair;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class InstanceInfo {
+    private final Instance instance;
+    private KeyPair keypair;
```
jglick, line 11:
```suggestion
    private final KeyPair keypair;
```

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -94,17 +95,24 @@ public void terminate() {
...
-                                        LOGGER.info("Terminated EC2 instance (terminated): " + instanceId);
+                                        LOGGER.info("Terminated EC2 spot instance (terminated): " + instanceId);
                                     } catch (AmazonClientException e) {
                                         // Spot request is no longer valid
                                         LOGGER.log(Level.WARNING, "Failed to terminate the Spot instance: " + instanceId, e);
                                     }
                                 }
                             }
-                        } catch (Exception e) {
+                        } catch (AmazonClientException e) {
```
jglick, line 115: Apparently unnecessary change:

```suggestion
                        } catch (Exception e) {
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1081,19 +1117,22 @@ public static URL checkEndPoint(String url) throws FormValidation {
...
     private static SSHUserPrivateKey getSshCredential(String id, ItemGroup context){
 
-        SSHUserPrivateKey credential = CredentialsMatchers.firstOrNull(
-            CredentialsProvider.lookupCredentials(
-                SSHUserPrivateKey.class, // (1)
-                context,
-                null,
-                Collections.emptyList()),
-            CredentialsMatchers.withId(id));
+        if (!id.isEmpty()) {
```
jglick, line 1120: I guess `id` is implicitly `@NonNull` here? Callers that I could see in this PR at least are checking for null first. Some are also checking `isEmpty`. A bit confusing; maybe review call sites and document the method appropriately.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -149,6 +162,7 @@ public abstract class EC2Cloud extends Cloud {
      */
     @CheckForNull
     private String credentialsId;
+
```
jglick, line 165: Unnecessary change. Please review your own diff and check for hunks which do not contribute anything to the change, so that reviewers can focus on the hunks which do.

```suggestion
```

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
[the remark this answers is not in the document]
jglick: Seems fragile. I would expect the serialized state of the agent to indicate whether it has an ephemeral keypair, and if so, its name. Otherwise you are back to the situation where switching from one mode to the other could result in generated keypairs not being deleted automatically. I thought this had been fixed at some point earlier in this PR’s history but it is now regressed?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1134,7 +1174,7 @@ public FormValidation doCheckSshKeysCredentialsId(@AncestorInPath ItemGroup cont
                 return FormValidation.ok();
             }
             if (value == null || value.isEmpty()){
-                return FormValidation.error("No ssh credentials selected");
+                return FormValidation.ok("No SSH credentials selected, so SSH key management will be handled dynamically by Jenkins");
```
jglick, line 1177: It would be helpful in this case in `doTestConnection` to actually try generating and then deleting a keypair, to validate that the IAM principal has the corresponding permissions. Otherwise someone who originally configured the plugin with a static keypair by selecting the minimal EC2 permissions in some policy might try switching to dynamic mode and only see it fail when trying to run builds.

src/main/java/hudson/plugins/ec2/Eucalyptus.java
```diff
@@ -48,7 +49,7 @@ public class Eucalyptus extends EC2Cloud {
     private final URL s3endpoint;
 
     @DataBoundConstructor
-    public Eucalyptus(String name, URL ec2EndpointUrl, URL s3EndpointUrl, boolean useInstanceProfileForCredentials, String credentialsId, String privateKey, String sshKeysCredentialsId, String instanceCapStr, List<SlaveTemplate> templates, String roleArn, String roleSessionName) {
+    public Eucalyptus(String name, URL ec2EndpointUrl, URL s3EndpointUrl, boolean useInstanceProfileForCredentials, String credentialsId, String privateKey, Secret sshPrivateKeySecret, String sshKeysCredentialsId, String instanceCapStr, List<SlaveTemplate> templates, String roleArn, String roleSessionName) {
```
jglick, line 52: AFAICT all the changes to this class should be reverted?

src/main/java/hudson/plugins/ec2/InstanceInfo.java
```diff
@@ -0,0 +1,39 @@
...
+        return keypair;
+    }
+
+
+    public static List<InstanceInfo> fromInstances(List<Instance> instances, EC2Cloud cloud) {
+        try {
+            KeyPair keyPair = cloud.resolveKeyPair();
+            return instances.stream().map(obj -> new InstanceInfo(obj, keyPair)).collect(Collectors.toList());
+        } catch (java.io.IOException e) {
+            throw new RuntimeException(e);
```
jglick, line 36: Would be more natural to just throw `IOException` out of this method and let callers deal with it.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1164,44 +1167,68 @@ private List<EC2AbstractSlave> provisionOndemand(Image image, int number, EnumSe
...
+    List<InstanceInfo> createNewInstances(AmazonEC2 ec2, RunInstancesRequest riRequest) {
+        if (isUsingDynamicSshKeys()) {
+            // no static ssh key defined, so submit riRequest.maxCount requests
+            // each with a unique keypair (instead of a single request for multiple identical instances)
+            LOGGER.fine(() -> "no static ssh credential configured, will use dynamic ssh keys instead");
+            List<InstanceInfo> instances = new ArrayList<>();
+            int maxRequested = riRequest.getMaxCount();
+            riRequest.setMaxCount(1);
+            for (int i=0;i < maxRequested; i++) {
+                String keyName = "cloudbees-" + UUID.randomUUID();
```
jglick, line 1218:
```suggestion
                String keyName = "jenkins-" + UUID.randomUUID();
```

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1164,44 +1167,68 @@ private List<EC2AbstractSlave> provisionOndemand(Image image, int number, EnumSe
...
     }
 
+    List<InstanceInfo> createNewInstances(AmazonEC2 ec2, RunInstancesRequest riRequest) {
+        if (isUsingDynamicSshKeys()) {
+            // no static ssh key defined, so submit riRequest.maxCount requests
+            // each with a unique keypair (instead of a single request for multiple identical instances)
+            LOGGER.fine(() -> "no static ssh credential configured, will use dynamic ssh keys instead");
+            List<InstanceInfo> instances = new ArrayList<>();
+            int maxRequested = riRequest.getMaxCount();
+            riRequest.setMaxCount(1);
```
jglick, line 1216: Looks correct but surprising—you do not normally expect some object like this to be mutated by a subroutine. While I guess `RunInstancesRequest` is just some mutable bean that gets serialized anew for every `runInstances` call, it would be clearer if the caller of this method were to pass in the necessary parameters for a `RunInstancesRequest` (or a `Supplier<RunInstancesRequest>`) so that a fresh request object would be constructed on each API call.

mikecirioli:
i have manually tested switching between dynamic and static ssh keys (using the new `none` option in the credentials dropdown without encountering any issues.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
mikecirioli: AFAICT the setter is calling `fixEmpty()` so i will remove the additional check

jglick:
Other than https://github.com/jenkinsci/ec2-plugin/pull/993#discussion_r1824375122 I do not think I have outstanding blocking comments, though I am not a plugin maintainer and I certainly do not understand all aspects of its operation.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -195,6 +195,9 @@ protected EC2Cloud(String name, boolean useInstanceProfileForCredentials, String
         this.roleSessionName = roleSessionName;
         this.credentialsId = Util.fixEmpty(credentialsId);
         this.sshKeysCredentialsId = Util.fixEmpty(sshKeysCredentialsId);
+        if (this.sshKeysCredentialsId == null) {
+            this.sshKeysCredentialsId = null;
+        }
```
jglick, line 200: Uh, what??

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1116,20 +1120,24 @@ public static URL checkEndPoint(String url) throws FormValidation {
...
-                        null,
-                        Collections.emptyList()),
-                CredentialsMatchers.withId(id));
-
-        if (credential == null) {
-            LOGGER.log(Level.WARNING, "EC2 Plugin could not find the specified credentials ({0}) in the Jenkins Global Credentials Store, EC2 Plugin for cloud must be manually reconfigured", new String[]{id});
-        }
+    private static SSHUserPrivateKey getSshCredential(String id, ItemGroup context){
+
+        if (!id.isEmpty()) {
```
jglick, line 1125: So is `id` still supposed to be `@NonNull`, as implied by the NPE here otherwise?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: PEBKAB

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: yes, let me fix that

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: i forgot to add the check for `isBlank()` - basically i want to ensure that field is null if there is no possible real value (ie. null or empty)

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
mikecirioli: I can take a look at this later today, although i don't promise it will be in this PR

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -195,7 +195,7 @@ protected EC2Cloud(String name, boolean useInstanceProfileForCredentials, String
         this.roleSessionName = roleSessionName;
         this.credentialsId = Util.fixEmpty(credentialsId);
         this.sshKeysCredentialsId = Util.fixEmpty(sshKeysCredentialsId);
-        if (this.sshKeysCredentialsId == null) {
+        if (this.sshKeysCredentialsId != null && this.sshKeysCredentialsId.isBlank()) {
             this.sshKeysCredentialsId = null;
```
jglick, line 199: Remember that this will have no effect on clouds loaded from disk (`$JENKINS_HOME/config.xml`), only new saved clouds. To really enforce null vs. nonblank, you would need to either use `readResolve`, or edit the getter and then consistently use that instead of the field.

mikecirioli: if i implement this in `readResolve` does that make the above code redundant?

mikecirioli:
i am putting this back in draft while i investigate an issue with dynamic keys not being removed after they are used

src/main/java/hudson/plugins/ec2/EC2Cloud.java
jglick: No, because `readResolve` is not called for newly created/resaved clouds; that goes through the constructor.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1496,13 +1497,20 @@ private List<EC2AbstractSlave> provisionSpot(Image image, int number, EnumSet<Pr
...
-                KeyPair kp = getParent().resolveKeyPair();
-                if (kp != null) {
-                    launchSpecification.setKeyName(kp.getKeyName());
+                LOGGER.fine(() -> "static ssh credential is configured, resolving keyname to use in spot launch request");
+                keypair = getParent().resolveKeyPair();
+                if (keypair != null) {
+                    launchSpecification.setKeyName(keypair.getKeyName());
                 }
+            } else {
+                String keyName = "dev-1-" + UUID.randomUUID();
```
jglick, line 1510: Is this supposed to be temporary code?

mikecirioli: yes, i did not mean to commit that change :face_exhaling:

src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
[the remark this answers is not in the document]
mikecirioli: i think its the result of the changes i did to always store the key  (regardless of using dynamic keys) - i should be able to unwind that and resolve this comment

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
mikecirioli: not implemented

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -1731,17 +1825,18 @@ private DescribeSecurityGroupsResult getSecurityGroupsBy(String filterName, Set<
     /**
      * Provisions a new EC2 agent based on the currently running instance on EC2, instead of starting a new one.
      */
-    public EC2AbstractSlave attach(String instanceId, TaskListener listener) throws AmazonClientException, IOException {
-        PrintStream logger = listener.getLogger();
+    public EC2AbstractSlave attach(String instanceId, TaskListener listener) throws Exception {
         AmazonEC2 ec2 = getParent().connect();
 
+        if (isUsingDynamicSshKeys()) {
+            throw new Exception("not allowed when using dynamic ssh keypair management");
```
mikecirioli, line 1832: what is a better exception to throw here?

PR #997, opened 2024-10-25
Replace dropdown button to remove dependency on YUI

zbynek:
Fixes https://issues.jenkins.io/browse/JENKINS-73541

Replace YUI button with a more modern component, analogous to https://github.com/jenkinsci/credentials-plugin/pull/551 as part of the YUI removal initiative (https://issues.jenkins.io/browse/JENKINS-73539 )

### Testing done

Manually checked that clicking the dropdown button goes to the provision page with the correct parameters.

image

### Submitter checklist
- [n/a] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/resources/hudson/plugins/ec2/EC2Cloud/computerSet.jelly
+++ b/src/main/resources/hudson/plugins/ec2/EC2Cloud/computerSet.jelly
@@ -22,28 +22,30 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 -->
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler"
+         xmlns:l="/lib/layout" xmlns:dd="/lib/layout/dropdowns" xmlns:f="/lib/form">
   <j:if test="${it.hasPermission(it.PROVISION)}">
     <tr>
       <td />
       <td colspan="${monitors.size()+1}">
+        <l:overflowButton icon="symbol-add"
+                          text="${%Provision via} ${it.displayName}"
+                          tooltip="${null}"
+                          clazz="jenkins-!-margin-top-2">
+          <j:forEach var="t" items="${it.templates}">
+            <dd:custom>
+              <button class="jenkins-dropdown__item"
+                      data-type="ec2-provision"
+                      data-url="${t.description}">
+                ${t.displayName}
+              </button>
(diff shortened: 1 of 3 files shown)


src/main/resources/hudson/plugins/ec2/EC2Cloud/provision.js
```diff
@@ -0,0 +1,8 @@
+Behaviour.specify("[data-type='ec2-provision']", 'ec2-provision', -99, function(e) {
+  e.addEventListener("click", function (event) {
+    const form = document.querySelector("form[name='provision']");
```
mawinter69, line 3: That query is problematic I think. The same query is used for the google-compute-engine-plugin and if both plugins are installed and configured, then it potentially chooses the wrong one. Ideally the query only select the form from the surrounding `td`

mawinter69 (changes requested):

src/main/resources/hudson/plugins/ec2/EC2Cloud/provision.js
zbynek: you're right, also with just this plugin you can have multiple EC2 clouds. I'll fix that.

mawinter69 (approved):

src/main/resources/hudson/plugins/ec2/EC2Cloud/computerSet.jelly
```diff
@@ -22,28 +22,32 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+              <button class="jenkins-dropdown__item"
+                      data-type="ec2-provision"
+                      data-form="${formId}"
+                      data-url="${t.description}">
+                ${t.displayName}
+              </button>
+            </dd:custom>
+          </j:forEach>
+        </l:overflowButton>
+        <f:form action="${rootURL}/${it.url}/provision" method="post" name="provision" id="${formId}">
```
timja, line 47: you could probably move it so the form is a parent of the j:forEach  (like it was before roughly)  and then use something like`e.closest('form')` to save you having to generate Ids.

(non-blocking)

timja (approved):

src/main/resources/hudson/plugins/ec2/EC2Cloud/computerSet.jelly
zbynek: I think that doesn't work because of how `overflowButton` works: the elements in `foreach` are added to `<template>` which is the cloned into a popup that's no longer part of the `form` but floats above it.

It should be possible to remove the form and use `fetch` like the OpenStack plugin does (https://github.com/jenkinsci/openstack-cloud-plugin/blob/fc5384db4d5f1f4f2ebdec2aec7e28ce312cf0df/plugin/src/main/resources/jenkins/plugins/openstack/compute/JCloudsCloud/computerSet.jelly#L23), but I didn't want to change the behavior.

res0nance (approved):

PR #1003, opened 2024-11-07
allow using a file based ssh credential via system property to avoid having to create a Jenkins ssh credential

mikecirioli:
For some use cases it may not be desirable to create a Jenkins ssh credential to use for connecting to agents, as this credential may be visible to users in the ux, or you may find it easier to manage this private key as a mounted k8s secret.  

To enable this functionality simply define a new `System.property` as follows:
`SSH_KEY_PAIR_PRIVATE_KEY_FILE=/path/to/plaintext/private_key.pem`

If this system property is defined then any credential configured in the cloud config page will be ignored. 

### Testing done

- [ ]  Manually tested and works as expected in my k8s test cluster. 
- [ ] fixup 1 failing unit test
- [ ] unit test for new functionality
 

### Submitter checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,14 +52,25 @@ the main "Manage Jenkins" \> "Configure System" page, and scroll down
 near the bottom to the "Cloud" section. There, you click the "Add a new
 cloud" button, and select the "Amazon EC2" option. This will display the
 UI for configuring the EC2 plugin.  Then enter the Access Key and Secret
-Access Key which act like a username/password (see IAM section). Because
-of the way EC2 works, you also need to have an RSA private key that the
+Access Key which act like a username/password (see IAM section). 
+
+Because of the way EC2 works, you also need to have an RSA private key that the
 cloud has the other half for, to permit sshing into the instances that
 are started. Please use the AWS console or any other tool of your choice
-to generate the private key to interactively log in to EC2 instances.
+to generate the private key to interactively log in to EC2 instances. 
 
-Once you have put in your Access Key and Secret Access Key, select a
-region for the cloud (not shown in screenshot). You may define only one
+Once you have generated the needed private key you must either store it as
+a Jenkins `SSH Private Key` credential (and select that credential in your cloud
+config).
+
+If you do not want to create a new Jenkins credential you may alterantively store it
+in plain text on disk, indicating its file path via the Jenkins system property
+`SSH_KEY_PAIR_PRIVATE_KEY_FILE`.  If this system property has a non-empty value then
+it will override the ssh credential specified in the cloud configuration page.  This
+approach works well for `k8s` secrets that are mounted in a jenkins container for example.
+
+Once you have put in your Access Key, Secret Access Key, and configured an ssh private key
+select a region for the cloud (not shown in screenshot). You may define only one
 cloud for each region, and the regions offered in the UI will show only
 the regions that you don't already have clouds defined for them.
 
--- a/src/main/java/hudson/plugins/ec2/EC2Cloud.java
+++ b/src/main/java/hudson/plugins/ec2/EC2Cloud.java
@@ -87,6 +87,9 @@ import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
@@ -126,6 +129,10 @@ public abstract class EC2Cloud extends Cloud {
     public static final String EC2_REQUEST_EXPIRED_ERROR_CODE = "RequestExpired";
 
     private static final SimpleFormatter sf = new SimpleFormatter();
+
+    // if this system property is defined and its value points to a valid ssh private key on disk
+    // then this will be used instead of any configured ssh credential
+    private static final String SSH_KEY_PAIR_PRIVATE_KEY_FILE = "SSH_KEY_PAIR_PRIVATE_KEY_FILE";
 
     private transient ReentrantLock slaveCountingLock = new ReentrantLock();
 
@@ -195,10 +202,27 @@ public abstract class EC2Cloud extends Cloud {
 
     @CheckForNull
     public EC2PrivateKey resolvePrivateKey(){
-        if (sshKeysCredentialsId != null) {
+        if (!System.getProperty(SSH_KEY_PAIR_PRIVATE_KEY_FILE, "").isEmpty()) {
+            LOGGER.fine(() -> "(resolvePrivateKey) secret key file configured, will load from disk");
+            return fetchPrivateKeyFromDisk();
+        } else if (sshKeysCredentialsId != null) {
+            LOGGER.fine(() -> "(resolvePrivateKey) Using jenkins ssh credential");
             SSHUserPrivateKey privateKeyCredential = getSshCredential(sshKeysCredentialsId, Jenkins.get());
             if (privateKeyCredential != null) {
                 return new EC2PrivateKey(privateKeyCredential.getPrivateKey());
+            }
+        }
+        return null;
+    }
+
+    private static EC2PrivateKey fetchPrivateKeyFromDisk()  {
+        String filename = System.getProperty(SSH_KEY_PAIR_PRIVATE_KEY_FILE, "");
+        if (!filename.isEmpty()) {
+            try {
+                return new EC2PrivateKey(new String(Files.readAllBytes(Paths.get(filename)), StandardCharsets.UTF_8));
+            } catch (IOException e) {
+                LOGGER.warning(() -> "unable to read private key from file " + filename);
+                return null;
             }
         }
         return null;
@@ -1122,6 +1146,7 @@ public abstract class EC2Cloud extends Cloud {
             AbstractIdCredentialsListBoxModel result = new StandardListBoxModel();
             if (Jenkins.get().hasPermission(Jenkins.ADMINISTER)) {
                 result = result
+                        .includeEmptyValue()
                         .includeMatchingAs(Jenkins.getAuthentication(), context, SSHUserPrivateKey.class, Collections.<DomainRequirement>emptyList(), CredentialsMatchers.always())
                         .includeMatchingAs(ACL.SYSTEM, context, SSHUserPrivateKey.class, Collections.<DomainRequirement>emptyList(), CredentialsMatchers.always())
                         .includeCurrentValue(sshKeysCredentialsId);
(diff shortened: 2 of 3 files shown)


src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -127,6 +130,10 @@ public abstract class EC2Cloud extends Cloud {
 
     private static final SimpleFormatter sf = new SimpleFormatter();
 
+    // if this system property is defined and its value points to a valid ssh private key on disk
+    // then this will be used instead of any configured ssh credential
+    private static final String SSH_KEY_PAIR_PRIVATE_KEY_FILE = "SSH_KEY_PAIR_PRIVATE_KEY_FILE";
```
jglick, line 135: Conventionally, system property used in Jenkins follow a format of `SomeClass.class.getName() + ".camelCase"`. This looks like an environment variable.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -204,6 +215,19 @@ public EC2PrivateKey resolvePrivateKey(){
         return null;
     }
 
+    private static EC2PrivateKey fetchPrivateKeyFromDisk()  {
+        String filename = System.getProperty(SSH_KEY_PAIR_PRIVATE_KEY_FILE, "");
+        if (!filename.isEmpty()) {
+            try {
+                return new EC2PrivateKey(new String(Files.readAllBytes(Paths.get(filename)), StandardCharsets.UTF_8));
```
jglick, line 222: `Files.readString`?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -204,6 +215,19 @@ public EC2PrivateKey resolvePrivateKey(){
...
+            } catch (IOException e) {
+                LOGGER.warning(() -> "unable to read private key from file " + filename);
```
jglick, line 224: You should include the stack trace. It may be relevant.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1135,16 +1160,22 @@ public FormValidation doCheckSshKeysCredentialsId(@AncestorInPath ItemGroup cont
                 // Don't do anything if the user is only reading the configuration
                 return FormValidation.ok();
             }
-            if (value == null || value.isEmpty()){
-                return FormValidation.error("No ssh credentials selected");
-            }
 
-            SSHUserPrivateKey sshCredential = getSshCredential(value, context);
             String privateKey = "";
```
jglick, line 1164:
```suggestion
            String privateKey;
```
?

mikecirioli: that was my bad, i just shifted the original code around and didn't catch that

mikecirioli:
feedback addressed @jglick , i am fixing  up the testing now

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -216,12 +216,12 @@ public EC2PrivateKey resolvePrivateKey(){
...
     private static EC2PrivateKey fetchPrivateKeyFromDisk()  {
-        String filename = System.getProperty(SSH_KEY_PAIR_PRIVATE_KEY_FILE, "");
+        String filename = System.getProperty(sshPrivateKeyFilePath, "");
         if (!filename.isEmpty()) {
             try {
-                return new EC2PrivateKey(new String(Files.readAllBytes(Paths.get(filename)), StandardCharsets.UTF_8));
+                return new EC2PrivateKey(Files.readString(Paths.get(filename), StandardCharsets.UTF_8));
             } catch (IOException e) {
-                LOGGER.warning(() -> "unable to read private key from file " + filename);
+                LOGGER.warning(() -> "unable to read private key from file " + filename + ": " + e);
```
jglick, line 224: just
```suggestion
                LOGGER.warning(e, () -> "unable to read private key from file " + filename);
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -204,6 +215,19 @@ public EC2PrivateKey resolvePrivateKey(){
         return null;
     }
 
+    private static EC2PrivateKey fetchPrivateKeyFromDisk()  {
```
amuniz, line 218:
```suggestion
    @CheckForNull
    private static EC2PrivateKey fetchPrivateKeyFromDisk()  {
```

README.md
```diff
@@ -52,14 +52,25 @@ the main "Manage Jenkins" \> "Configure System" page, and scroll down
...
-to generate the private key to interactively log in to EC2 instances.
+to generate the private key to interactively log in to EC2 instances. 
+
+Once you have generated the needed private key you must either store it as
+a Jenkins `SSH Private Key` credential (and select that credential in your cloud
+config).
+
+If you do not want to create a new Jenkins credential you may alterantively store it
+in plain text on disk, indicating its file path via the Jenkins system property
+`SSH_KEY_PAIR_PRIVATE_KEY_FILE`.  If this system property has a non-empty value then
```
fcojfernandez, line 68: IIUC
```suggestion
`hudson.plugins.ec2.EC2Cloud.sshPrivateKeyFilePath`.  If this system property has a non-empty value then
```

amuniz (approved):
LGTM

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -204,6 +215,25 @@ public EC2PrivateKey resolvePrivateKey(){
...
     }
 
+    /* visible for testing */
+    @CheckForNull
+    public static EC2PrivateKey fetchPrivateKeyFromDisk() {
+        return fetchPrivateKeyFromDisk(System.getProperty(SSH_PRIVATE_KEY_FILEPATH, ""));
+    }
+
+    @CheckForNull
+    public static EC2PrivateKey fetchPrivateKeyFromDisk(String filepath) {
```
fcojfernandez, line 225: Please, move these two methods to `EC2PrivateKey` where they make more sense. And doing so, you could even simplify the name: `fetchFromDisk` as it is clear when calling `EC2PrivateKey.fetchFromDisk`

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1122,6 +1152,7 @@ public ListBoxModel doFillSshKeysCredentialsIdItems(@AncestorInPath ItemGroup co
             AbstractIdCredentialsListBoxModel result = new StandardListBoxModel();
             if (Jenkins.get().hasPermission(Jenkins.ADMINISTER)) {
                 result = result
+                        .includeEmptyValue()
```
fcojfernandez, line 1155: What if now a user select the doesn't select a value? How is configured the cloud? Should it have a FormValidation?

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -204,6 +215,25 @@ public EC2PrivateKey resolvePrivateKey(){
...
 
+    /* visible for testing */
+    @CheckForNull
+    public static EC2PrivateKey fetchPrivateKeyFromDisk() {
+        return fetchPrivateKeyFromDisk(System.getProperty(SSH_PRIVATE_KEY_FILEPATH, ""));
+    }
+
+    @CheckForNull
+    public static EC2PrivateKey fetchPrivateKeyFromDisk(String filepath) {
+        if (!(filepath == null) && !(filepath.isEmpty())) {
```
fcojfernandez, line 226: Simpler
```suggestion
        if (StringUtils.isNotEmpty(filepath)) {
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1135,16 +1166,26 @@ public FormValidation doCheckSshKeysCredentialsId(@AncestorInPath ItemGroup cont
...
+                } else {
+                    return FormValidation.error("Failed to find credential \"" + value + "\" in store.");
+                }
             } else {
-                return FormValidation.error("Failed to find credential \"" + value + "\" in store.");
+                EC2PrivateKey k = fetchPrivateKeyFromDisk();
+                if (k == null) {
+                    return FormValidation.error("Failed to find private key file " + System.getProperty(SSH_PRIVATE_KEY_FILEPATH));
+                }
+                privateKey = k.getPrivateKey();
```
fcojfernandez, line 1188: Let's suppose a user is setting the System property AND selecting a credentials, which is even valid as it was in use previously. This message might make the user get confused as the credentials selected is valid.

This whole method now should add a warning as well everytime that the credentials will be overridden by the system property. `FormValidation.aggregate` will do it

src/test/java/hudson/plugins/ec2/EC2CloudTest.java
```diff
@@ -27,6 +27,11 @@
 @RunWith(MockitoJUnitRunner.class)
 public class EC2CloudTest {
 
+    @Test
+    public void testFileBasedSShKey() {
+        assertTrue("file content should not have been empty", EC2Cloud.fetchPrivateKeyFromDisk(getClass().getClassLoader().getResource("hudson/plugins/ec2/test.pem").getPath()) != null);
```
fcojfernandez, line 32: please, use `assertNull` or `assertNotNull` if that's what you want to check. If the test fails for whatever reason, although we get the `file content should not have been empty` message, the reason of the failure would be `expected true but was false. `assertTrue` and `assertFalse` should be used only to assert boolean values

fcojfernandez (commented):
would it be possible to have more tests? Something to test what prevails if the user includes the credentials but the system property is overriding the value

src/main/java/hudson/plugins/ec2/EC2Cloud.java
mikecirioli: it already does

fcojfernandez: true, I just didn't connect this with the validation

```
                if (value == null || value.isEmpty()) {
                    return FormValidation.error("No ssh credentials selected");
                }
```

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1163,6 +1183,14 @@ public FormValidation doCheckSshKeysCredentialsId(@AncestorInPath ItemGroup cont
             if (!hasEnd)
                 return FormValidation
                         .error("The private key is missing the trailing 'END RSA PRIVATE KEY' marker. Copy&paste error?");
+
+            if (System.getProperty(SSH_PRIVATE_KEY_FILEPATH, "").isEmpty()) {
```
amuniz, line 1187: Are you missing a `!` here?

mikecirioli:
added screenshot and validated manually again as awell @fcojfernandez

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1135,18 +1144,31 @@ public FormValidation doCheckSshKeysCredentialsId(@AncestorInPath ItemGroup cont
...
+                } else {
+                    return FormValidation.error("Failed to find credential \"" + value + "\" in store.");
+                }
             } else {
-                return FormValidation.error("Failed to find credential \"" + value + "\" in store.");
+                EC2PrivateKey k = EC2PrivateKey.fetchFromDisk();
+                if (k == null) {
+                    return FormValidation.error("Failed to find private key file " + System.getProperty(SSH_PRIVATE_KEY_FILEPATH));
+                }
+                privateKey = k.getPrivateKey();
```
fcojfernandez, line 1167: still missing the aggregate here

car-roll (approved):

fcojfernandez (approved):

PR #998, opened 2024-10-28
support fail fast for node removal

car-roll:
https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/372 added fail fast for cloud node removal. However, this feature [is not supported](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/372#discussion_r1795825003) by the `ec2` plugin.

This PR refactors `EC2AbstractSlave` and `EC2Computer` in order to take advantage of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/372 

### Testing done

- [ ] Perform manual testing by killing spot instance manually and watching startup

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #1015, opened 2024-12-02
Use AbstractCloudSlave and AbstractCloudComputer

car-roll:
supersedes #998 

### Testing done

- [ ] Manual testing

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
@@ -4,9 +4,11 @@ import com.amazonaws.AmazonClientException;
 import com.amazonaws.services.ec2.AmazonEC2;
 import com.amazonaws.services.ec2.model.*;
 import hudson.Extension;
+import hudson.Functions;
 import hudson.model.Computer;
 import hudson.model.Descriptor.FormException;
 import hudson.model.Node;
+import hudson.model.TaskListener;
 import hudson.plugins.ec2.ssh.EC2MacLauncher;
 import hudson.plugins.ec2.ssh.EC2UnixLauncher;
 import hudson.plugins.ec2.win.EC2WindowsLauncher;
@@ -435,7 +437,7 @@ public class EC2OndemandSlave extends EC2AbstractSlave {
      * Terminates the instance in EC2.
      */
     @Override
-    public void terminate() {
+    protected void _terminate(TaskListener listener) throws IOException, InterruptedException {
         if (terminateScheduled.getCount() == 0) {
             synchronized (terminateScheduled) {
                 if (terminateScheduled.getCount() == 0) {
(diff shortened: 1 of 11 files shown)


src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -456,7 +458,8 @@ public void terminate() {
                             Jenkins.get().removeNode(this);
                             LOGGER.info("Removed EC2 instance from jenkins controller: " + getInstanceId());
```
jglick, line 459: Some of these `INFO` log messages could probably be replaced with messages to the `TaskListener`.

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -196,7 +198,8 @@ public void terminate() {
                                 LOGGER.info("Cancelled Spot request: " + spotInstanceRequestId);
```
jglick, line 198: ditto here

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -214,15 +217,14 @@ public void terminate() {
...
+                                                listener.error("Failed to terminate the Spot instance: " + instanceId));
                                     }
                                 }
                             }
                         } catch (Exception e) {
-                            LOGGER.log(Level.WARNING, "Failed to remove agent: ", e);
+                            Functions.printStackTrace(e, listener.error("Failed to remove agent"));
                         } finally {
                             // Remove the instance even if deletion failed, otherwise it will hang around forever in
                             // the nodes page. One way for this to occur is that an instance was terminated
```
jglick, line 230: Pay attention to https://github.com/jenkinsci/jenkins/blob/4248fc0ca6ce108364ed23208063ff20753f5e70/core/src/main/java/hudson/slaves/AbstractCloudSlave.java#L88-L94 which means that after switching from`terminate` to `_terminate` it is no longer your responsibility to `removeNode`.

src/main/java/hudson/plugins/ec2/EC2OndemandSlave.java
```diff
@@ -456,7 +458,8 @@ public void terminate() {
                             Jenkins.get().removeNode(this);
```
jglick, line 458: ditto here

src/main/java/hudson/plugins/ec2/EC2SpotSlave.java
```diff
@@ -178,7 +180,7 @@ protected boolean isAlive(boolean force) {
      * Cancel the spot request for the instance. Terminate the instance if it is up. Remove the agent from Jenkins.
      */
     @Override
-    public void terminate() {
+    protected void _terminate(TaskListener listener) throws IOException, InterruptedException {
```
jglick, line 183: The impl seems mostly the same as in `EC2OndemandSlave`. Is there a reason the method cannot simply be pulled up into `EC2AbstractSlave`.

PR #1021, opened 2024-12-17
[JENKINS-73640] Use AWS SDK for Java 2.x

basil:
See [JENKINS-73640](https://issues.jenkins.io/browse/JENKINS-73640). Builds on the work started by @dblanchette to migrate this plugin to AWS SDK for Java 2.x. The main improvement over the initial contribution is that I have fixed all integration tests so that `mvn clean verify` passes. I have not attempted to test this plugin against a real AWS account. This diff still needs some more work to retain compatibility with older CasC configurations. Currently, support for older CasC configurations is broken because the names of the enumeration are different for instance types.

### Testing done

`mvn clean verify`

### Submitter checklist


src/test/resources/hudson/plugins/ec2/WindowsData.yml
```diff
@@ -9,7 +9,7 @@ jenkins:
           - description:
             ami: "ami-abc123"
             labelString: "windows vs2019"
-            type: "M5aLarge"
+            type: "M5_A_LARGE"
```
basil, line 12: @jetersen @timja Any suggestions on how to retain compatibility with the old enumeration names for CasC purposes? This is an interesting case because the new `software.amazon.awssdk.services.ec2.model.InstanceType` is an enumeration, just like the old `com.amazonaws.services.ec2.model.InstanceType`, just using upper snake case rather than Pascal case.

timja: Not really other than that you map old to new yourself and potentially a new field name and deprecate the old one.

dblanchette: I'm still working on it, but what I'm trying to achieve on my side is to accept strings and then resolve from the text value of the Enum.
Both enums in the old and new SDKs resolve to the actual API name
e.g. `t2.large` is `T2_LARGE("t2.large"),` in the new one and `T2Large("t2.large"),` in the old one.
We would then accept either `T2Large` (legacy) or `t2.large`.
This has the benefit of removing the implementation detail leaking into the choices in the UI while allowing the old configs to still be loaded

timja: That should work fine, best to use the name that users see in the UI / AWS documentation. If that's `t2.large` then great, (and supporting the old format)

basil: I also worry that we might need to do the same thing for data serialized to XML via XStream.

dblanchette: I'm not too sure how that works, if it uses the same `@DataBoundConstructor`, my fix could work.

Here are my changes to allow loading with the old and new instance type string: https://github.com/coveord/ec2-plugin/tree/fix/JENKINS-73640 (I made the branch on top of yours, so it should merge cleanly)

dblanchette: Just tested:
- Checked out master
- Ran a test instance
- Added a configuration with instance type `T3Micro`
- Stopped the test instance
- Checked out my branch
- Ran a test instance
Unfortunately, the instance type is now the first one in the list `a1.2xlarge`. Will try to figure out why

dblanchette: Fixed in [my branch](https://github.com/coveord/ec2-plugin/tree/fix/JENKINS-73640), it was only a UI thing

So the test goes:
- Checked out master
- Ran a test instance
- Added a configuration with instance type T3Micro
  - In `work/config.xml`, the instance type is saved as `T3Micro`
- Stopped the test instance
- Checked out my branch
- Ran a test instance
- Go to the same cloud configuration
  - Now, `t3.micro` is selected
- Save the new configuration
  - In `work/config.xml`, the instance type is now saved as `t3.micro`
- Reload the cloud configuration page
  - `t3.micro` is still selected

basil: Thanks @dblanchette, I added a variation of your code to my branch. Are you interested in testing the latest incremental build with a real EC2 account?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -2645,6 +2830,10 @@ protected Object readResolve() {
             amiType = new UnixData(rootCommandPrefix, slaveCommandPrefix, slaveCommandSuffix, sshPort, null);
         }
 
+        if (type != null && !type.isEmpty()) {
+            type = (new InstanceTypeCompat(type)).getInstanceType().toString();
+        }
```
basil, line 2835: Compatibility with serialized XStream XML.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -343,7 +347,9 @@ public SlaveTemplate(
         this.securityGroups = securityGroups;
         this.remoteFS = remoteFS;
         this.amiType = amiType;
-        this.type = type;
+        this.type = type != null && !type.isEmpty()
+                ? (new InstanceTypeCompat(type)).getInstanceType().toString()
+                : null;
```
basil, line 352: Compatibility with CasC.

src/test/resources/hudson/plugins/ec2/WindowsData.yml
dblanchette: I have a former production instance on which I can install it, but I'm not sure when I'll have time. Thanks for integrating my changes!

cr3ative:
Many thanks to all approaching this task - to our surprise, AWS isn't adding any new instance types to the V1 SDK despite it being in "maintenance mode" until the end of the year. When this plugin is ready to move to V2, we'll be able to use Graviton4 instances, which we're eagerly anticipating.

danbf:
Thanks very much for taking this on.  This is a super useful plugin for those of us running our CI on AWS.

basil:
> Thanks very much for taking this on.

@danbf Thank you for your kind words, but to clarify—this ticket is unassigned, and I haven't taken this on. I am simply enabling members of the Jenkins community to test this change by providing them with an incremental build.

hazmeister:
We've been running Version1878.vb_3312ec51e84 for two weeks without any issues on our jenkins master.  This has allowed us to start using the newer **i7ie.xlarge** instances for that 4Ghz CPU for our builds.  We spin nodes up on demand and turn off after idling for a minute.

src/test/resources/hudson/plugins/ec2/Mac.yml
```diff
@@ -9,6 +9,6 @@ jenkins:
           - description:
             ami: "ami-5678"
             labelString: "mac clear"
-            type: "Mac1Metal"
+            type: "mac1.metal"
```
res0nance, line 12: Is this a breaking change?

basil: @res0nance No, it is not. Old values are still accepted.

res0nance (approved):

PR #1027, opened 2024-12-21
Remove Eucalyptus support

basil:
### Motivation

The AWS SDK for Java v2 has some changes in how EC2 clients are constructed with regard to the region, as described [here](https://issues.jenkins.io/browse/JENKINS-73640?focusedId=451456&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-451456). It is not clear how to implement these changes for Eucalyptus clouds, which do not have a region. After thinking about this some more, I decided it would be simpler to remove support for Eucalyptus clouds. This solves the problem about how to adapt to v2 API changes by eliminating the problem for Eucalyptus clouds, leaving only Amazon EC2 clouds for which the solution is straightforward.

The last commit to https://github.com/eucalyptus/eucalyptus was 7 years ago, and the entire GitHub organization was archived on August 30, 2024. This technology seems dead. I could find no references in open PRs after 2013, except for developers working on routine maintenance wondering how to deal with this dead code (usually guessing at the changes needed, or making changes without testing). The last references I can find in Jira are from 2016/2017, which roughly matches the timeline [here](https://en.wikipedia.org/wiki/Eucalyptus_(software)) showing that the last release of Eucalyptus was in 2018. Since this was over 6 years ago, I can't imagine anyone still using this technology today, and if they are then it seems like lack of support in the Jenkins EC2 plugin would be the least of their problems.

### Problem

Eucalyptus support is dead code.

### Solution

[Flense it](https://en.wikipedia.org/wiki/Flensing).

### Implementation

Delete all Eucalyptus-related code. With that out of the way, flatten the class hierarchy by deprecating `AmazonEC2Cloud` (now the only type of `EC2Cloud` again, as it was before Eucalyptus support was introduced) and pushing up all its functionality into `EC2Cloud`. This restores the status quo before Eucalyptus support was added.

Preserve compatibility for serialized XML and JCasC. For XML, update the XStream alias to add `EC2Cloud` as an implementation so that deserialized instances of `EC2Cloud` are `EC2Cloud` rather than the deprecated `AmazonEC2Cloud` (which is also needed for JCasC export to work, since the descriptor is now part o `EC2Cloud`). Retain the alias, though, so that instances of `AmazonEC2Cloud` that are created via Groovy scripts still get serialized to the same `EC2Cloud` class as before this PR. Add `@Symbol("amazonEC2")` so that JCasC import/export works as before with no changes to import or export.

### Testing done

Confirmed that JCasC export had the same symbols as before. Confirmed that XML from the old version was readable and that new clouds saved with the changes from this PR were saved with the same XML as before this PR.

Confirmed that Groovy scripts with the old types still worked using the deprecated constructors and were serialized into the same format as before this PR.

Ran JCasC tests with these changes and https://github.com/jenkinsci/configuration-as-code-plugin/pull/2611 successfully.


README.md
```diff
@@ -363,7 +362,7 @@ def slaveTemplateUsEast1Parameters = [
   nodeProperties:                null
 ]
 
-def AmazonEC2CloudParameters = [
+def EC2CloudParameters = [
```
basil, line 365: Tested this Groovy scripts after this PR with both the old and the new class names. Confirmed XStream serialization was the same as before.

src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -23,53 +23,13 @@
...
-     * Represents the region. Can be null for backward compatibility reasons.
-     */
-    private String region;
-
-    private String altEC2Endpoint;
-
-    private boolean noDelayProvisioning;
-
-    @DataBoundConstructor
     public AmazonEC2Cloud(
```
basil, line 33: Preserved for binary compatibility with Groovy scripts.

src/main/java/hudson/plugins/ec2/AmazonEC2Cloud.java
```diff
@@ -85,16 +45,15 @@ public AmazonEC2Cloud(
...
                 sshKeysCredentialsId,
                 instanceCapStr,
                 templates,
                 roleArn,
                 roleSessionName);
-        this.region = region;
     }
 
-    @Deprecated
     public AmazonEC2Cloud(
```
basil, line 57: Preserved for binary compatibility with Groovy scripts.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -127,7 +136,7 @@
  *
  * @author Kohsuke Kawaguchi
  */
-public abstract class EC2Cloud extends Cloud {
+public class EC2Cloud extends Cloud {
```
basil, line 139: Restoring the status quo before Eucalyptus support was added.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -189,12 +198,23 @@ public abstract class EC2Cloud extends Cloud {
...
     private transient KeyPair usableKeyPair;
 
+    /**
+     * Represents the region. Can be null for backward compatibility reasons.
+     */
+    private String region;
+
+    private String altEC2Endpoint;
+
+    private boolean noDelayProvisioning;
```
basil, line 208: Pushed up from `AmazonEC2Cloud` with no changes.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -260,9 +306,63 @@ public EC2PrivateKey resolvePrivateKey() {
         return null;
     }
 
-    public abstract URL getEc2EndpointUrl() throws IOException;
+    /**
```
basil, line 309: Pushed up from `AmazonEC2Cloud` with no changes.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -636,15 +736,9 @@ private int countCurrentEC2SpotSlaves(SlaveTemplate template, String jenkinsServ
...
-                sirs = sirResp.getSpotInstanceRequests();
-                dsir.setNextToken(sirResp.getNextToken());
-            } catch (Exception ex) {
-                // Some ec2 implementations don't implement spot requests (Eucalyptus)
-                LOGGER.log(Level.FINEST, "Describe spot instance requests failed", ex);
-                break;
-            }
+            sirResp = connect().describeSpotInstanceRequests(dsir);
+            sirs = sirResp.getSpotInstanceRequests();
+            dsir.setNextToken(sirResp.getNextToken());
```
basil, line 741: Review with "Hide Whitespace".

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1075,7 +1169,12 @@ public boolean canProvision(Label label) {
     }
 
     protected AWSCredentialsProvider createCredentialsProvider() {
-        return createCredentialsProvider(useInstanceProfileForCredentials, credentialsId);
+        return createCredentialsProvider(
+                isUseInstanceProfileForCredentials(),
+                getCredentialsId(),
+                getRoleArn(),
+                getRoleSessionName(),
+                getRegion());
```
basil, line 1177: Pushed up from `AmazonEC2Cloud` with no changes.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1385,27 +1491,25 @@ public FormValidation doCheckSshKeysCredentialsId(
...
          * @param roleSessionName
-         * @param region
          * @return the validation result
          * @throws IOException
          * @throws ServletException
          */
-        @POST
-        protected FormValidation doTestConnection(
+        @RequirePOST
+        public FormValidation doTestConnection(
```
basil, line 1505: Merging this method with the version in `AmazonEC2Cloud` which just did some prep work and then delegated to this method.

src/main/java/hudson/plugins/ec2/EC2Cloud.java
```diff
@@ -1484,6 +1592,75 @@ public ListBoxModel doFillCredentialsIdItems(@AncestorInPath ItemGroup context)
                             Collections.emptyList(),
                             CredentialsMatchers.always());
         }
+
+        @POST
+        public FormValidation doCheckCloudName(@QueryParameter String value) {
```
basil, line 1597: Pushed up from `AmazonEC2Cloud` with no changes.

src/main/java/hudson/plugins/ec2/PluginImpl.java
```diff
@@ -77,7 +77,7 @@ public String getDisplayName() {
     @Override
     public void postInitialize() throws IOException {
         // backward compatibility with the legacy class name
-        Jenkins.XSTREAM.alias("hudson.plugins.ec2.EC2Cloud", AmazonEC2Cloud.class);
+        Jenkins.XSTREAM.alias("hudson.plugins.ec2.EC2Cloud", AmazonEC2Cloud.class, EC2Cloud.class);
```
basil, line 80: Serialize `AmazonEC2Cloud` to `EC2Cloud`, and deserialize `EC2Cloud` to `EC2Cloud` rather than `AmazonEC2Cloud`.

There is a minor issue with this: if creating an `AmazonEC2Cloud` with a Groovy script, it will be serialized to `EC2Cloud` but remain in memory as `AmazonEC2Cloud` until the next Jenkins restart, which will also means JCasC export won't work until the Jenkins restart. This seems acceptable, especially because any user who cares about this behavior can always fix the problem by migrating to `EC2Cloud` in their Groovy script.

src/test/java/hudson/plugins/ec2/EC2CloudUnitTest.java
```diff
@@ -0,0 +1,453 @@
...
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.stubbing.Answer;
+
+/**
+ * Unit tests related to {@link EC2Cloud}, but do not require a Jenkins instance.
+ */
+@RunWith(MockitoJUnitRunner.Silent.class)
+public class EC2CloudUnitTest {
```
basil, line 65: Not new code, but rather a combination of the old `EC2Test` and the old `AmazonEC2CloudUnitTest`.

src/test/java/hudson/plugins/ec2/EC2CloudTest.java
```diff
@@ -1,35 +1,73 @@
...
-import org.mockito.invocation.InvocationOnMock;
-import org.mockito.junit.MockitoJUnitRunner;
-import org.mockito.stubbing.Answer;
+import org.xml.sax.SAXException;
 
-@RunWith(MockitoJUnitRunner.class)
+/**
+ * @author Kohsuke Kawaguchi
+ */
 public class EC2CloudTest {
```
basil, line 61: The equivalent of the old `AmazonEC2CloudTest`. The tests that previously lived in this class have been moved to `EC2CloudUnitTest` (along with the other tests that used to be in `AmazonEC2CloudUnitTest`.

Vlatombe (approved):
LGTM

PR #1044, opened 2025-02-07
Remove deprecations and use Java 17 language features

strangelookingnerd:
## Require Jenkins 2.479.1 and Jakarta EE 9

Jenkins 2.479.1 provides Jakarta EE 9, Eclipse Jetty 12, Spring Security 6, and Java 17.

* Update plugin pom and baseline
* Remove deprecations
* Use Java 17 language features

### Testing done

Confirmed that tests pass.

### Submitter checklist
* [x]  Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
* [x]  Ensure that the pull request title represents the desired changelog entry
* [x]  Please describe what you did
* [x]  Link to relevant issues in GitHub or Jira
* [x]  Link to relevant pull requests, esp. upstream and downstream changes
* [x]  Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/hudson/plugins/ec2/util/AmazonEC2FactoryImpl.java
+++ b/src/main/java/hudson/plugins/ec2/util/AmazonEC2FactoryImpl.java
@@ -2,7 +2,7 @@ package hudson.plugins.ec2.util;
 
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.services.ec2.AmazonEC2;
-import com.amazonaws.services.ec2.AmazonEC2Client;
+import com.amazonaws.services.ec2.AmazonEC2ClientBuilder;
 import hudson.Extension;
 import hudson.plugins.ec2.EC2Cloud;
 import java.net.URL;
@@ -12,8 +12,10 @@ public class AmazonEC2FactoryImpl implements AmazonEC2Factory {
 
     @Override
     public AmazonEC2 connect(AWSCredentialsProvider credentialsProvider, URL ec2Endpoint) {
-        AmazonEC2 client =
-                new AmazonEC2Client(credentialsProvider, EC2Cloud.createClientConfiguration(ec2Endpoint.getHost()));
+        AmazonEC2 client = AmazonEC2ClientBuilder.standard()
+                .withCredentials(credentialsProvider)
+                .withClientConfiguration(EC2Cloud.createClientConfiguration(ec2Endpoint.getHost()))
+                .build();
         client.setEndpoint(ec2Endpoint.toString());
         return client;
     }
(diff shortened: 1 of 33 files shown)


basil:
Can you please split these into three separate commits/PRs? This is a bit too much to review in a single PR. Thanks!

strangelookingnerd:
Split up to multiple commits as requested.

basil:
Conflicted.

strangelookingnerd:
Resolved conflicts.

basil:
The changes generally look good, but the switch statement improvements conflict with #1021, which has already been tough to keep rebased as it is. Assuming those switch statement improvements were generated with automation, perhaps we could undo them and save them for after #1021 goes in.

strangelookingnerd:
> The changes generally look good, but the switch statement improvements conflict with #1021, which has already been tough to keep rebased as it is. Assuming those switch statement improvements were generated with automation, perhaps we could undo them and save them for after #1021 goes in.

Understood, I reverted the enhanced switch statements.

basil:
Failing Spotless now.

basil (approved):
Thanks for the PR!

src/main/java/hudson/plugins/ec2/util/AmazonEC2FactoryImpl.java
```diff
@@ -12,8 +12,10 @@ public class AmazonEC2FactoryImpl implements AmazonEC2Factory {
 
     @Override
     public AmazonEC2 connect(AWSCredentialsProvider credentialsProvider, URL ec2Endpoint) {
-        AmazonEC2 client =
-                new AmazonEC2Client(credentialsProvider, EC2Cloud.createClientConfiguration(ec2Endpoint.getHost()));
+        AmazonEC2 client = AmazonEC2ClientBuilder.standard()
+                .withCredentials(credentialsProvider)
+                .withClientConfiguration(EC2Cloud.createClientConfiguration(ec2Endpoint.getHost()))
+                .build();
```
basil, line 18: @strangelookingnerd Causes [JENKINS-75390](https://issues.jenkins.io/browse/JENKINS-75390).

strangelookingnerd: I’ll look into it and check if there is a possibility that does not require to revert to the deprecated implementation.

jglick:
Note that the switch to `CloudState` in `NoDelayProvisionerStrategy` passing a zero argument is misleading. The point of the `CloudState` overload is to pass an actual number here, like in https://github.com/jenkinsci/kubernetes-plugin/pull/835. At any rate, this has no effect since `EC2Cloud` is still using the deprecated overloads, ignoring any `additionalPlannedCapacity`.

PR #1046, opened 2025-02-10
[JENKINS-75230] Fixes the use of an init script with Windows AMI using a Linux launcher

jmdesprez:
Fix for https://issues.jenkins.io/browse/JENKINS-75230

Previously, the `init.sh` script was uploaded without the `write` permission. On Windows, this creates a `read-only` file. This read-only file is then impossible to replace when the end user wants to reuse the same node. This read-only file causes `scp` to get stuck.

All uploaded files (the init script ant the remoting jar) now have the `write` permissions set.

# :rotating_light: Important notice :rotating_light: 

Users with version `1822.v87175d209b_b_5` or later using a Windows image and the `unix` launcher who upgrade the plugin will have to disable manually the read-only flag on the `init.sh` and `remoting.jar` located in the temporary folder of the Windows machine or, alternatively, delete the the files.

#### Windows AMI

A dedicated AMI was created, based on `ami-0abaed814109888a5` (Microsoft Windows Server 2025 Base). The created AMI contains: `openssh` (configured to use the RSA key), `GitBash` and `java`.

The connection was successful:

```
INFO: Connection allowed after the host key has been verified
<===[JENKINS REMOTING CAPACITY]===>Remoting version: 3248.3250.v3277a_8e88c9b_
Launcher: EC2UnixLauncher
Communication Protocol: Standard in/out
This is a Windows agent
NOTE: Relative remote path resolved to: C:\Users\Administrator
Agent successfully connected and online
```

The configured init script:
```
date >> ~/init.txt
```

The init script was properly executed:
image

#### Linux AMI

AMI used: `ami-08b1d20c6a69a7100` (Amazon Linux 2023 AMI)

The connection was successful:

```
<===[JENKINS REMOTING CAPACITY]===>Remoting version: 3248.3250.v3277a_8e88c9b_
Launcher: EC2UnixLauncher
Communication Protocol: Standard in/out
This is a Unix agent
NOTE: Relative remote path resolved to: /home/ec2-user
Agent successfully connected and online
```

The configured init script:
```
echo "Init script install java"
sudo yum install -y java
```

The init script was properly executed:
```
INFO: Executing init script
Init script install java
Amazon Linux 2023 repository                     41 MB/s |  31 MB     00:00    
Amazon Linux 2023 Kernel Livepatch repository    55 kB/s |  14 kB     00:00    
[...]
```

#### MacLauncher

Same configuration as the Linux AMI, but using the MacLauncher:
```
Launcher: EC2MacLauncher
Communication Protocol: Standard in/out
This is a Unix agent
NOTE: Relative remote path resolved to: /home/ec2-user
Agent successfully connected and online
```

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java
@@ -23,8 +23,6 @@
  */
 package hudson.plugins.ec2.ssh;
 
-import static org.apache.sshd.client.session.ClientSession.REMOTE_COMMAND_WAIT_EVENTS;
-
 import com.amazonaws.AmazonClientException;
 import com.amazonaws.services.ec2.model.Instance;
 import com.amazonaws.services.ec2.model.KeyPair;
@@ -66,19 +64,15 @@ import java.nio.file.Files;
 import java.nio.file.attribute.PosixFilePermission;
 import java.security.PublicKey;
 import java.time.Duration;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.Jenkins;
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.sshd.client.SshClient;
 import org.apache.sshd.client.channel.ChannelExec;
-import org.apache.sshd.client.channel.ClientChannel;
-import org.apache.sshd.client.channel.ClientChannelEvent;
 import org.apache.sshd.client.future.ConnectFuture;
 import org.apache.sshd.client.keyverifier.ServerKeyVerifier;
 import org.apache.sshd.client.session.ClientSession;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java
```diff
@@ -244,75 +238,29 @@ protected void launchScript(EC2Computer computer, TaskListener listener)
...
                     if (StringUtils.isNotBlank(initScript)
                             && !executeRemote(clientSession, "test -e ~/.hudson-run-init", logger)) {
-                        logInfo(computer, listener, "Executing init script");
+                        logInfo(computer, listener, "Upload init script");
                         scp.upload(
                                 initScript.getBytes(StandardCharsets.UTF_8),
                                 tmpDir + "/init.sh",
-                                List.of(PosixFilePermission.OWNER_EXECUTE, PosixFilePermission.OWNER_READ),
+                                List.of(
+                                        PosixFilePermission.OWNER_READ,
```
pankajy-dev, line 246: The changes look good to me. Just want to understand why all the permissions are enabled?

jmdesprez: At least, `OWNER_WRITE` is needed, so that Windows doesn't set the `read-only` flag. `OWNER` should suffice, I'm also adding `GROUP_XXX` and `OTHERS_XXX` to make sure that the permissions will no longer be a problem.

pankajy-dev (approved):

src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java
Priya-CB: `scp.put(initScript.getBytes(StandardCharsets.UTF_8), "init.sh", tmpDir, "0700")`
Before the Mina changes, the script was uploaded with permissions set to 0700, ensuring that only the owner could read, write, and execute it. I guess maintaining the same level of restriction (0700: OWNER_READ, OWNER_WRITE, OWNER_EXECUTE) should be sufficient, as it prevents unintended access by other users.

PereBueno: Agree with @Priya-CB , we should keep permissions as they were before this bug was introduced (https://github.com/jenkinsci/ec2-plugin/blob/1797.ve8a_edb_7e5f6a_/src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java#L216), I'd remove `GROUP_*` and `OTHER_*`

src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -245,76 +239,29 @@ protected void launchScript(EC2Computer computer, TaskListener listener)
...
+                                List.of(
+                                        PosixFilePermission.OWNER_READ,
+                                        PosixFilePermission.OWNER_WRITE,
+                                        PosixFilePermission.OWNER_EXECUTE,
+                                        PosixFilePermission.GROUP_READ,
+                                        PosixFilePermission.GROUP_WRITE,
+                                        PosixFilePermission.GROUP_EXECUTE,
+                                        PosixFilePermission.OTHERS_READ,
+                                        PosixFilePermission.OTHERS_WRITE,
+                                        PosixFilePermission.OTHERS_EXECUTE),
```
PereBueno, line 255: ditto

PereBueno (commented):
Can we verify if group and others permissions are really needed?

jmdesprez:
Thanks for the feedback.
I updated the permissions: 4ede131d77aec00ae843fe3f6fbbcc114cb823b6

Screenshot of the permission for the Linux image
image

pankajy-dev (approved):
LGTM

PereBueno (approved):
LGTM :)

Priya-CB (approved):

tejasdrolia (approved):

fcojfernandez (approved):

PR #1059, opened 2025-03-18
Do not allow Win NTLM connections in FIPS mode

ampuscas:
The plugin is wrapping the `com.hierynomus:smbj` library, which implements the Windows Server Message Block Protocol and includes NT LanMan authentication (NTLM). NTLM is not FIPS-140 approved and this is an issue if Jenkins runs on FIPS mode.

This library is used in the `WinConnection` class, therefore, in FIPS mode this class cannot be instantiated.

Windows agent cannot be used if configured with winrm. As an alternative, use an ssh server and the `UnixLauncher`:
* Enable SSH server (https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse?tabs=gui&pivots=windows-server-2025).
* Use the "UNIX" template
* The plugin will be able to connect to the Windows machine using Mina SSH, but an init script must not be used as it will create a `.sh` file
    * If an init script is desired, then `Git Bash` is the simplest option and the installer will create the file association between `.sh` extension and `Git Bash` making possible to use an init script.

PS: Do not use the init script (must be empty) if there is no shell interpreter configured. Otherwise Windows will hand forever.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/ec2/win/WinConnection.java
+++ b/src/main/java/hudson/plugins/ec2/win/WinConnection.java
@@ -47,7 +47,7 @@ public class WinConnection {
 
     public WinConnection(String host, String username, String password, boolean allowSelfSignedCertificate) {
         if (FIPS140.useCompliantAlgorithms()) {
-            throw new IllegalArgumentException(Messages.EC2Cloud_classNotAllowedInFIPSMode());
+            throw new IllegalArgumentException(Messages.EC2Cloud_ntlmNotAllowedInFIPSMode());
         }
         this.host = host;
         this.username = username;
--- a/src/main/java/hudson/plugins/ec2/win/WinConnection.java
+++ b/src/main/java/hudson/plugins/ec2/win/WinConnection.java
@@ -12,6 +12,8 @@ import com.hierynomus.smbj.share.DiskShare;
 import hudson.plugins.ec2.util.FIPS140Utils;
 import hudson.plugins.ec2.win.winrm.WinRM;
 import hudson.plugins.ec2.win.winrm.WindowsProcess;
+import jenkins.security.FIPS140;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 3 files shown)


olamy (approved):

src/main/resources/hudson/plugins/ec2/Messages.properties
```diff
@@ -18,4 +18,5 @@ EC2Cloud.invalidKeySizeInFIPSMode=Invalid key size, at least 2048 is needed in F
 EC2Cloud.invalidKeySizeECInFIPSMode=Invalid curve size, at least 224 is needed in FIPS mode.
 EC2Cloud.keyIsMandatoryInFIPSMode=The key is mandatory in FIPS mode.
 EC2Cloud.passwordLengthInFIPSMode=When running in FIPS compliance mode, the password must be at least 14 characters long.
+EC2Cloud.classNotAllowedInFIPSMode=NTLM is not FIPS-140 approved.
```
fcojfernandez, line 21:
```suggestion
EC2Cloud.ntlmNotAllowedInFIPSMode=NTLM is not FIPS-140 approved.
```

src/main/java/hudson/plugins/ec2/win/WinConnection.java
```diff
@@ -45,9 +46,9 @@ public WinConnection(String host, String username, String password) {
     }
 
     public WinConnection(String host, String username, String password, boolean allowSelfSignedCertificate) {
-        FIPS140Utils.ensureNoSelfSignedCertificate(allowSelfSignedCertificate);
-        FIPS140Utils.ensurePasswordLength(password);
-
+        if (FIPS140.useCompliantAlgorithms()) {
+            throw new IllegalArgumentException(Messages.EC2Cloud_classNotAllowedInFIPSMode());
```
fcojfernandez, line 50:
```suggestion
            throw new IllegalArgumentException(Messages.EC2Cloud_ntlmNotAllowedInFIPSMode());
```

fcojfernandez (commented):
I know it's a bit picky, but the name of the property and the text of the message doesn't match quite well

fcojfernandez (approved):

PR #1098, opened 2025-06-02
[JENKINS-75746] Removing logic to reuse orphaned EC2 instances 

nevingeorgesunny:
## Summary
This PR simplifies the EC2 Jenkins plugin logic by removing the use of the Max Total Uses field. Instead, it is now hardcoded to always be 1, meaning an EC2 agent will only be used for a single build and then terminated.

## Why This Change?
- The original logic was introduced a long time ago and assumes EC2 agents should be reused (Max Total Uses > 1) to optimize billing and startup times.
- However, this reuse model creates complications in High Availability (HA) Jenkins setups.
- Specifically, when one Jenkins controller takes over from another, an EC2 agent that is still connected might be reassigned, causing inconsistencies or errors.

## Benefits
- Cleaner resource lifecycle — each EC2 agent is guaranteed to be terminated after a single build, avoiding unintended reuse.
- Each build is always run on a fresh EC2 instance

## JIRA
https://issues.jenkins.io/browse/JENKINS-75746

### Testing done
- In my controller, I have triggered many parallel jobs. all of them are started on new ec2 instances. And when the jobs are completed it terminates the instances


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -2138,24 +2122,8 @@ private List<EC2AbstractSlave> provisionOndemand(
 
         logProvisionInfo("Looking for existing instances with describe-instance: " + diRequest);
 
-        DescribeInstancesResponse diResult = ec2.describeInstances(diRequest);
```
nevingeorgesunny, line 2141: without `findOrphansOrStopped` `List<Instance> orphansOrStopped` will be redundant

src/test/java/hudson/plugins/ec2/EC2SlaveMonitorTest.java
```diff
@@ -26,6 +28,11 @@ void setUp(JenkinsRule rule) {
         Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
     }
 
+    @AfterEach
+    void clearInstances() {
+        AmazonEC2FactoryMockImpl.instances.clear();
```
nevingeorgesunny, line 33: Added a clean up , because , the test were failing when run the entire `EC2SlaveMonitorTest.java`.
But individually it was passing.
Could be because I removed the logic to re use orphaned ec2 instaces

nevingeorgesunny:
have removed idleTerminationMinutes and maxTotalUses from the UI for now.
These fields remain in the backend because removing them there would require more extensive changes.

This PR is intended as an initial step and for discussion. I think we should removing these fields from the backend as well.

MarkEWaite:
@fcojfernandez or @res0nance could you take a look at this pull request?  We have an important use case in the CloudBees high availability solution that needs this change.

nevingeorgesunny:
I will be closing this and opening 
https://github.com/jenkinsci/ec2-plugin/pull/1101

PR #1100, opened 2025-06-03
Utilization of Yum or Apt according Unix version

helviojunior:
Utilization of Yum or Apt according Unix version

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
@@ -277,20 +277,43 @@ public class EC2UnixLauncher extends EC2ComputerLauncher {
                         }
                     }
 
-                    executeRemote(
-                            computer,
-                            clientSession,
-                            javaPath + " -fullversion",
-                            "sudo amazon-linux-extras install java-openjdk11 -y; sudo yum install -y fontconfig java-11-openjdk",
-                            logger,
-                            listener);
-                    executeRemote(
-                            computer,
-                            clientSession,
-                            "which scp",
-                            "sudo yum install -y openssh-clients",
-                            logger,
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/ssh/EC2UnixLauncher.java
```diff
@@ -277,20 +277,43 @@ protected void launchScript(EC2Computer computer, TaskListener listener)
...
+
+                    }
+
+                    if (executeRemote(clientSession, "which apt", logger)) {
+                        logInfo(computer, listener, "Using apt as package manager...");
+                        executeRemote(
+                                computer,
+                                clientSession,
+                                javaPath + " -fullversion",
+                                "apt install -y --no-install-recommends fontconfig openjdk-11-jdk-headless",
```
jglick, line 306: Java 17 is the minimum required (21 recommended), so there is no way this could work. Which leads me to think that the Yum variant has not actually worked for a long time either, and nobody noticed because nobody was using this system. Nor should you. Installing Java at runtime forces every agent launch to take much longer, and abuses the bandwidth offered for free by Linux distro servers. 45bae19c5f9585f420e2c02fda801a983aca29a5 was a bad idea, https://github.com/jenkinsci/ec2-plugin/pull/245 made it worse, https://github.com/jenkinsci/ec2-plugin/pull/246 compounded the bad idea, and https://github.com/jenkinsci/ec2-plugin/pull/777 perpetuated it. Since the existing code does not work, it would be better to just delete it and not pretend.

jglick: (If you want to _play_ with the plugin without going to the bother of creating your own AMI, you can already install a Java package using the init script. But you should never do this in a production system or at scale.)

PR #1101, opened 2025-06-06
Adding toggle to prevent using orphaned EC2 instance

nevingeorgesunny:
### Added Toggle to Prevent EC2 Plugin from Reconnecting to Orphaned Instances

In High Availability (HA) Jenkins setups with multiple controller replicas, a condition can occur where EC2 agents are unintentionally "stolen" across replicas. This happens when:

- An EC2 instance is provisioned and connected to one Jenkins controller replica.
- A build is later triggered on a different replica.
- That second replica tries to reconnect and reuse the "orphaned" EC2 instance, unaware that it is already connected elsewhere.

This results in the EC2 agent being taken over by the second replica, disrupting the ongoing build on the first. 

To fix this , I have added a toggle to prevent trying to re connect an orphaned ec2 instance

### How I Tested
- I enabled the toggle and triggered multiple parallel builds.
- All builds successfully started on fresh new EC2 instances.
- I also tested without this feature to verify that the legacy behavior still works as expected — and it does.


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -903,6 +911,7 @@ public SlaveTemplate(
             boolean t2Unlimited,
             ConnectionStrategy connectionStrategy,
             int maxTotalUses,
+            boolean uaseOrphanedNodes,
```
gbhat618, line 914:
```suggestion
            boolean useOrphanedNodes,
```

(possibly typo?)

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -822,6 +829,7 @@ public SlaveTemplate(
...
```
gbhat618, line 832:
```suggestion
            boolean useOrphanedNodes,
```
(possibly typo?)

src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-avoidUsingOrphanedNodes.html
```diff
@@ -0,0 +1,4 @@
+If enabled, the EC2 plugin will not attempt to reconnect EC2 instances
+that are disconnected from this Jenkins controller, or those that are
+in a stopped state but still appear connected.When this option is enabled,
```
gbhat618, line 3:
```suggestion
in a stopped state but still appear connected. When this option is enabled,
```

src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-avoidUsingOrphanedNodes.html
```diff
@@ -0,0 +1,4 @@
...
+it is strongly recommended to set "Maximum Total Uses" to 1
```
gbhat618, line 4: Perhaps can we add a couple of words about what happens if not used `Maximum Totals Uses`?

alternatively when this option is used code can set the `Maximum Total Uses = 1` so it prevents users from running into uses if they missed this setting.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -323,6 +325,7 @@ public SlaveTemplate(
             boolean t2Unlimited,
             ConnectionStrategy connectionStrategy,
             int maxTotalUses,
+            boolean avoidUsingOrphanedNodes,
```
gbhat618, line 328: I see that due to this new field in constructor now you need to update in several below overloaded versions (most of which are also marked `@Deprecated`)

Instead of updating this one, would it be better to create a new constructor and marking this current one as Deprecated 🤔?

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
nevingeorgesunny: 🤦, I put this initially for testing , this is not required here . Removing it

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
nevingeorgesunny: 🤦, I put this initially for testing , this is not required here . Removing it

src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-avoidUsingOrphanedNodes.html
nevingeorgesunny: > alternatively when this option is used, code can set the Maximum Total Uses = 1 so it prevents users from running into issues if they missed this setting.

its still a valid use case . 

what can happen, is if Max use is > 1 , and this toggle is on

there may be EC2 instance that are in stoped state/ or non connected state that will not be "woken" up again.

left this for the user to choose. Im not too sure if we should enforce this.
because if we do then . Idle timeout , Max use will become redundant , and we will go to the problem discussed in https://cloudbees.slack.com/archives/CBX8ST5RS/p1749132298142429

MarkEWaite:
Thanks @nevingeorgesunny .  @fcojfernandez or @res0nance , could you review this pull request as the replacement for:

* #1098 

We have an important use case in the CloudBees high availability solution that needs this change, however, we don't want to break existing users of the plugin.

@dduportal is there any way that you could test drive it to confirm that it is well-behaved in the ci.jenkins.io use cases?

dduportal:
> @dduportal is there any way that you could test drive it to confirm that it is well-behaved in the ci.jenkins.io use cases?

Not until End of June. But ci.jenkins.io only used ephemeral instances in March/April when we tested drive EC2, even for Windows machines.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
nevingeorgesunny: I have tested with loading this with old config. 
i.e. a config xml where `avoidUsingOrphanedNodes` does not exist. and it loaded it as false , it did not break. Also tested with CasC

Im not sure exactly what does deprecated constructors do.

src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-avoidUsingOrphanedNodes.html
fcojfernandez: yes, please let's leave that to the user. I think @nevingeorgesunny have made a good point.

fcojfernandez: > left this for the user to choose. Im not too sure if we should enforce this.
because if we do then . Idle timeout , Max use will become redundant , and we will go to the problem discussed in https://cloudbees.slack.com/archives/CBX8ST5RS/p1749132298142429

Please @nevingeorgesunny don't use internal links to discussions in the public repositories. That makes the review harder for users outside of CloudBees. Please leave a summary in this thread

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -2139,19 +2149,22 @@ private List<EC2AbstractSlave> provisionOndemand(
...
         DescribeInstancesResponse diResult = ec2.describeInstances(diRequest);
-        List<Instance> orphansOrStopped = findOrphansOrStopped(diResult, number);
-
-        if (orphansOrStopped.isEmpty()
-                && !provisionOptions.contains(ProvisionOptions.FORCE_CREATE)
-                && !provisionOptions.contains(ProvisionOptions.ALLOW_CREATE)) {
-            logProvisionInfo("No existing instance found - but cannot create new instance");
-            return null;
-        }
+        List<Instance> orphansOrStopped = new ArrayList<>();
```
fcojfernandez, line 2152: before this change `orphansOrStopped` being empty meant the `return null` in the past line 2148. Now, being empty means to continue with `RunInstancesRequest.Builder riRequestBuilder = riRequest.toBuilder();` If I'm understanding well the change and discussions, this is desired, but I'd like to confirm that @nevingeorgesunny

nevingeorgesunny: Yeah , previously if `orphansOrStopped` is empty and NOT `FORCE_CREATE` ,`ALLOW_CREATE` , it will return null

But now with this toggle enabled  it will always try to  start a new ec2 instance

src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-avoidUsingOrphanedNodes.html
nevingeorgesunny: Context for the problem :
If Max use is set to 1 by force . there  are users that  need to re use the agents like (EC2 MacOS instance to be reused)

Their use case will be disrupted

nevingeorgesunny: or more specifically 
- If agents are forced to terminate after a single use (i.e., max use = 1), builds depending on persistent or reusable EC2 agents (e.g., MacOS or Windows instances) will break.
- These users often reuse agents to optimize high startup cost
- In such scenarios, ephemeral agents are not feasible due to long boot times or expensive billing cycles (e.g., EC2 Mac instances billed hourly).
- Forcing ephemeral-only behavior ignores valid OSS use cases,

fcojfernandez (approved):

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -222,6 +222,8 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
     public int maxTotalUses;
 
+    private boolean avoidUsingOrphanedNodes;
```
jglick, line 225: As a rule, we prefer to avoid negative terminology in options. Would have been clearer (IMO) to name the option, e.g., `reuseOrphanNodes`, then have it default to `true` in `readResolve` (using `Boolean`) for compatibility reasons (probably leave it off for newly configured clouds).

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
jglick, line 2152: More clearly, `List.of()` (or the older `Collections.emptyList()`) to emphasize that we do not intend to mutate it. Clearer still,

```java
List<Instance> orphansOrStopped;
if (!avoidUsingOrphanedNodes) {
    orphansOrStopped = findOrphansOrStopped(diResult, number);
    // …
} else {
    orphansOrStopped = List.of();
}
```

PR #1115, opened 2025-07-16
Orphan node cleanup

nevingeorgesunny:
With the addition of https://github.com/jenkinsci/ec2-plugin/pull/1101 (which introduces a toggle to prevent the reuse of orphaned EC2 instances), there's an increased possibility that orphaned EC2 instances may accumulate in AWS over time.

This PR addresses that potential issue by introducing a PeriodicWork. This task routinely scans for EC2 instances that were provisioned by the EC2 plugin—identified by the tags `jenkins_slave_type` and `jenkins_server_url`  and checks if they are currently connected to a Jenkins controller.

- If the instance is connected to a Jenkins controller, the PeriodicWork task updates the `jenkins_node_last_refresh` tag with the current timestamp.
- This effectively acts as a heartbeat, continuously extending the expiration time of connected EC2 instances.
- If the instance is not connected for a prolonged period, it qualifies as "orphaned" and will be terminated.

This mechanism ensures eventual cleanup of orphaned EC2 instances—even those left behind by other controllers that may have used the same AWS account.

Testing:
- I have tested by connecting to 2 AWS clouds 
- I created some orphaned nodes in each, and they are being cleaned up as expceted

TODO:
- write an IT (had some difficulty writing test with JenkinsRule , maybe with RealJenkinsRule)
     - I will write IT that will be ignored , it will be using real AWS credentials. it will be ment to only run locally
     - FYI , there are no IT in this plugin , opened this PR for now to get reviews
- ~~Test will higher load of running 100+ builds to see if everything okay~~

PR #1122, opened 2025-07-31
[JENKINS-75942] Fixing Orphan node cleanup terminates instances for another Jenkins

nevingeorgesunny:
JENKINS-75942 : https://issues.jenkins.io/browse/JENKINS-75942

With the addition of https://github.com/jenkinsci/ec2-plugin/pull/1115, a Jenkins controller could mistakenly clean up orphaned nodes that belong to a different controller/cloud—even if that other controller/cloud has not opted into the cleanup mechanism.

This PR addresses that issue by introducing an additional safeguard: it now checks both the existing `jenkins_server_ur` tag and a newly added `jenkins_cloud_name` tag. The `jenkins_cloud_name` tag associates each EC2 instance with the specific Jenkins cloud configuration that provisioned it, preventing unauthorized cleanup by unrelated controllers/cloud


src/main/java/hudson/plugins/ec2/EC2CleanupOrphanedNodes.java
```diff
@@ -270,7 +272,10 @@ private void terminateInstance(String instanceId, Ec2Client connection) {
         }
     }
 
-    private Filter tagFilter(String tagName) {
```
nevingeorgesunny, line 273: previously only checking if the tag was present to determine that ec2 instance was started by `ec2-plugin` , now we will check for the value i.e. cloud name and server url

src/main/java/hudson/plugins/ec2/EC2CleanupOrphanedNodes.java
```diff
@@ -129,11 +131,11 @@ private Set<Instance> getAllRemoteInstance(Ec2Client connection, EC2Cloud cloud)
                                             InstanceState.PENDING.getCode(),
                                             InstanceState.STOPPING.getCode())
                                     .build(),
-                            tagFilter(EC2Tag.TAG_NAME_JENKINS_SLAVE_TYPE),
-                            tagFilter(EC2Tag.TAG_NAME_JENKINS_SERVER_URL));
+                            tagFilter(EC2Tag.TAG_NAME_JENKINS_SERVER_URL,jenkinsLocation.getUrl()),
+                            tagFilter(EC2Tag.TAG_NAME_JENKINS_CLOUD_NAME, cloud.getDisplayName()));
```
nevingeorgesunny, line 135: fetching only ec2 instances with the `jenkins_server_url` and `jenkins_cloud_name`

nevingeorgesunny:
[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-61042) FTR

PR #1125, opened 2025-07-31
Discard idle instances during shutdown

jglick:
Extracted from #1121.

While testing the spare instances feature I noticed that while the controller launches a pool of spare instances when it starts, it does not terminate them when it shuts down. If the controller is immediately restarted, this is not a big deal since they will be connected again, but it can leave a mess in a CloudBees CI HA controller where each replica has its own set of spare instances. #1115 will clean them up but it could take hours. With this patch, the spares are neatly cleaned up.

I needed to introduce a `Future` into the internal API to make sure that the termination request is processed before the web app is unloaded and class loading from plugins becomes impossible; otherwise the termination is unreliable and mostly does not work at all. This does clash a bit with #1015.

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
+++ b/src/main/java/hudson/plugins/ec2/EC2AbstractSlave.java
@@ -44,6 +44,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -757,7 +758,7 @@ public abstract class EC2AbstractSlave extends Slave {
     /**
      * Terminates the instance in EC2.
      */
-    public abstract void terminate();
+    public abstract Future<?> terminate();
 
     void stop() {
         try {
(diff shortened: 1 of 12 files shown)


jglick:
> all my EC2 Agents will be terminated?

Not exactly. Only idle agents; and only up to the number of configured spare instances, if that option is even used. So at most as many would have been launched when the controller starts.

If you think even that is going too far, that some people may be configuring spare instances but prefer for the unused spares to be left running when the controller terminates, then yes of course it could be made into (sigh) yet another checkbox in the agent template.

jglick:
(BTW I have not yet looked into the options for covering this and my other PRs via automated tests. I do not trust the mocks much, but I guess this plugin is set up neither for LocalStack container testing nor for live testing against AWS with a bring-your-own credentials model.)

jglick:
> instances will be in a Stopped state (i.e. not running, but not terminated). What we would like is for those instances to remain as they are

I am definitely not planning to implement that, but if that is something that currently works, then it sounds like the behavior in this PR must not be enabled for users with requirement. So, I will make it an option.

> assistance with basic testing in our AWS environment

Sorry, I was unclear—I have already tested this in an EKS cluster with IAM-based access to EC2. Of course additional sanity checks by others are welcome. I was just uncertain what _automated_ test coverage is expected in general for changes in this plugin.

jglick:
Thanks. @res0nance / @fcojfernandez?

fcojfernandez (approved):

PR #1138, opened 2025-08-28
Removed associatePublicIp check for building the NW interface obj

vwagh-dev:
### Issue

See: https://issues.jenkins.io/browse/JENKINS-75002
The issue with the plugin arises when the "Associate Public IP" checkbox is not selected. In this case, it neither sends the network interface nor the associatePublicIpAddress field, relying entirely on the subnet configuration.

### RCA
`NetworkInterface` object was getting build only if & only if `allowPublicIp` value is `true` otherwise not.

### Fix
Now irrespective `allowPublicIp` value, `NetworkInterface` object will get build.

### Testing done
1. Build is green.
2. Added 2 new unit tests which checks for existence of the `networkInterface` object irrespective of `allowedPublicIpAddress` values.
3. Tried provisioning the EC2 instance with changes & its launching the instance in both cases.
    - Checked `Allow Public IP` : It launches the EC2 instance & connect to it. ✅ 
    - Unchecked `Allow Public IP` ⚠️ 
      - **In localhost (Launch but not connects):**  it just launch the EC2 instance but do not connect ❌ 
      - **On AWS cloud (Launch & connects):** Where jenkin instance on AWS cloud, it launch as well connects to EC2 instance ✅ 

### Submitter checklist

### DIFF
--- a/src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
+++ b/src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
@@ -1656,7 +1656,7 @@ class SlaveTemplateTest {
         return r.createWebClient().goTo(ac.getUrl() + "configure").getFormByName("config");
     }
 
-     @Test
+    @Test
     void testNetworkInterfaceAssociatePublicIpTrue() throws Exception {
         SlaveTemplate template = mock(SlaveTemplate.class);
         when(template.getAssociatePublicIp()).thenReturn(true);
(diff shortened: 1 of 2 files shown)


src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
```diff
@@ -1655,4 +1655,34 @@ public void testWindowsSSHConfigRoundTrip() throws Exception {
     private HtmlForm getConfigForm(EC2Cloud ac) throws IOException, SAXException {
         return r.createWebClient().goTo(ac.getUrl() + "configure").getFormByName("config");
     }
+
+     @Test
```
RanjitCS, line 1659: **nitpick**: This doesnt look aligned, extra space?

amuniz (approved):

src/test/java/hudson/plugins/ec2/SlaveTemplateTest.java
vwagh-dev: Thanks a lot for pointing out this stupid mistake. I thought that it is fixed & didn't looked again. Now it is fixed.

PR #1141, opened 2025-09-04
Jenkins 76058

Dohbedoh:
* Fix `OpenSshCertificate` public key handling. If the a `PublicKey` of type `OpenSsfCertificate` is passed by SSHD verifier, then we must unwrap the `getCertPubKey` for host key verification.
* Only manipulate the `PublicKey` to a `HostKey` if necessary (that would prevent future problem related to Algorithm type handling when No Verification is selected).

### Testing done

* Added unit tests
* No manual testing doen yet

### Submitter checklist


src/main/java/hudson/plugins/ec2/util/KeyHelper.java
```diff
@@ -152,4 +154,28 @@ public static String getSshAlgorithm(@NonNull PublicKey serverKey) {
...
+        // Emulate Oracle so that the algorithm returned by
+        // org.bouncycastle.jcajce.provider.asymmetric.edec.BCEdDSAPublicKey.getAlgorithm
+        // is the one expected by org.apache.sshd.common.config.keys.KeyUtils
+        try {
+            Properties.setThreadOverride(Properties.EMULATE_ORACLE, true);
+            // Generate MD5 fingerprint just like ssh-keygen
+            byte[] rawFingerprint = KeyUtils.getRawFingerprint(BuiltinDigests.md5.get(), serverKey);
+            return BufferUtils.toHex(':', rawFingerprint).toLowerCase();
+        } catch (Exception e) {
+            return "";
```
jglick, line 176: Is some exception expected here? Seems like it should at least log a warning.

Dohbedoh: Not expected. It all goes does to implementation of Digest / decoder I guess.
We can add a warning.

Dohbedoh:
Thanks @jglick

olamy (approved):
LGTM

src/main/java/hudson/plugins/ec2/ssh/verifiers/HostKeyHelper.java
```diff
@@ -57,6 +64,24 @@ public static HostKeyHelper getInstance() {
...
+     * @throws IOException if the key cannot be processed
+     */
+    @CheckForNull
+    public HostKey getHostKey(@NonNull PublicKey serverKey) throws IOException {
+        String sshAlgorithm = KeyHelper.getSshAlgorithm(serverKey);
+        if (sshAlgorithm == null) {
+            return null;
+        }
+        AsymmetricKeyParameter parameters = PublicKeyFactory.createKey(serverKey.getEncoded());
+        return new HostKey(serverKey.getAlgorithm(), OpenSSHPublicKeyUtil.encodePublicKey(parameters));
```
Dohbedoh, line 82: This should not be `serverKey.getAlgorithm()` but `sshAlgorithm`. This is causing an issue https://issues.jenkins.io/browse/JENKINS-76077

PR #1142, opened 2025-09-05
reconnect when agent is offline

car-roll:
When an EC2 agent loses connection, the pipeline will detect this and terminate the job after 5 minutes. However, there is no retry logic. So in the case of an unstable connection, one lost connection means the pipeline will fail. Previously, the internal check of `EC2RetentionStrategy` only checks if a node goes idle. It never actually checked if the node was offline first.

### Testing done

Manual testing:
Created a pipeline that called a long `sleep` on an ec2 agent.
Once EC2 agent was spun up, went to script console and terminated the `sshd` connection
Waited for the pipeline to reconnect to the agent.
Cancelled multiple times, reconnection always happened

TODO: kill sshd connection just as job is about to complete, see if it reconnects and ends the job

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
+++ b/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
@@ -117,9 +117,9 @@ public class EC2RetentionStrategy extends RetentionStrategy<EC2Computer> impleme
 
     private long internalCheck(EC2Computer computer) {
         /*
-         * If we've been told never to terminate, or node is null(deleted), no checks to perform
+         * If we've been told never to terminate, or node is null(deleted), no checks to perform, or retention strategy disabled
          */
-        if (idleTerminationMinutes == 0 || computer.getNode() == null) {
+        if (idleTerminationMinutes == 0 || computer.getNode() == null || DISABLED) {
             return CHECK_INTERVAL_MINUTES;
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -139,23 +139,23 @@ private long internalCheck(EC2Computer computer) {
...
-                // info
-                uptime = computer.getUptime();
-                launchedAt = computer.getLaunchTime();
-            } catch (SdkException | InterruptedException e) {
-                // We'll just retry next time we test for idleness.
-                LOGGER.fine("Exception while checking host uptime for " + computer.getName()
-                        + ", will retry next check. Exception: " + e);
-                return CHECK_INTERVAL_MINUTES;
-            }
+        final long uptime;
```
jglick, line 142: (ignore WS)

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -139,23 +139,23 @@ private long internalCheck(EC2Computer computer) {
...
+            uptime = computer.getUptime();
+            launchedAt = computer.getLaunchTime();
+        } catch (SdkException | InterruptedException e) {
+            // We'll just retry next time we test for idleness.
+            LOGGER.fine("Exception while checking host uptime for " + computer.getName()
+                    + ", will retry next check. Exception: " + e);
+            return CHECK_INTERVAL_MINUTES;
+        }
 
+        if (computer.isIdle()) {
```
jglick, line 158: Not sure I see the point of this part of the patch. `state` / `uptime` / `launchedAt` are still used only if `idle`, so why calculate them if it is not?

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -245,6 +245,16 @@ private long internalCheck(EC2Computer computer) {
                 }
             }
         }
+
+        if (computer.isOffline()) {
+            if (!computer.isConnecting()) {
+                // Retry connection to agent until the job times out
+                // see: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/463
```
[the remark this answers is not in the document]
car-roll: Yeah, I only left that as a comment as "proof" that the job would time out, so I was just trying indefinitely here.

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
car-roll: Yeah, I actually pulled them out when I refactored teh code differently. I'll move them back in

src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java
```diff
@@ -103,6 +103,7 @@ public long check(EC2Computer c) {
                 long currentTime = this.clock.millis();
 
                 if (currentTime > nextCheckAfter) {
+                    attemptReconnectIfOffline(c);
```
jglick, line 106: Do we want to do this even if `DISABLED`, in which case AFAICT the other behaviors of the strategy are a no-op?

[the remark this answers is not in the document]
car-roll: bad choice of words. I meant after i clicked the `update PR` button. But yes, noted

mikecirioli (approved):
LGTM

mikecirioli (approved):
I have validated this fix as working by reproducing the original issue, updating the plugin, and testing again.  After the update the agent is successfully reconnected and the build finishes correctly (from the perspective of the controller).

jglick:
I guess up to @res0nance @fcojfernandez et al. to decide whether/when to merge? `bug` label recommended.

res0nance (approved):

Shohou:
I believe this broke the flow when instance is kept in Stopped state after job is finished and idle timeout has passed, which is controlled by "Stop/Disconnect on Idle Timeout" checkbox setting. Created an issue - https://issues.jenkins.io/browse/JENKINS-76151

PR #1148, opened 2025-09-26
Deprecate boolean `associatePublicIp` and add `AssociateIpStrategy` enum for IP assignment control 

apuig:
[JENKINS-75002 - Allow blocking EC2 public IP](https://issues.jenkins.io/browse/JENKINS-75002)

Currently, public IP assignment behavior is inconsistent:

- `associatePublicIp=true`:
  - Creates a `NetworkInterface`
  - Sets `associatePublicIpAddress(true)`
  - Configures subnet and security groups on the interface

- `associatePublicIp=false`:
  - **On-demand**:
    - Does *not* use a `NetworkInterface`
    - Configures subnet/security groups at the RunInstance level
    - Public IP assignment depends on the subnet's `MapPublicIpOnLaunch` setting (inherited)
  - **Spot**:
    - Uses a `NetworkInterface`
    - Sets `associatePublicIpAddress(false)`, enforcing *no* public IP (does *not* inherit)

This prevents reliable enforcement of "no public IP" via AWS Service Control Policies. The current logic follows [PR 447](https://github.com/jenkinsci/ec2-plugin/pull/447), which reverted [JENKINS-58578](https://issues.jenkins.io/browse/JENKINS-58578), losing the ability of `associatePublicIp=false` to always disable public IP assignment.

### Proposal

- **Always create** the primary `NetworkInterface`.
- Replace the boolean `associatePublicIp` with the `AssociateIpStrategy` enum:
    - `PUBLIC_IP`: Always assign a public IP (`associatePublicIpAddress=true`)
    - `PRIVATE_IP`: Never assign a public IP (`associatePublicIpAddress=false`)
    - `SUBNET`: Inherit the subnet's `MapPublicIpOnLaunch` (leave unset)
    - `DEFAULT`: Matches current behavior: `PRIVATE_IP` for spot, `SUBNET` for on-demand

:speaking_head:  The `DEFAULT` option aims to easy backward compatibility and is used as the UI default.

### Testing done

* :construction:  work in progress
* 

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -217,7 +217,10 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
 
     public HostKeyVerificationStrategyEnum hostKeyVerificationStrategy;
 
-    public final boolean associatePublicIp;
+    public AssociateIPStrategy associateIPStrategy;
+
+    @Deprecated
+    public transient boolean associatePublicIp;
 
     protected transient EC2Cloud parent;
 
@@ -325,7 +328,7 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
             boolean deleteRootOnTermination,
             boolean useEphemeralDevices,
             String launchTimeoutStr,
-            boolean associatePublicIp,
+            AssociateIPStrategy associateIPStrategy,
             String customDeviceMapping,
             boolean connectBySSHProcess,
             boolean monitoring,
@@ -382,7 +385,6 @@ public class SlaveTemplate implements Describable<SlaveTemplate> {
         this.subnetId = subnetId;
         this.tags = tags;
         this.idleTerminationMinutes = idleTerminationMinutes;
-        this.associatePublicIp = associatePublicIp;
         this.connectionStrategy = connectionStrategy == null ? ConnectionStrategy.PRIVATE_IP : connectionStrategy;
         this.useDedicatedTenancy = tenancy == Tenancy.Dedicated;
         this.connectBySSHProcess = connectBySSHProcess;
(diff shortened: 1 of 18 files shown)


src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -431,9 +433,106 @@ public SlaveTemplate(
         this.metadataHopsLimit =
                 metadataHopsLimit != null ? metadataHopsLimit : EC2AbstractSlave.DEFAULT_METADATA_HOPS_LIMIT;
         this.enclaveEnabled = enclaveEnabled != null ? enclaveEnabled : EC2AbstractSlave.DEFAULT_ENCLAVE_ENABLED;
+        this.associateIPStrategy = associateIPStrategy != null ? associateIPStrategy : AssociateIPStrategy.DEFAULT;
```
Dohbedoh, line 436: If the `associateIPStrategy` is `null`, shouldn't we check want to use `AssociateIPStrategy.backwardsCompatible(associatePublicIp)` instead of the `AssociateIPStrategy.DEFAULT` ?

apuig:
Testing notes have been updated and things are looking good ! :candy: 
Would love to hear thoughts from more experienced contributors.

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -2011,11 +2115,7 @@ HashMap<RunInstancesRequest, List<Filter>> makeRunInstancesRequestAndFilters(
 
         InstanceNetworkInterfaceSpecification.Builder netBuilder = InstanceNetworkInterfaceSpecification.builder();
         if (StringUtils.isNotBlank(subnetId)) {
-            if (getAssociatePublicIp()) {
-                netBuilder.subnetId(subnetId);
-            } else {
-                riRequestBuilder.subnetId(subnetId);
-            }
```
jeromepochat, line 2018: **Question:** No sure to get the logic change. I would expect the new code to preserve the current logic testing `associateIPStrategy == AssociateIPStrategy.PUBLIC_IP`. Is it expected to not set anymore `subnetId` (and other properties bellow) to `riRequestBuilder`?

IIUC, this is not necessary anymore because `riRequestBuilder` [now always gets the network interface](https://github.com/jenkinsci/ec2-plugin/pull/1148/files#diff-2569749802877edc1d4e6fb86580ac97ae86dda6ddfc31d85807032b4200cec9R2164) from `netBuilder`, right?

apuig: Yes, this is expected, and it’s actually a key change needed to make this work. The PR description outlines the inconsistency of only setting the network at the RunInstance (RI) level when using on-demand instances with `associatePublicIp=false`. Let’s elaborate:

As documented in the [SubnetId documentation for RunInstances (RI)](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html):
> If you specify a network interface, you must specify any subnets as part of the network interface instead of using this parameter.

By configuring the NetworkInterface (NI) directly, we gain precise control over network-related attributes. If we don’t do this, AWS will create a default primary network interface with default settings; which previously meant that not setting associatePublicIp would leave the decision to the subnet configuration.

It’s also worth noting that trying to provide configuration options in both RI and NI led to regressions in a previous attempt to address this. For reference, see [PR #1143](https://github.com/jenkinsci/ec2-plugin/pull/1143). A search in Jenkins JIRA for “Network interfaces and an instance-level subnet ID may not be specified on the same request” reveals many issues related to this error, showing problems when users tried to use both methods in the same request.

Similarly, for spot instances, the [SpotLaunchSpecification](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RequestSpotLaunchSpecification.html) allows setting subnetId and securityGroups directly, but it also supports providing an NI. The current spot instance implementation always uses NI configuration.

In my live tests I was able to verify the new paths. NI with `associatePublicIp=null` and `associatePublicIp=false`

jeromepochat (approved):

src/main/java/hudson/plugins/ec2/SlaveTemplate.java
```diff
@@ -325,7 +328,7 @@ public SlaveTemplate(
             boolean deleteRootOnTermination,
             boolean useEphemeralDevices,
             String launchTimeoutStr,
-            boolean associatePublicIp,
+            AssociateIPStrategy associateIPStrategy,
```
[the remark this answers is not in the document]
apuig: thanks for this hint, keeping it in mind

PR #1153, opened 2025-10-14
Fix associatePublicIp backward compatible.

apuig:
jenkinsci/ec2-plugin#1148
JENKINS-75002

As described there is a problem with the backward compatibility on casc.
https://github.com/jenkins-infra/helpdesk/issues/4832

Tests were ~~skipped~~ ignored and not failing (invalid yaml file)

### Submitter checklist

### DIFF
--- a/src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIp.yml
+++ b/src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIp.yml
@@ -1,9 +1,11 @@
 ---
 configuration-as-code:
+  deprecated: warn
 jenkins:
   clouds:
     - amazonEC2:
-        name: "staging"
+        name: "us-east-1"
         privateKey: "${PRIVATE_KEY}"
         templates:
-        - associatePublicIp: true
+          - description:
+            associatePublicIp: true
--- a/src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIpFalse.yml
+++ b/src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIpFalse.yml
@@ -1,9 +1,11 @@
 ---
 configuration-as-code:
+  deprecated: warn
 jenkins:
   clouds:
     - amazonEC2:
-        name: "staging"
+        name: "us-east-1"
         privateKey: "${PRIVATE_KEY}"
         templates:
-        - associatePublicIp: false
+          - description:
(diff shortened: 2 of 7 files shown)


src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIpFalse.yml
```diff
@@ -1,9 +1,11 @@
 ---
 configuration-as-code:
+  deprecated: warn
 jenkins:
   clouds:
     - amazonEC2:
-        name: "staging"
+        name: "us-east-1"
```
jeromepochat: **Question:** dito

src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIpFalseSpot.yml
```diff
@@ -1,11 +1,13 @@
 ---
 configuration-as-code:
+  deprecated: warn
 jenkins:
   clouds:
     - amazonEC2:
-        name: "staging"
+        name: "us-east-1"
```
jeromepochat: **Question:** dito

src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIp.yml
```diff
@@ -1,9 +1,11 @@
 ---
 configuration-as-code:
+  deprecated: warn
 jenkins:
   clouds:
     - amazonEC2:
-        name: "staging"
+        name: "us-east-1"
```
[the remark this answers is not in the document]
apuig: The issue was caused because I based my test on the existing `BackwardsCompatibleConnectionStrategy.yml`, which actually contains invalid YAML. As a result, none of the tests were executed—neither failed nor skipped—they simply weren't reported, which I initially missed.

So, the tests on `ConfigurationAsCodeTest`  uses  `Jenkins.get().getCloud("us-east-1")`, so the YAML needs to define an element with that exact name.

Sorry for overlooking this earlier!

src/test/resources/hudson/plugins/ec2/BackwardsCompatibleAssignIPStrategyUsePublicIpFalseSpot.yml
apuig: see https://github.com/jenkinsci/ec2-plugin/pull/1153/files#r2431460596

PR #1157, opened 2025-12-04
Ignore Spotless reformatting for blame

jglick:
Forgotten after https://github.com/jenkinsci/ec2-plugin/pull/1011 (which conventionally should have been squash-merged BTW, though I guess it does not matter much).


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# https://github.com/jenkinsci/ec2-plugin/pull/1011 Spotless
+98fc9a96b65375210d2dd3f9285d924d91582f3f
```
jglick: 98fc9a96b65375210d2dd3f9285d924d91582f3f

res0nance (approved):

PR #2000, opened 2026-03-25
Ludicrous Mode: Lightweight queuing of work

samrocketman:
What is ludicrous speed?
------------------------

[Obligatory video of ludicrous speed](https://youtu.be/oApAdwuqtn8?si=uJHPP5OXz9GNllY0)

<details><summary>Ludicrous speed is faster than light speed</summary>

---

spaceballs ludicrous speed

</details>

---

Patch details
-------------

`Queue.maintian()` is a critical path in Jenkins core.  Any performance delays can significantly impact Jenkins scheduling work on agents.  At around 800 agents, Jenkins becomes unavailable since it becomes nearly permanently locked with work queued up.  This patch was tested with over 1000 Jenkins agents in AWS and work was scheduled near-instantly.

The EC2 plugin's retention strategy (`check()`, `start()`) is called under the Queue lock and made live EC2 API calls for every agent. The patch splits all EC2 operations into two categories:

- **Lightweight** (runs under Queue lock): Returns immediately with no AWS interaction. Schedules heavy work on background thread pools and returns.
- **Heavyweight** (runs outside Queue lock): EC2 API calls (describeInstances, getState, connect), idle timeout, reconnect — all run on dedicated `HEAVY_WORK_EXECUTOR` and `PROVISIONING_EXECUTOR` thread pools.

Provisioning now returns `PlannedNode` futures immediately instead of blocking on `runInstances`. Instance counts are cached with a 30-second TTL to avoid repeated `describeInstances` calls. Dead-node detection uses batch `describeInstances` (one API call per cloud) instead of per-node calls. When agents come online, `scheduleMaintenance()` is called to trigger an immediate Queue re-evaluation rather than waiting for the next 5-second timer tick.

Patch series
------------

This patch is part of a series which enables my production instance to scale work up to ludicrous amounts of agents.

- ec2 plugin (this patch)
- [job-restrictions plugin]
- [leastload plugin]

AI Analysis
-----------

See detailed [AI analysis for ec2 plugin][1].

Testing done
------------

First pressure tested in a non-production environment. It is now running in production for me for about a week.

Advisory
--------

I plan to update Jenkins core with warnings when there's plugins which adversely affect Jenkins queue performance.  Please keep in mind not merging changes with this strategy will involve this plugin being called out for tanking Jenkins performance.  This patch solves a critical performance issue for your plugin so that it does not tank Jenkins performance (in how fast Jenkins schedules work on agents).

Submitter checklist
-------------------

[1]: https://github.com/samrocketman/jenkins-ai-analysis/blob/main/ludicrous-mode-analysis/ec2-plugin/
[job-restrictions plugin]: https://github.com/jenkinsci/job-restrictions-plugin/pull/210
[leastload plugin]: https://github.com/jenkinsci/leastload-plugin/pull/22

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2ComputerListener.java
+++ b/src/main/java/hudson/plugins/ec2/EC2ComputerListener.java
@@ -4,6 +4,7 @@ import hudson.Extension;
 import hudson.model.Computer;
 import hudson.model.TaskListener;
 import hudson.slaves.ComputerListener;
+import jenkins.model.Jenkins;
 
 @Extension
 public class EC2ComputerListener extends ComputerListener {
@@ -12,6 +13,10 @@ public class EC2ComputerListener extends ComputerListener {
     public void onOnline(Computer c, TaskListener listener) {
         if (c instanceof EC2Computer) {
             ((EC2Computer) c).onConnected();
+        }
+        Jenkins j = Jenkins.getInstanceOrNull();
+        if (j != null) {
+            j.getQueue().scheduleMaintenance();
         }
     }
 }
--- a/src/main/java/hudson/plugins/ec2/CloudHelper.java
+++ b/src/main/java/hudson/plugins/ec2/CloudHelper.java
@@ -3,7 +3,9 @@ package hudson.plugins.ec2;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.logging.Logger;
 import org.apache.commons.lang.StringUtils;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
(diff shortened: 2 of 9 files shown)


src/main/java/hudson/plugins/ec2/EC2ComputerListener.java
```diff
@@ -13,5 +14,9 @@ public void onOnline(Computer c, TaskListener listener) {
         if (c instanceof EC2Computer) {
             ((EC2Computer) c).onConnected();
         }
+        Jenkins j = Jenkins.getInstanceOrNull();
+        if (j != null) {
+            j.getQueue().scheduleMaintenance();
```
res0nance, line 19: Should we actually do this? Won't this schedule in a couple seconds anyway?

samrocketman: It makes a huge difference.

Queue maintenance period is 5 seconds.  Queue maintenance itself completes in a few milliseconds (with over 1000 ec2 agents in my benchmarks scaled up by ec2 plugin).  It is now a cheap call unlike before this patch.

Without it, in a large pipeline ec2 plugin would be artificially inflating the build time by up to 5 seconds every time this isn't called.  In my infrastructure this would artificially add 30 seconds or more to an overall build time.

I am thinking from the perspective of what Jenkins is competing with. GitHub actions completes simple 2-stage pipelines in under 10 seconds. Without this 10 seconds in Jenkins would be spent waiting to schedule the work when there's clearly available capacity.

This is an important piece that shouldn't be removed.

samrocketman: I share a bit more personal perspective in this comment under "Background" heading.  https://github.com/jenkinsci/jenkins/pull/26309#issuecomment-4127904264

res0nance: > It makes a huge difference.
> 
> Queue maintenance period is 5 seconds. Queue maintenance itself completes in a few milliseconds (with over 1000 ec2 agents in my benchmarks scaled up by ec2 plugin). It is now a cheap call unlike before this patch.
> 
> Without it, in a large pipeline ec2 plugin would be artificially inflating the build time by up to 5 seconds every time this isn't called. In my infrastructure this would artificially add 30 seconds or more to an overall build time.
> 
> I am thinking from the perspective of what Jenkins is competing with. GitHub actions completes simple 2-stage pipelines in under 10 seconds. Without this 10 seconds in Jenkins would be spent waiting to schedule the work when there's clearly available capacity.
> 
> This is an important piece that shouldn't be removed.

Clear that makes sense. I just want to make sure that there are no averse side effects from such usage.

@jglick do you have an opinion on if such a pattern is ok?

jglick: From a two-minute glance it makes sense. I am not deeply familiar with this cloud plugin specifically.

res0nance (approved):

samrocketman:
@res0nance I will also look at implementing a label-based hot spares setting in ec2 plugin.

We're using one shot executors but even with schedules the hot spares is unnecessarily expensive.

mwebber:
@samrocketman For us, this is broken - see issue #2002.

samrocketman:
I commented in the issue and will coordinate a fix.

I am OK with this feature being reverted and me working on a revert or I can get a fix out today.  Whichever is faster/acceptable for maintainers.

Either way I'll fix it today.

PR #2007, opened 2026-04-17
Fix SshClient thread leak in EC2SSHLauncher.launchRemotingAgent()

olamy:
- each launchRemotingAgent() creates an Apache MINA SshClient (which spawns internal thread pools nio acceptor/selectors) but only registers cleanup via a Channel.Listener callback at the end when happens of the method. If any exception occurs before computer.setChannel() is reached, the client is never cleaned up and so leaking many threads.
  - With Windows SSH agents that retry frequently during boot, this produces thousands of leaked sshd-SshClient threads.

Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist


olamy:
this is conservative option for #2008

jglick (approved):
Looks reasonable. Has this been tested in a real controller connecting to real AWS, reproducing the stated error and confirming that this addresses it?

olamy:
better fix with #2008

PR #2008, opened 2026-04-17
Use shared SshClient to prevent sshd-SshClient thread leak, to prevent excessive number of threads created

olamy:
- Replace per-call SshClient creation with a single shared instance managed by SSHClientManager, started once at plugin init and stopped at Jenkins shutdown. This
  mirrors the pattern used in the cloudbees-ssh-slaves-plugin.
  - Move signature algorithm preferences from per-client setSignatureFactories() to per-session KexProposalOption.SERVERKEYS reordering via a SessionListener, and
  move host key verification from per-client setServerKeyVerifier() to per-session DelegatingServerKeyVerifier with session metadata.
  - This eliminates the thread leak in launchRemotingAgent() where each failed SSH attempt leaked ~100 sshd-SshClient threads because the SshClient cleanup was only
  registered via a Channel.Listener callback that was never reached on exception paths.
  - Expose performance tuning via system properties: nioWorkers, idleTimeoutMs, heartbeatIntervalMs, authTimeoutMs.

Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java
+++ b/src/main/java/hudson/plugins/ec2/ssh/EC2MacLauncher.java
@@ -31,7 +31,6 @@ import hudson.plugins.ec2.EC2Readiness;
 import hudson.plugins.ec2.SlaveTemplate;
 import hudson.plugins.ec2.util.KeyHelper;
 import hudson.plugins.ec2.util.KeyPair;
-import hudson.plugins.ec2.util.SSHClientHelper;
 import hudson.slaves.CommandLauncher;
 import hudson.slaves.ComputerLauncher;
 import java.io.File;
@@ -45,7 +44,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import org.apache.commons.lang.StringUtils;
-import org.apache.sshd.client.SshClient;
 import org.apache.sshd.client.session.ClientSession;
 import org.apache.sshd.scp.client.CloseableScpClient;
 import org.apache.sshd.scp.common.helpers.ScpTimestampCommandDetails;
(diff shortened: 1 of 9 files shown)


olamy:
this PR is a bit more aggressive especially applying some performance changes (so there is a conservative option to at least fix the thread leak bug #2007)

res0nance:
Are you running this in your own environment? Has it solve your problems

jglick:
@nevingeorgesunny you did some changes in this plugin not long ago; perhaps you want to review this and/or https://github.com/jenkinsci/ec2-plugin/pull/2007?

@olamy there is no information on testing here. Was it tested at scale somehow, and confirmed to fix the reported problem?

[CloudBees-internal link](https://cloudbees.atlassian.net/browse/BEE-71188)

olamy:
Well the other #2007 is just fixing the symptom but not the root (performance bad pattern) cause. While this PR will really fix the performance root cause and obviously the associated leak fixed by #2007 because not re creating client for each connection tentative. 

### Some figures

A default NIO2 client creates three pools on `start()`:
                                                                                                                                                                                                                                                            
| Pool | Size | Source |                                                                                                                                                                                                                                  
|---|---|---|
| `<client>-timer` | 1 scheduled |[`AbstractFactoryManager.java#L539-L542`](https://github.com/apache/mina-sshd/blob/1285419f2f1bef5fad23f001f74d0f07cc469b5b/sshd-core/src/main/java/org/apache/sshd/common/helpers/AbstractFactoryManager.java#L539-L542)|                               
| `<client>-nio2` | `availableProcessors() + 1` | [`Nio2ServiceFactory.java#L48`](https://github.com/apache/mina-sshd/blob/1285419f2f1bef5fad23f001f74d0f07cc469b5b/sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2ServiceFactory.java#L48) |
| `<client>-nio2-resume` | `availableProcessors() + 1` |[`Nio2ServiceFactory.java#L51`](https://github.com/apache/mina-sshd/blob/1285419f2f1bef5fad23f001f74d0f07cc469b5b/sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2ServiceFactory.java#L51)|                                                   

So on 8 cores that's 19 threads. For a single client and so per slave created by this plugin.

### One client for many servers
The [`SshClient` javadoc (lines 157-162)](https://github.com/apache/mina-sshd/blob/1285419f2f1bef5fad23f001f74d0f07cc469b5b/sshd-core/src/main/java/org/apache/sshd/client/SshClient.java#L157-L162) says it clearly `one instance per application, reused across sessions, safe from multiple threads as long as you don't reconfigure it while it's running`.

The pools above are paid once per client, not once per connection. Sessions are cheap: each one registers its socket with the existing `AsynchronousChannelGroup`. 
Idle sessions don't hold threads. So one client handling 500 sessions still sits at roughly 19 threads. 
50 clients handling the same 500 sessions is ~950 threads, each with a 512KB-1MB ish stack. You've spent ~500MB on thread stacks before any real work happens, and you've duplicated the AsynchronousChannelGroup 50 times, plus 50 extra passes over `known_hosts`, identity files, and `ssh_config`.                                                                                                                                         

The CPU side follows the same bad pattern. Each client runs his own `AsynchronousChannelGroup` which runs its own selector loop and its own systems calls: 50 of them and 50 loops wake up, context-switch, and compete for the same cores to do work one single loop could've handled.

Maybe the reason to spin up a client per server was per-server behavior: different host key, different identity, different cipher preferences. You don't need a new client for any of that. See the code in the PR which is customizing this via session context.

Re (nothing urgent though) `AsynchronousChannelGroup` (the standard jdk nio), this could be improved as well by using sshd-mina or sshd-netty, both are improving performance issues with this but this is probably not really significant here (compared to the one client per server current pattern)


Why using a nio library because the  per-connection SshClient is a worst-of-both-worlds pattern (nio vs bio). 
You get the resource cost of NIO without any of the multiplexing benefit, and you pay more than a blocking library would for the same work. Using something such JSch would be less resources consuming than the current pattern.

jglick:
@olamy all sensible in principle, it is just not apparent from the PR description whether the purported problem was reproduced in a test controller, and this PR confirmed to solve it.

olamy:
> @olamy all sensible in principle, it is just not apparent from the PR description whether the purported problem was reproduced in a test controller, and this PR confirmed to solve it.

the fix has been confirmed to work in a production system with very large number of agents

res0nance (approved):

res0nance:
@olamy your commit message is a little lacking it seems to indicate adding some options but its a lot more complicated than that. could you adjust the commit?

olamy:
> @olamy your commit message is a little lacking it seems to indicate adding some options but its a lot more complicated than that. could you adjust the commit?

argh sorry bad local manual rebase of multiple commits

olamy:
@res0nance done

PR #2009, opened 2026-04-17
Update m8i and m8a instance types to InstanceTypeCompat

peterzhuamazon:
Update m8i and m8a instance types to InstanceTypeCompat.
Similar to: https://github.com/jenkinsci/ec2-plugin/pull/1139

Available instance types:
https://aws.amazon.com/ec2/instance-types/m8i/
https://aws.amazon.com/ec2/instance-types/m8a/

### Submitter checklist


peterzhuamazon:
Hi @res0nance would you mind take a look and approve/merge?
We are trying to add new m8 instance types.
Let me know if this is the right way to proceed. Thanks!

andrross (approved):
Looks good to me! I cross-checked the instance types against the AWS instance type names and they appear to be correct.

eukazakov:
Hi!
We are looking forward to using these instance types on Jenkins.
Could you share when we can expect it to be released?
Thanks!

res0nance:
Does updating your aws-sdk not bring the latest instance types?

eukazakov:
I don't see any updates for the aws-sdk for Jenkins
we have the next versions
Amazon EC2Version 2049.v127f0f109c29 
Amazon Web Services SDK 2 :: EC2Version 2.42.33-70.vd69c0763fa_60

res0nance:
We should not have to add to this list https://github.com/jenkinsci/ec2-plugin/pull/1065/changes#r2023225515

eukazakov:
then what should we do to access these m8a instance types on Jenkins?
we don't have them in the list atm

peterzhuamazon:
Yes @res0nance since we are not familiar with this plugin, I have same question as @eukazakov . What should we do to access newer instance types like m8. Thanks.

peterzhuamazon:
Hi @res0nance is the any update here?
Or could you help suggest how to actually enable m8 instances without this change on existing EC2 Plugin release?

Thanks!

maxe2y (approved):

harley84:
Hi all, any updates about this issue?

peterzhuamazon:
I am still not sure how to onboard m8 instance without this change.
And we moved on to use m7 which is sufficient for now.
But if @res0nance could give a hint that would be valuable, as m8 indeed is faster overall.

res0nance:
> I am still not sure how to onboard m8 instance without this change. And we moved on to use m7 which is sufficient for now. But if @res0nance could give a hint that would be valuable, as m8 indeed is faster overall.

This change should not be needed as its for instance type compatibility for serialization between sdk v1 and sdk v2. The instancetypes comes from the SDK listing all of its known values. If you don't see it its likely an issue with your sdk https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L3645-L3654

PR #2023, opened 2026-07-15
Remove Commons Lang 2 usage

gbhat618:
Jenkins core is removing commons-lang2 dependency - https://github.com/jenkinsci/jenkins/pull/26105. This project is using lang2 transitively, needing to drop those usages. Instead of migrating and depending on commons-lang3-api, due to the simple usages, this PR migrates to JDK apis.

**Testing**

```console
$ grep -rn 'import org\.apache\.commons\.lang[.;]' --include='*.java' src/
Exit code: 1
```

```console
$ mvn clean test-compile -DskipTests -Dban-commons-lang-2.skip=false
[INFO] BUILD SUCCESS
[INFO] Total time:  8.115 s
```

```console
$ mvn clean compile spotbugs:check -DskipTests -Dban-commons-lang-2.skip=false
[INFO] Done SpotBugs Analysis....
[INFO] BUILD SUCCESS
[INFO] Total time:  12.231 s
```

### DIFF
--- a/src/main/java/hudson/plugins/ec2/EC2PrivateKey.java
+++ b/src/main/java/hudson/plugins/ec2/EC2PrivateKey.java
@@ -38,7 +38,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.crypto.Cipher;
 import jenkins.bouncycastle.api.PEMEncodable;
-import org.apache.commons.lang.StringUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import software.amazon.awssdk.core.exception.SdkException;
@@ -145,7 +144,7 @@ public class EC2PrivateKey {
             cipher.init(
                     Cipher.DECRYPT_MODE,
                     PEMEncodable.decode(privateKey.getPlainText()).toPrivateKey());
-            byte[] cipherText = Base64.getDecoder().decode(StringUtils.deleteWhitespace(encodedPassword));
+            byte[] cipherText = Base64.getDecoder().decode(encodedPassword.replaceAll("\\s+", ""));
             byte[] plainText = cipher.doFinal(cipherText);
             return new String(plainText, StandardCharsets.US_ASCII);
         } catch (Exception e) {
@@ -161,7 +160,7 @@ public class EC2PrivateKey {
 
     @CheckForNull
     public static EC2PrivateKey fetchFromDisk(String filepath) {
-        if (StringUtils.isNotEmpty(filepath)) {
+        if (filepath != null && !filepath.isEmpty()) {
             try {
                 return new EC2PrivateKey(Files.readString(Paths.get(filepath), StandardCharsets.UTF_8));
             } catch (IOException e) {
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,7 @@ THE SOFTWARE.
     <hpi.compatibleSinceVersion>1885</hpi.compatibleSinceVersion>
     <spotless.check.skip>false</spotless.check.skip>
     <ban-junit4-imports.skip>false</ban-junit4-imports.skip>
+    <ban-commons-lang-2.skip>false</ban-commons-lang-2.skip>
   </properties>
 
   <dependencyManagement>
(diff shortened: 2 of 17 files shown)


jglick (approved):
Looks OK from a quick read.

src/main/java/hudson/plugins/ec2/EC2PrivateKey.java
```diff
@@ -145,7 +144,7 @@ public String decryptWindowsPassword(String encodedPassword) throws SdkException
             cipher.init(
                     Cipher.DECRYPT_MODE,
                     PEMEncodable.decode(privateKey.getPlainText()).toPrivateKey());
-            byte[] cipherText = Base64.getDecoder().decode(StringUtils.deleteWhitespace(encodedPassword));
+            byte[] cipherText = Base64.getDecoder().decode(encodedPassword.replaceAll("\\s+", ""));
```
apuig, line 147: noticed the impl of `deleteWhitespace` is quite different than a regex replacement. I'm tempted to suggest inlining ~15LoC to avoid surprises

apuig (commented):
Thanks for the thoughtful cleanup.
I’m waiting for your comments on `deleteWhitespace` before approving.

src/main/java/hudson/plugins/ec2/EC2PrivateKey.java
gbhat618: the `deleteWhiteSpace` documents it deletes all matching `Character.isWhiteSpace` (https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isWhitespace-char-); however here the `encodedPassword` is a base64 string, so can't have other than `\r\n`. The input is API controlled - it is for windows VMs a password is generated by the AWS provisioner using the keys provided, this password is later retrieved by the plugin and used in auth workflow - so it is guaranteed to be base64 itself.

gbhat618:
Hi @res0nance, can you also help review and merge if all is good.
I am not sure what label to put for this, but we will need a release for it to later test and update in bom https://github.com/jenkinsci/bom/pull/6234

src/main/java/hudson/plugins/ec2/EC2PrivateKey.java
apuig: Thanks for this thorough check

apuig (approved):

gbhat618:
>I am not sure what label to put for this

per https://github.com/jenkinsci/conventional-commits-plugin/blob/1ecc2648f16172b7e6e4f9525830edec236e1760/.github/workflows/cd.yaml#L64 and https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.yml

these labels that result in a release are,
```
  ┌───────┬──────────────────────────┬─────────────────────────────────────────────────────────┐
  │ Emoji │         Category         │                      Labels to use                      │
  ├───────┼──────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 💥    │ Breaking changes         │ breaking                                                │
  ├───────┼──────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 🚨    │ Removed                  │ removed                                                 │
  ├───────┼──────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 🎉    │ Major features           │ major-enhancement, major-rfe                            │
  ├───────┼──────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 🐛    │ Bug fixes (both tiers)   │ major-bug, bug, fix, bugfix, regression, regression-fix │
  ├───────┼──────────────────────────┼─────────────────────────────────────────────────────────┤
  │ ⚠     │ Deprecated               │ deprecated                                              │
  ├───────┼──────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 🚀    │ New features             │ enhancement, feature, rfe                               │
  ├───────┼──────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 👷    │ Plugin developer changes │ developer                                               │
  └───────┴──────────────────────────┴─────────────────────────────────────────────────────────┘
```

maybe putting the `developer` is appropriate as compared to other labels.. 🤔

jglick:
`developer` is the appropriate label for changes of this kind. `removed` is used for user functionality which has been dropped.

PR #2027, opened 2026-07-27
Improve multi-AZ provisioning by being capacity-aware

samrocketman:
Jenkins LoadStatistics clock cycle is 10 seconds by default.  When configuring a template with multiple availability zones some AZs may not have capacity.  If provisioning throws an exception there's no fallback AZ and current provisioning is delayed by LoadStatistics clock cycle.

### Testing done

Testing is WIP

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
+++ b/src/main/java/hudson/plugins/ec2/SlaveTemplate.java
@@ -53,6 +53,7 @@ import jakarta.servlet.ServletException;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
+import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
@@ -67,6 +68,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


samrocketman:
We're hitting physical capacity limits for on-demand provisioning in AWS confirmed by Amazon support during peak usage times.  So AZs have become more critical than before.  Specifically in 32GB RAM instance types (`c*.4xlarge` and `m*.2xlarge`).

This is more of a scaling problem and I am hitting edge case problems in the ec2-plugin so there's at least a few more PRs I need to make around this.  Like dynamic hot spares based on label load instead of per cloud template (autoscale hot spares up and down). Hot spares with one shot executors are really problematic and so I plan to make PRs redesigning it a bit.  I have a suboptimal solution using complementary schedules (tripled my templates with 3 sets of schedules; on-hours hot spares, off hours hot spares, and weekend hot spares) but hot spares have a ton of unnecessary cost waste which would be solved by autoscaling hot spares.

Should I file an issue to discuss design? I won't make any unrelated changes in this PR.

sgleske-ias:
I have finished testing this in a live environment.  I think it is safe to merge and the fallback behavior skips AZs with no hardware available

Testing notes:

* I used a known type that was in one AZ but not another during testing.
* Capacity (and zones with no capacity) were confirmed by Amazon Support before doing the testing.
* Realtime physical capacity of different instance types is not publicly available without engaging Amazon Support.

sgleske-ias:
@res0nance take a look when you get a chance.  I'm going to implement #2028 building on this code since it will likely conflict if I base the feature on `master` branch.

I will keep an eye out for post-merge reports and support like I did with ludicrous patch series.

res0nance (approved):

PR #2039, opened 2026-09-16
Fix duplicate root block device mappings in setupRootDevice

boyadzhievb:
## Summary
- `setupRootDevice` produced duplicate `/dev/xvda` entries because the `deleteRootOnTermination` block and the encryption block each independently added to the device mapping list
- This caused spot instance launches to fail with three conflicting block device specifications for the same mount point
- Fixed by building all attributes on a single builder and using `set(0, ...)` to replace once instead of multiple `add(0, ...)`

## Test plan
- [x] Existing `testSetupRootDevice*` unit tests pass (4/4)
- [x] Verified via CloudTrail that RunInstances now sends exactly one `/dev/xvda` block device mapping
- [x] Spot instance launches successfully with the fix


sgleske-ias:
Can you describe more about your template?  I use custom device mapping with encrypted devices and had no issues launching spot so I'd like to understand more about the problem before I look through the fix.

In some cases, there can legitimately be multiple devices attached for a launch (such as an extra storage volume, an nvme, etc).  Are you defining spot price or not?

Can you post your template XML with infrastructure-specific values removed? (redact AMI/subnets, etc)

What version of the plugin are you on when you experience this?  In the past week there's been a rapid release of updates around spot performance.

samrocketman:
I think I've confirmed the bug but I would like to see your config sample. Also, this patch is hiding an issue without addressing the bug if I understand it correctly.  When I have your config I'll be able to confirm my assumptions.

boyadboz:
## Why Three `/dev/xvda` Entries Is a Misconfiguration

Three block devices all mapped to `/dev/xvda` — this is a conflict, not redundancy.

1. **Same device name, three times.** All three entries target `/dev/xvda`. A device name is a unique mount point — you can't attach three different EBS volumes to the same path. The last one processed wins (or AWS rejects the request outright), making the other two dead config that adds confusion.

2. **Contradictory specs.** The three entries disagree on what `/dev/xvda` should look like:
   - Entry 1: gp3, 50 GB, from snapshot, **not encrypted**, 3000 IOPS, 125 throughput
   - Entry 2: gp3, 50 GB, from snapshot, **encrypted**, 3000 IOPS, 125 throughput
   - Entry 3: **no type, no snapshot, no IOPS, no throughput** — just 50 GB with deleteOnTermination

   Which one actually applies? It depends on how the API resolves duplicates, which is undefined behavior you don't want to rely on.

3. **Snapshot + encryption mismatch.** Entry 2 adds `"encrypted": true` on top of a snapshot that may or may not already be encrypted. If the snapshot is unencrypted, AWS will re-encrypt it with the default KMS key — but entry 1 uses the same snapshot without encryption. Having both in the same config signals nobody decided whether this volume should be encrypted.

4. **Entry 3 is a bare-minimum fallback that undoes everything.** It drops the snapshot, volume type, IOPS, and throughput. If this entry wins, the instance boots with a blank 50 GB volume (no OS, no data) on whatever the default volume type is — likely not what anyone intended.

### The EBS Encryption Problem in This Event

Looking at the block device mappings in the request, there are 3 conflicting `/dev/xvda` entries:

| # | Device | Encrypted | KMS Key | Snapshot |
|---|--------|-----------|---------|----------|
| 1 | /dev/xvda | ❌ Not set | None | snap-030c6ae46a263eebf |
| 2 | /dev/xvda | ✅ True | None (default AWS key) | snap-030c6ae46a263eebf |
| 3 | /dev/xvda | ❌ Not set | None | None (50GB blank) |

This is a misconfiguration in the Jenkins EC2 Cloud plugin — the same device `/dev/xvda` is being specified 3 times. Entry #2 triggers encryption using the default AWS managed EBS key, which then requires `kms:CreateGrant` from the `AWSServiceRoleForEC2Spot` role.

### What It Should Be

A single `/dev/xvda` entry with one clear, intentional specification — pick encrypted or not, pick the snapshot or not, and define the volume type once.



What i have in RunInstance:
```
        "userData": "<sensitiveDataRemoved>",
        "instanceType": "t3.xlarge",
        "blockDeviceMapping": {
            "items": [
                {
                    "deviceName": "/dev/xvda",
                    "ebs": {
                        "snapshotId": "snap-030c6ae46a263eebf",
                        "volumeSize": 50,
                        "deleteOnTermination": true,
                        "volumeType": "gp3",
                        "iops": 3000,
                        "throughput": 125
                    }
                },
                {
                    "deviceName": "/dev/xvda",
                    "ebs": {
                        "snapshotId": "snap-030c6ae46a263eebf",
                        "volumeSize": 50,
                        "deleteOnTermination": true,
                        "volumeType": "gp3",
                        "iops": 3000,
                        "encrypted": true,
                        "throughput": 125
                    }
                },
                {
                    "deviceName": "/dev/xvda",
                    "ebs": {
                        "volumeSize": 50,
                        "deleteOnTermination": true
                    }
                }
            ]
```

samrocketman:
Can you share your Jenkins config? Specifically the template.

boyadboz:
For sure, if this is not problem you can disregard the PR , as my test shows that it is working without the change as well, but i got all this recommendation from AWS Q, It could be wrong. 
```
jenkins:
  clouds:
    - amazonEC2:
        name: "<cloud-name>"
        instanceCapStr: 80
        credentialsId: "<aws-credentials-id>"
        sshKeysCredentialsId: "<ssh-key-credentials-id>"
        useInstanceProfileForCredentials: false
        noDelayProvisioning: true
        region: "eu-west-1"
        templates:
          - description: "Docker"
            type: "T3Xlarge"
            labelString: "docker"
            instanceCapStr: 15
            minimumNumberOfSpareInstances: 0
            customDeviceMapping: "/dev/xvda=:50"
            ami: "<ami-id>"
            subnetId: "${EC2_CLOUD_SUBNETS}"
            initScript: |-
              sudo cloud-init status --wait
            launchTimeoutStr: "240"
            remoteFS: "./jenkins"
            spotConfig:
              spotMaxBidPrice: "0.09"  # $0.1824 ~50% of t3.xlarge on-demand price
              fallbackToOndemand: true
              useBidPrice: true  
            amiType:
              unixData:
                slaveCommandPrefix: "sudo"
                sshPort: "22"
                bootDelay: "30"
            remoteAdmin: "ec2-user"
            <<: *ec2_data_anchor
```

samrocketman:
No worries I think you found a legitimate bug, your config helps.  I'll look through it today

